PR #78, opened 2019-12-20
Replace str concatination with os.path.join in saving

mishinma:
Say you have a detector `od` and you want to save it in `mysavepath`. If you call `save_detector(od, "mysavepath/")` the result will be as expected:
```
mysavepath/meta.pickle
mysavepath/[od.name].picle
mysavepath/model/[weights]
```
but if your input filepath doesn't have a trailing slash  `save_detector(od, "mysavepath")` you will get:
```
mysavepathmeta.pickle
mysavepath[od.name].picle
mysavepathmodel/[weights]
```

This PR replaces all occurrences of path concatenation using the  `+` operator with `os.path.join` in the `utils.saving` module. This makes handling path operations in the module more robust.

### DIFF
--- a/alibi_detect/utils/saving.py
+++ b/alibi_detect/utils/saving.py
@@ -64,7 +64,7 @@ def save_detector(detector: Data,
         os.mkdir(filepath)
 
     # save metadata
-    with open(filepath + 'meta.pickle', 'wb') as f:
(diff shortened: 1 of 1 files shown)


mishinma:
The tests passed locally

PR #116, opened 2020-04-04
Update alibi-detect integrations

cliveseldon:
* Update alibi-detect cloudevents server
 * Update cifar10 outlier detection example for kfserving
 * Add drift example for kfserving
 * Utilize alibi-detect-server for both examples


ryandawsonuk:
This is being replaced by PRs to seldon-core and KFServing. This code will come out of alibi-detect.

PR #133, opened 2020-09-14
Model distillation

gipster:
* Added `model_distillation.py` module under ad
* Added `loss_distillation` function in `losses.py`
* Added `cd_distillation_cifar10.ipynb` example notebook in examples


arnaudvl:
- [ ] fix flake8 error 
- [ ] fix `saving.py` (*'AdversarialAE'* seems to have been accidentally replaced with *'AdversarialMD'* from DEFAULT_DETECTORS)
- [ ] add tests for `model_distillation.py`
- [ ] extend tests for `saving.py` with model distillation saving fn
- [ ] extend tests for `losses.py` with new loss fn
- needs documentation
  * [ ] update README
  * [ ] method description under adversarial detection
  * [ ] update algorithm overview
  * [ ] add drift example
- `losses.py`:
  * [ ] don't think *scale_preds* option should be there b/c then you apply the same temperature scaling to both the original and distilled model predictions
  * [ ] typos, e.g. *x_true* vs. *y_true*
  * [ ] incomplete docstrings
  * [ ] comment on K-L divergence ignores *'xent'* option
- `model_distillation.py`: 
  * [ ] unused imports
  * [ ] might be best to change *ancillary_model* to *distilled_model* as the former is not very clear/not commonly used in this context
  * [ ] *TypeError* comment incorrect, still referring to AdversarialAE
  * [ ] *# 'xent* comment in docstrings
  * [ ] maybe change *Model distillation concept drift and anomaly detector.* to *Model distillation for adversarial detection and concept drift.*
  * [ ] references to *outlier* should probably be replaced by *adversarial*
  * [ ] references to AdversarialAE/reconstruction etc need to be corrected, incl. in docstrings for various functions
  * [ ] score defaults to *kld* but should depend on `self.loss_type`
  * [ ] return dict is incorrect: *is_outlier* field needs to be *is_adversarial*
- `saving.py`
  * [ ] distilled and original models are being saved differently although they are both of the same type; best to change to consistent way of saving; can use `save_tf_model`
  * [ ] same for loading distilled and original models; can just use `load_tf_model`. Makes `load_tf_md` redundant
  * [ ] type for *ad* in `state_adv_md`  is incorrect and needs to be ModelDistillation. Docstrings also need to be corrected.
  * [ ] `self.loss_type` needs to be saved in `state_adv_md` otherwise corrected version for scoring fn will not work
  * [ ] (optional, b/c might not be relevant anymore if discrepancy between saving distilled and original model is solved) type for *detector* in `save_tf_md` is incorrect, *Union* is not needed + docstrings need to be corrected
  * [ ] `init_ad_md`: correct docstrings (*ModelDistillation* instead of *AdversarialMD*) + set `loss_type` kwarg
- CIFAR10 example
  * [ ] rename to `cd_distillation_cifar10.ipynb` to be consistent with other examples
  * [ ] shorter title otherwise won't really fit into docs well
  * [ ] intro can be similar to method description in docs, but needs references
  * [ ] fix typos
  * [ ] remove *"We also check for drift against the original test set with class imbalances."* from intro b/c is not tested I think
  * [ ] remove `%set_env CUDA_VISIBLE_DEVICES=2`
  * [ ] remove unused imports, e.g. *l1_l2, l2, Conv2DTranspose*
  * [ ] remove *ae_arch* b/c is not used in this method
  * [ ] make sure visualizations are properly scaled (e.g. [0,1] for cell 7)
  * [ ] *"Not harmful means that the prediction was unchanged after the corruption."* -> "Not harmful means that the **correct** prediction was unchanged after the corruption."

**General Q**
- has the correction mechanism from *AdversarialAE* been tested on *ModelDistillation*?

alibi_detect/ad/model_distillation.py
```diff
@@ -0,0 +1,220 @@
...
+
+        if threshold is None:
+            logger.warning('No threshold level set. Need to infer threshold using `infer_threshold`.')
+
+        self.threshold = threshold
+        self.model = model
+        for layer in self.model.layers:  # freeze model layers
+            layer.trainable = False
+
+        # check if model can be loaded, otherwise initialize AE model
```
arnaudvl, line 54: Comment not really relevant b/c model not loaded. Also is not an AE. Would just delete comment.

alibi_detect/ad/model_distillation.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+
+        # model predictions
+        y = predict_batch(self.model, X, batch_size=batch_size, proba=True)
+        y_distilled = predict_batch(self.distilled_model, X, batch_size=batch_size, proba=True)
+
+        # scale predictions
+        if self.temperature != 1.:
+            y = y ** (1 / self.temperature)  # type: ignore
+            y = y / tf.reshape(tf.reduce_sum(y, axis=-1), (-1, 1))
```
arnaudvl, line 175: Related to #138 : change to 
```python
y = (y / tf.reshape(tf.reduce_sum(y, axis=-1), (-1, 1))).numpy()
```
Otherwise returns mix of numpy arrays and tf tensors.

alibi_detect/models/losses.py
```diff
@@ -199,3 +199,49 @@ def loss_adv_ae(x_true: tf.Tensor,
...
+
+
+def loss_distillation(x_true: tf.Tensor,
+                      y_pred: tf.Tensor,
+                      model: tf.keras.Model = None,
+                      loss_type: str = 'kld',
+                      temperature: float = 1.,
+                      ) -> tf.Tensor:
+    """
+    Loss function used for AdversarialAE.
```
arnaudvl, line 211: Change to `Loss function used for Model Distillation.`

alibi_detect/models/losses.py
```diff
@@ -199,3 +199,49 @@ def loss_adv_ae(x_true: tf.Tensor,
...
+
+    Parameters
+    ----------
+    x_true
+        Batch of data points.
+    y_pred
+        Batch of prediction from the distilled model.
+    model
+        Keras model.
```
arnaudvl, line 220: `tf.keras model`. Just don't want to confuse people that only Keras can be used.

alibi_detect/utils/saving.py
```diff
@@ -1006,6 +1033,33 @@ def init_ad_ae(state_dict: Dict,
...
+               distilled_model: tf.keras.Model,
+               model: tf.keras.Model) -> ModelDistillation:
+    """
+    Initialize ModelDistillation.
+
+    Parameters
+    ----------
+    state_dict
+        Dictionary containing the parameter values.
+    ancillary_model
```
arnaudvl, line 1046: Change to `distilled_model`

alibi_detect/models/tests/test_losses.py
```diff
@@ -48,3 +48,17 @@ def test_loss_adv_ae():
...
+layers = [tf.keras.layers.InputLayer(input_shape=(x.shape[1],)),
+          tf.keras.layers.Dense(5, activation=tf.nn.softmax)]
+distilled_model = tf.keras.Sequential(layers)
+
+
+def test_loss_adv_md():
+    y_true = distilled_model(x).numpy()
+    loss_kld = loss_distillation(x, y_true, model, loss_type='kld')
+    loss_xent = loss_distillation(x, y_true, model, loss_type='xent')
+    print(loss_kld, loss_xent)
```
arnaudvl, line 62: Remove print statement.

doc/source/overview/getting_started.md
```diff
@@ -76,6 +77,7 @@ For detailed information on the **outlier detectors**:
 Similar for **adversarial detection**:
 
 * [Adversarial AE Detector](../methods/adversarialae.ipynb)
+* [Adversarial AE Detector](../methods/modeldistillation.ipynb)
```
arnaudvl, line 80: Change to
```python
* [Model Distillation Detector](../methods/modeldistillation.ipynb)
```

doc/source/methods/modeldistillation.ipynb
```diff
@@ -0,0 +1,180 @@
...
+   "metadata": {},
+   "source": [
+    "[source](../api/alibi_detect.ad.model_distillation.rst)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Model distillation adversarial and drift detector"
```
arnaudvl, line 14: To fit with the naming of the other components in the docs, please change to `Model Distillation`.

doc/source/methods/modeldistillation.ipynb
```diff
@@ -0,0 +1,180 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Overview\n",
+    "\n",
+    "[Model distillation](https://arxiv.org/abs/1503.02531) is a technique that is used to transfer knowledge from a large network to a smaller network. Tipically, it consists of training an ancillary model with a simplified architecture on soft targets (the output distributions or the logits) obtained from the original model. \n",
```
arnaudvl, line 23: Typo: "typically" and maybe replace "ancillary" with just "second"?

doc/source/methods/modeldistillation.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "\n",
+    "Here, we apply model distillation to obtain harmfulness scores, by comparing the output distributions of the original model with the output distributions \n",
+    "of the distilled model, in order to detect adversarial data, malicious data drift or data corruption.\n",
+    "We use the following definition of harmful and harmless data points:\n",
+    "* Harmful data points are defined as inputs for which the model's predictions on the uncorrupted data are correct while the model's predictions on the corrupted data are wrong.\n",
```
arnaudvl, line 28: Need an extra line for bullet points to render in readthedocs.

doc/source/methods/modeldistillation.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "* Harmless data points are defined as inputs for which the model's predictions on the uncorrupted data are correct and the model's predictions on the corrupted data remain correct.\n",
+    "\n",
+    "Analogously to the [adversarial detector](https://arxiv.org/abs/2002.09364), which is also part of the library, the model distillation detector picks up drift that reduces the performance of the classification model. \n",
+    "\n",
+    "The detector can be used as follows:\n",
+    "\n",
+    "* An adversarial score $S$ is computed. $S$ equals the loss function employed for distillation between the model predictions on $x$ and $x'$.\n",
+    "\n",
+    "* If $S$ is above a threshold (explicitly defined or inferred from training data), the instance is flagged as adversarial."
```
arnaudvl, line 37: I find the paragraph a bit confusing because it jumps back and forth between adversarial and drift detection. I think it would be helpful to emphasise that the detector component is an adversarial detector, and explain that it computes the adversarial score by computing a divergence metric like the K-L between the prediction distributions of the original and distilled model. If those prediction distributions diverge and the adversarial score is above a threshold, the instance is flagged as adversarial.

However, the adversarial scores can also be used in the context of drift detection. By computing e.g. a K-S or MMD statistic between the adversarial scores on a batch of reference data and a batch of incoming test data, malicious data drift between both batches can be detected. Then we can add the definition of "malicious", and also mention that this is very similar to how the Adversarial Auto-Encoder detector is used for drift detection and provide links to the relevant notebooks.

doc/source/methods/modeldistillation.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "* `temperature`: Temperature used for model prediction scaling. Temperature <1 sharpens the prediction probability distribution which can be beneficial for prediction distributions with high entropy.\n",
+    "\n",
+    "* `data_type`: can specify data type added to metadata. E.g. *'tabular'* or *'image'*.\n",
+    "\n",
+    "Initialized detector example:\n",
+    "\n",
+    "```python\n",
+    "from alibi_detect.ad import ModelDistillation\n",
+    "\n",
+    "ad = AdversarialAE(\n",
```
arnaudvl, line 83: Needs to be changed to `ModelDistillation`

doc/source/methods/modeldistillation.ipynb
```diff
@@ -0,0 +1,180 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Fit\n",
+    "\n",
+    "We then need to train the detector. The following parameters can be specified:\n",
+    "\n",
+    "* `X`: training batch as a numpy array.\n",
+    "\n",
+    "* `loss_fn`: loss function used for training. Defaults to the custom adversarial loss.\n",
```
arnaudvl, line 101: "Defaults to the custom model distillation loss"

doc/source/methods/modeldistillation.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "* `log_metric`: additional metrics whose progress will be displayed if verbose equals True.\n",
+    "\n",
+    "* `preprocess_fn`: optional data preprocessing function applied per batch during training.\n",
+    "\n",
+    "\n",
+    "```python\n",
+    "ad.fit(X_train, epochs=50)\n",
+    "```\n",
+    "\n",
+    "The threshold for the adversarial / harmfulness score can be set via ```infer_threshold```. We need to pass a batch of instances $X$ and specify what percentage of those we consider to be normal via `threshold_perc`. Even if we only have normal instances in the batch, it might be best to set the threshold value a bit lower (e.g. $95$%) since the the model could have misclassified training instances leading to a higher score if the reconstruction picked up features from the correct class or some instances might look adversarial in the first place.\n",
```
arnaudvl, line 120: Remove "leading to a higher score if the reconstruction picked up features from the correct class" b/c it relates to the adversarial AE.

doc/source/methods/modeldistillation.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "```"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Examples\n",
+    "\n",
+    "### Image\n",
```
arnaudvl, line 153: Maybe change to "Image drift detection" b/c it is not an adversarial detection example.

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Detecting harmuful corruption through model distillation.\n",
```
arnaudvl, line 7: Maybe change to "Model Distillation drift detector on CIFAR-10"?

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+    "\n",
+    "### Method\n",
```
arnaudvl, line 9: See comments from "Method" section.

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+    "of the distilled model, in order to detect adversarial data, malicious data drift or data corruption.\n",
+    "We use the following definition of harmful and harmless data points:\n",
+    "* Harmful data points are defined as inputs for which the model's predictions on the uncorrupted data are correct while the model's predictions on the corrupted data are wrong.\n",
+    "* Harmless data points are defined as inputs for which the model's predictions on the uncorrupted data are correct and the model's predictions on the corrupted data remain correct.\n",
+    "\n",
+    "Analogously to the [adversarial detector](https://arxiv.org/abs/2002.09364), which is also part of the library, the model distillation detector picks up drift that reduces the performance of the classification model. \n",
+    "\n",
+    "Moreover, in this example a drift detector that applies two-sample [Kolmogorov-Smirnov](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) (K-S) tests to the scores is employed. The p-values obtained are used to assess the harmufullness of the data. \n",
```
arnaudvl, line 21: Typo: "harmfulness"

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+    "import tensorflow as tf\n",
+    "from tensorflow.keras.layers import Conv2D, Dense, Flatten, InputLayer\n",
+    "from tensorflow.keras.regularizers import l1\n",
+    "from alibi_detect.cd import KSDrift\n",
+    "from alibi_detect.ad import ModelDistillation\n",
+    "\n",
+    "from alibi_detect.models.resnet import scale_by_instance\n",
+    "from alibi_detect.utils.fetching import fetch_tf_model, fetch_detector\n",
+    "from alibi_detect.utils.prediction import predict_batch\n",
+    "from alibi_detect.utils.saving import save_detector, load_detector\n",
```
arnaudvl, line 48: `load_detector` is an unused import

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+    "tags": [
+     "hide_input"
+    ]
+   },
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "Clipping input data to the valid range for imshow with RGB data ([0..1] for floats or [0..255] for integers).\n"
```
arnaudvl, line 159: *X_test* is scaled by instance, the corrupted datasets in a *[0,1]* range. They should be on the same scale and the plotted instances rescaled in [0,1].

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+    "corruption = ['gaussian_noise', 'motion_blur', 'brightness', 'pixelate']\n",
+    "X_corr, y_corr = fetch_cifar10c(corruption=corruption, severity=5, return_X_y=True)\n",
+    "X_corr = X_corr.astype('float32') / 255"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We split the original test set in a reference dataset and a dataset which should not be rejected under the *H<sub>0</sub>* of the K-S test. We also split the corrupted data by corruption type:"
```
arnaudvl, line 123: There is no reference or *H0* dataset created. Both are necessary to test the validity of the detector. If e.g. also drift occurs vs. *H0* the results on the corrupted data carry little value.

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+    "    print('{} {:.4f}'.format(corruption[_], acc))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Model distillation as a malicious drift detector\n",
+    "\n",
+    "Analogously to the [adversarial detector](https://arxiv.org/abs/2002.09364), which uses an autoencoder to reproduce the output distribution of a classifier and produce adversarial scores, the model distillation detector achieves the same goal by using a simple classifier in place of the autoencoder. This approach is more flexible respect to the autoencoder approach since it bypasses the instance's generation step, and it can be applied in a straightforward way to a variety of data sets such as text or time series.\n",
```
arnaudvl, line 283: Maybe best to refer to *Adversarial AE detector* since the *Model Distillation detector* is also an adversarial detector.

arnaudvl, line 283: Typo: "compared to" instead of "respect to"

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+    "\n",
+    "We can leverage the scores obtained trained on normal data and employ model distillation as a data drift detector. The score function of the detector becomes the preprocessing function for the drift detector. The K-S test is then a simple univariate test on the harmfulness scores. Importantly, an harmfulness detector flags **malicious data drift**. We can fetch the pretrained model distillation detector from a [Google Cloud Bucket](https://console.cloud.google.com/storage/browser/seldon-models/alibi-detect/ad/cifar10/resnet32) or train one from scratch:"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Definition and distillation of the ancillary model"
```
arnaudvl, line 292: I would not use "ancillary" or variations in different places since it means something a bit different and is by my knowledge not used in this context.

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+    "load_pretrained = True"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "filepath = '/home/fdz/.keras/datasets/' # change to (absolute) directory where model is downloaded\n",
```
arnaudvl, line 355: Best not to hardcode custom paths. Alternative: `filepath = os.path.join(os.getcwd(), 'model_distillation')`

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+   "source": [
+    "### Model distillation as a malicious drift detector\n",
+    "\n",
+    "Analogously to the [adversarial detector](https://arxiv.org/abs/2002.09364), which uses an autoencoder to reproduce the output distribution of a classifier and produce adversarial scores, the model distillation detector achieves the same goal by using a simple classifier in place of the autoencoder. This approach is more flexible respect to the autoencoder approach since it bypasses the instance's generation step, and it can be applied in a straightforward way to a variety of data sets such as text or time series.\n",
+    "\n",
+    "We can leverage the scores obtained trained on normal data and employ model distillation as a data drift detector. The score function of the detector becomes the preprocessing function for the drift detector. The K-S test is then a simple univariate test on the harmfulness scores. Importantly, an harmfulness detector flags **malicious data drift**. We can fetch the pretrained model distillation detector from a [Google Cloud Bucket](https://console.cloud.google.com/storage/browser/seldon-models/alibi-detect/ad/cifar10/resnet32) or train one from scratch:"
```
arnaudvl, line 285: "We can leverage the scores obtained trained on normal data and employ model distillation as a data drift detector. " sounds a bit confusing. Maybe clearer as "We can use the adversarial scores produced by the Model Distillation detector in the context of drift detection." and later "The K-S test is then a simple univariate test between the adversarial scores of the reference batch and the test data. Higher adversarial scores indicate more harmful drift."

Also: incorrect GCP link, this one links to a ResNet32 classifier.

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+    "\n",
+    "* We sample batches of data of size `batch_size` contaminated with an increasing number of harmful corrupted data and harmless corrupted data. \n",
+    "\n",
+    "* We apply the K-S test to these batches and obtain the p-values for each of them. \n",
+    "\n",
+    "* We observe that contamination of the batches with harmful data reduces the p-values much faster than contamination with harmless data, in which cases the p-values remains above the detection treshold even when the batch is heavily contaminated.  \n",
+    "\n",
+    "In this examples, each batch containes 100 instances where a fraction p consists of harmful or harmless corrupted data. The instances are sampled randomly from the test set and the harmfull or harmless corrupted set. For each batch a K-S test is performed with respect to a reference set of 1000 instance sampled from the uncorrupted test set. The K-S test returns a p-values which represents the probability that the batch and the reference set are sampled from the same probability distribution. \n",
```
arnaudvl, line 663: Typo: "example" and "contains"

arnaudvl, line 663: "The K-S test returns a p-values which represents the probability that the batch and the reference set are sampled from the same probability distribution."
This is not exactly right. p-values are the probability of obtaining results at least as extreme as the observed results under the assumption that the null hypothesis holds.

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+   "source": [
+    "#### Plot p-values for contaminated batches\n",
+    "\n",
+    "In order to simulate a realistic scenario, we perform a K-S test on batches of instance which are increasingly contaminated with corrupted data. The following steps are implemented:\n",
+    "\n",
+    "* We randomly pick `n_ref` samples from the non-currupted test set to be used as a reference set in the initialization of the K-S drift detector.  \n",
+    "\n",
+    "* We sample batches of data of size `batch_size` contaminated with an increasing number of harmful corrupted data and harmless corrupted data. \n",
+    "\n",
+    "* We apply the K-S test to these batches and obtain the p-values for each of them. \n",
```
arnaudvl, line 659: "We apply the K-S test to these batches and obtain the p-values for each of them."
Maybe change to:
"The K-S detector predicts whether drift occurs between the contaminated batches and the reference data and returns the p-values of the test."

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+    "\n",
+    "* We observe that contamination of the batches with harmful data reduces the p-values much faster than contamination with harmless data, in which cases the p-values remains above the detection treshold even when the batch is heavily contaminated.  \n",
```
arnaudvl, line 661: "We observe that contamination of the batches with harmful data reduces the p-values much faster than contamination with harmless data, in which cases the p-values remains above the detection treshold even when the batch is heavily contaminated."
Maybe change to:
"We observe that contamination of the batches with harmful data reduces the p-values much faster than contamination with harmless data. In the latter case, the p-values remain above the detection threshold even when the batch is heavily contaminated."

examples/cd_distillation_cifar10.ipynb
arnaudvl, line 663: This paragraph could actually be removed since the procedure is already explained just before in the bullet points. Only need to specify the nb of instances in each batch and the reference set, which can already be done in the bullet points as well.

examples/cd_distillation_cifar10.ipynb
```diff
@@ -0,0 +1,781 @@
...
+    "\n",
+    "In this examples, each batch containes 100 instances where a fraction p consists of harmful or harmless corrupted data. The instances are sampled randomly from the test set and the harmfull or harmless corrupted set. For each batch a K-S test is performed with respect to a reference set of 1000 instance sampled from the uncorrupted test set. The K-S test returns a p-values which represents the probability that the batch and the reference set are sampled from the same probability distribution. \n",
+    "\n",
+    "We repeat the test for 100 randomly sampled batches and we plot the mean and the maximum p-values for each level of severity and contamination below. We can see from the plot that the detector is able to cleary detect a batch contaminated with harmfull data with respect to a batch contaminated with harmless data when the contamination probability reaches 0.2-0.3.  "
```
arnaudvl, line 665: Typo: "to clearly"

doc/source/methods/modeldistillation.ipynb
gipster: Harmful drift detection through model distillation on CIFAR10

examples/cd_distillation_cifar10.ipynb
gipster: The reference set is created below, when the KS detector is initialized. I think it's clear from the plots at the end that a set with no contamination would not be rejected under the null hypothesis. The comment is misplaced however, I removed it

examples/cd_distillation_cifar10.ipynb
gipster: The detection is effective when the contamination probability is 0.2-0.3, which means 20-30% of the points in the batch are corrupted

PR #155, opened 2020-12-28
add warning and adjustment for odd latent_dim seq2seq

mbrner:
Odd values for `latent_dim` cause errors because of '''Bidirectional(LSTM(latent_dim // 2,...)''' in the encoder of od.Seq2Seq.

In the PR I added a warning for odd values and they are changed to the next smaller even value.


arnaudvl:
Good catch, thanks for the contribution @mbrner .

PR #156, opened 2021-01-05
Drift detection for mixed datasets - KS and Chi Square

Clusks:
Hi all,

Hope this isn't too out of place to do this - but this was brought on by my own needs after using alibi-detect! Due to using data sets that have both discrete and continuous variables in, the KS drift couldn't work on its own. So, I adapted the `feature_score()` function to allow a switch between KS and Chi-square.

Variables are automatically detected to be either discrete or continuous. If the unique count of values is < 5% of all values (unique values / size of overall sample), then a column is considered discrete.

I've inherited from the KSDrift class, as I only really wanted to overwrite one function and then add another.

Happy to make any changes to improve!

Thanks,

Rob

### DIFF
--- a/alibi_detect/cd/kschi.py
+++ b/alibi_detect/cd/kschi.py
@@ -1,0 +1,138 @@
+import logging
+import numpy as np
+from ks import KSDrift
+from scipy.stats import ks_2samp, chisquare
+from typing import Callable, Dict, Optional, Tuple, Union
+
+logger = logging.getLogger(__name__)
+
+
+class KSChiDrift(KSDrift):
+    def __init__(
+        self,
+        p_val: float = 0.05,
+        X_ref: Union[np.ndarray, list] = None,
+        preprocess_X_ref: bool = True,
+        update_X_ref: Optional[Dict[str, int]] = None,
+        preprocess_fn: Optional[Callable] = None,
+        preprocess_kwargs: Optional[dict] = None,
+        correction: str = "bonferroni",
+        alternative: str = "two-sided",
+        n_features: Optional[int] = None,
+        n_infer: int = 2,
+        input_shape: Optional[tuple] = None,
+        data_type: Optional[str] = None,
+    ) -> None:
+        """
+        Adapted version of the base Kolmogorov-Smirnov (K-S) data drift detector with Bonferroni or
+        False Discovery Rate (FDR) correction for multivariate data that allows a mix of continuous
+        and discrete variables. Continous data continues to use the K-S significance test, while
+        discrete variables use a Chi-square test.
+
+        Parameters
+        ----------
+        p_val
+            p-value used for significance of the K-S test for each feature. If the FDR correction method
+            is used, this corresponds to the acceptable q-value.
+        X_ref
+            Data used as reference distribution.
+        preprocess_X_ref
+            Whether to already preprocess and store the reference data.
+        update_X_ref
+            Reference data can optionally be updated to the last n instances seen by the detector
+            or via reservoir sampling with size n. For the former, the parameter equals {'last': n} while
+            for reservoir sampling {'reservoir_sampling': n} is passed.
+        preprocess_fn
+            Function to preprocess the data before computing the data drift metrics.
+            Typically a dimensionality reduction technique.
+        preprocess_kwargs
+            Kwargs for `preprocess_fn`.
+        correction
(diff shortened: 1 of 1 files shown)


arnaudvl:
Hi @Clusks , thanks for the welcome contribution, definitely not out of place! Drift detection on mixed data was on our minds so it's great to get a start on this. I will have a look at the PR in more detail and provide feedback.

arnaudvl:
@Clusks , I have time to pick this up by the end of next week so we can include the mixed type drift functionality in the library.

Clusks:
> 
> 
> @Clusks , I have time to pick this up by the end of next week so we can include the mixed type drift functionality in the library.

OK, great! As mentioned, happy to make any changes or do anything to help it move along.

arnaudvl:
@Clusks I am now in the process of including a separate categorical-only Chi-Square  drift detector which should be merged early next week and will then incorporate your mixed-type data detector by mid-week.

Clusks:
> @Clusks I am now in the process of including a separate categorical-only Chi-Square drift detector which should be merged early next week and will then incorporate your mixed-type data detector by mid-week.

Great to hear!

Just thinking as well, might it be best if I add in something for the output object so it tells the user whether KS or CS test was performed for each feature? As that info is made in the class, so would be easy enough to stick into the output.

arnaudvl:
Hi @Clusks , update on this: in #162 I added a separate *ChiSquareDrift* detector and let both the *KSDrift* and the *ChiSquareDrift* inherit from a *BaseUnivariateDrift* class, based on your implementation for the mixed KS and Chi-Square detector. As a result they both need to only override the `feature_score` method. This approach can be extended to other univariate statistical tests with optional multivariate correction. Then I added in the same way simply the mixed-type tabular drift detector as *TabularDrift*, where you would only need to specify the categorical feature columns in advance. This is similar to your implementation here, but inheriting form the *BaseUnivariateDrift* class and specifying a priori which features are categorical since I believe it to more prudent than inferring it based on heuristics. You could actually just use the *TabularDrift* detector if you only have numerical/categorical tabular data as well. The PR also contains an example, more detailed method descriptions and the usual out of the box `save_detector`/`load_detector` functionality. While it is 100% based on your very helpful contribution, it was easier to implement it in a separate PR. Let me know what you think!

Clusks:
Aye, I was originally thinking maybe I should let them both inherit from a single class, but didn't want to mess about too much with it since. Declaring the columns is good as well, although I think it may still benefit from the option of inference if someone wants that.

Just a question though (sorry, as I'm not too experienced with these things!), but does this mean my pull request gets closed? Would I show up as a contributor for the library at all? I was hoping this year to get more involved in these sort of things and contribute a fair bit to some of libraries I like such as alibi-detect!

arnaudvl:
@Clusks Your PR would get closed, but I can add you as a co-author in #162 via a [commit with multiple co-authors](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) to make you a contributor. I would just need your email used for your github account or the github `no-reply` email.

Clusks:
@arnaudvl Oh that would be so great! Thank you so much, you don't know how much it means! I'm already working on another idea which will hopefully be useful.

Is there a secure way to send my email over? (Or is the noreply email basically that?)

arnaudvl:
Sounds great! Thanks for providing the feature in the first place.

From what I can gather `no-reply` is exactly that: _"If a person chooses to keep their email address private, you should use their GitHub-provided no-reply email to protect their privacy. Otherwise, the co-author's email will be available to the public in the commit message. If you want to keep your email private, you can choose to use a GitHub-provided no-reply email for Git operations and ask other co-authors to list your no-reply email in commit trailers.".

arnaudvl:
Hi @Clusks , could you provide me with the relevant email because I would like to merge the PR soon. Thanks

Clusks:
Sorry, been a bit busy this morning. Give this one a go:

30694221+Clusks **at** users.noreply.github.com

Obviously replacing at with @!

arnaudvl:
Ok cool, tagged you in the commit so should hopefully be all good.

Clusks:
Thanks for doing it, although doesn't look like I'm listed as a contributor :( oh well.

arnaudvl:
@Clusks Hmm weird, let me look into it and check how I can add you after all.

Clusks:
@arnaudvl OK, keep me up to date! I might see if there's anything else I can add anyway, possibly some unit tests.

PR #162, opened 2021-01-29
Mixeddrift

arnaudvl:
Categorical and mixed type tabular drift detectors.

- [x] Chi2 drift
- [x] Mixed type tabular drift
- [x] saving/loading fn
- [x] example notebook
- [x] tests
- [x] update docs and readme

### DIFF
--- a/alibi_detect/cd/base.py
+++ b/alibi_detect/cd/base.py
@@ -78,7 +78,8 @@ class BaseUnivariateDrift(BaseDetector):
 
         # optionally already preprocess reference data
         self.preprocess_X_ref = preprocess_X_ref
-        self.X_ref = self.preprocess_fn(X_ref) if preprocess_X_ref else X_ref
+        self.X_ref = (self.preprocess_fn(X_ref) if preprocess_X_ref and isinstance(self.preprocess_fn, Callable)
+                      else X_ref)
         self.update_X_ref = update_X_ref
         self.n = X_ref.shape[0]  # type: ignore
         self.p_val = p_val
--- a/alibi_detect/cd/base.py
+++ b/alibi_detect/cd/base.py
@@ -1,3 +1,4 @@
+from abc import abstractmethod
 from functools import partial
 import logging
 import numpy as np
@@ -78,8 +79,10 @@ class BaseUnivariateDrift(BaseDetector):
 
         # optionally already preprocess reference data
         self.preprocess_X_ref = preprocess_X_ref
-        self.X_ref = (self.preprocess_fn(X_ref) if preprocess_X_ref and isinstance(self.preprocess_fn, Callable)
-                      else X_ref)
+        if preprocess_X_ref and isinstance(self.preprocess_fn, Callable):  # type: ignore
+            self.X_ref = self.preprocess_fn(X_ref)
+        else:
+            self.X_ref = X_ref
         self.update_X_ref = update_X_ref
         self.n = X_ref.shape[0]  # type: ignore
         self.p_val = p_val
@@ -129,6 +132,10 @@ class BaseUnivariateDrift(BaseDetector):
             return X_ref, X
         else:
             return self.X_ref, X
+
+    @abstractmethod
+    def feature_score(self, X_ref: np.ndarray, X: np.ndarray):
+        pass
 
     def score(self, X: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
         """
--- a/alibi_detect/cd/base.py
+++ b/alibi_detect/cd/base.py
@@ -36,7 +36,8 @@ class BaseUnivariateDrift(BaseDetector):
             p-value used for significance of the statistical test for each feature. If the FDR correction method
             is used, this corresponds to the acceptable q-value.
         X_ref
-            Data used as reference distribution.
+            Data used as reference distribution. Can be a list for text data which is then turned into an array
+            after the preprocessing step.
         preprocess_X_ref
             Whether to already preprocess and store the reference data.
         update_X_ref
--- a/alibi_detect/cd/base.py
+++ b/alibi_detect/cd/base.py
@@ -1,0 +1,207 @@
+from functools import partial
+import logging
+import numpy as np
+from typing import Callable, Dict, Optional, Tuple, Union
+from alibi_detect.base import BaseDetector, concept_drift_dict
+from alibi_detect.cd.preprocess import preprocess_drift
+from alibi_detect.cd.utils import update_reference
+from alibi_detect.utils.statstest import fdr
+
+logger = logging.getLogger(__name__)
+
+
+class BaseUnivariateDrift(BaseDetector):
+
+    def __init__(self,
+                 p_val: float = .05,
+                 X_ref: Union[np.ndarray, list] = None,
+                 preprocess_X_ref: bool = True,
+                 update_X_ref: Optional[Dict[str, int]] = None,
+                 preprocess_fn: Optional[Callable] = None,
+                 preprocess_kwargs: Optional[dict] = None,
+                 correction: str = 'bonferroni',
+                 n_features: Optional[int] = None,
+                 n_infer: int = 2,
+                 input_shape: Optional[tuple] = None,
+                 data_type: Optional[str] = None
+                 ) -> None:
+        """
+        Generic drift detector components which serves as a base class for methods using
+        univariate tests with multivariate correction.
+
+        Parameters
+        ----------
+        p_val
+            p-value used for significance of the statistical test for each feature. If the FDR correction method
+            is used, this corresponds to the acceptable q-value.
+        X_ref
+            Data used as reference distribution.
+        preprocess_X_ref
+            Whether to already preprocess and store the reference data.
+        update_X_ref
+            Reference data can optionally be updated to the last n instances seen by the detector
+            or via reservoir sampling with size n. For the former, the parameter equals {'last': n} while
+            for reservoir sampling {'reservoir_sampling': n} is passed.
+        preprocess_fn
+            Function to preprocess the data before computing the data drift metrics.
+            Typically a dimensionality reduction technique.
+        preprocess_kwargs
+            Kwargs for `preprocess_fn`.
+        correction
+            Correction type for multivariate data. Either 'bonferroni' or 'fdr' (False Discovery Rate).
+        n_features
+            Number of features used in the statistical test. No need to pass it if no preprocessing takes place.
+            In case of a preprocessing step, this can also be inferred automatically but could be more
+            expensive to compute.
+        n_infer
+            Number of instances used to infer number of features from.
+        input_shape
+            Shape of input data.
+        data_type
+            Optionally specify the data type (tabular, image or time-series). Added to metadata.
+        """
+        super().__init__()
+
+        if p_val is None:
+            logger.warning('No p-value set for the drift threshold. Need to set it to detect data drift.')
+
+        if isinstance(preprocess_kwargs, dict):  # type: ignore
+            if not isinstance(preprocess_fn, Callable):  # type: ignore
+                preprocess_fn = preprocess_drift
+            self.preprocess_fn = partial(
+                preprocess_fn,
+                **preprocess_kwargs
+            )
+            keys = list(preprocess_kwargs.keys())
+        else:
+            self.preprocess_fn, keys = None, []
+
+        # optionally already preprocess reference data
+        self.preprocess_X_ref = preprocess_X_ref
+        self.X_ref = self.preprocess_fn(X_ref) if preprocess_X_ref else X_ref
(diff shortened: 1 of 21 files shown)


doc/source/methods/chisquaredrift.ipynb
```diff
@@ -0,0 +1,141 @@
...
+   "source": [
+    "## Examples\n",
+    "\n",
+    "[Drift detection on income prediction](../examples/cd_chi2ks_adult.nblink)"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python [conda env:detect] *",
```
jklaise, line 122: I would change this to the following:
```
"display_name": "Python 3",
"language": "python",
"name": "python3"
```

Reason - the users don't have the same conda environment so on opening the notebook you're greeted with an ugly popup saying the env doesn't exist and to choose one. Setting this to the above will select the default Python 3 environment in which the notebook was openend.

This is something to do for most existing examples, would be good to open a separate issue.

examples/cd_chi2ks_adult.ipynb
```diff
@@ -0,0 +1,592 @@
+{
```
jklaise, line 1: <p>I would cleanup the outputs here as they are not relevant to the example.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

alibi_detect/cd/base.py
```diff
@@ -0,0 +1,215 @@
...
+        """
+        if isinstance(self.preprocess_fn, Callable):  # type: ignore
+            X = self.preprocess_fn(X)
+            X_ref = self.X_ref if self.preprocess_X_ref else self.preprocess_fn(self.X_ref)
+            return X_ref, X
+        else:
+            return self.X_ref, X
+
+    @abstractmethod
+    def feature_score(self, X_ref: np.ndarray, X: np.ndarray):
```
jklaise, line 137: Add return type ` -> Tuple[np.ndarray, np.ndarray]` ?

alibi_detect/cd/base.py
```diff
@@ -0,0 +1,215 @@
...
+from typing import Callable, Dict, Optional, Tuple, Union
+from alibi_detect.base import BaseDetector, concept_drift_dict
+from alibi_detect.cd.preprocess import preprocess_drift
+from alibi_detect.cd.utils import update_reference
+from alibi_detect.utils.statstest import fdr
+
+logger = logging.getLogger(__name__)
+
+
+class BaseUnivariateDrift(BaseDetector):
```
jklaise, line 14: Unfortunately, because of inheriting from this class, the sphinx API docs don't have entries for the methods `predict` etc. for the classes that inherit from this. Suggest opening a new issue to resolve this separately.

alibi_detect/cd/chisquare.py
```diff
@@ -0,0 +1,119 @@
...
+        Parameters
+        ----------
+        p_val
+            p-value used for significance of the Chi-Squared test for each feature. If the FDR correction method
+            is used, this corresponds to the acceptable q-value.
+        X_ref
+            Data used as reference distribution.
+        preprocess_X_ref
+            Whether to already preprocess and infer categories and frequencies for reference data.
+        categories_per_feature
```
jklaise, line 39: The order in the dosctring is incorrect from the method signature.

alibi_detect/cd/tabular.py
```diff
@@ -0,0 +1,130 @@
...
+        Parameters
+        ----------
+        p_val
+            p-value used for significance of the K-S and Chi2 test for each feature.
+            If the FDR correction method is used, this corresponds to the acceptable q-value.
+        X_ref
+            Data used as reference distribution.
+        preprocess_X_ref
+            Whether to already preprocess and infer categories and frequencies for categorical reference data.
+        categories_per_feature
```
jklaise, line 40: The order in the dosctring is incorrect from the method signature.

alibi_detect/cd/base.py
```diff
@@ -0,0 +1,215 @@
...
+                 preprocess_fn: Optional[Callable] = None,
+                 preprocess_kwargs: Optional[dict] = None,
+                 correction: str = 'bonferroni',
+                 n_features: Optional[int] = None,
+                 n_infer: int = 2,
+                 input_shape: Optional[tuple] = None,
+                 data_type: Optional[str] = None
+                 ) -> None:
+        """
+        Generic drift detector components which serves as a base class for methods using
```
jklaise, line 30: components --> component

alibi_detect/cd/base.py
```diff
@@ -0,0 +1,215 @@
...
+        data_type
+            Optionally specify the data type (tabular, image or time-series). Added to metadata.
+        """
+        super().__init__()
+
+        if p_val is None:
+            logger.warning('No p-value set for the drift threshold. Need to set it to detect data drift.')
+
+        if isinstance(preprocess_kwargs, dict):  # type: ignore
+            if not isinstance(preprocess_fn, Callable):  # type: ignore
```
jklaise, line 70: What are the `mypy` errors for these (and the following) annotations?

alibi_detect/cd/base.py
```diff
@@ -0,0 +1,215 @@
...
+        elif isinstance(X_ref, np.ndarray):
+            self.input_shape = X_ref.shape[1:]
+
+        # compute number of features for the univariate tests
+        if isinstance(n_features, int):
+            self.n_features = n_features
+        elif not isinstance(preprocess_fn, Callable):
+            self.n_features = X_ref.reshape(X_ref.shape[0], -1).shape[-1]
+        elif preprocess_X_ref:  # infer features from preprocessed reference data
+            self.n_features = self.X_ref.reshape(self.X_ref.shape[0], -1).shape[-1]
```
jklaise, line 104: I think we can use `self.X_ref` in both of these, even better, collapse to one branch?

alibi_detect/cd/base.py
```diff
@@ -0,0 +1,215 @@
...
+        else:  # infer number of features after applying preprocessing step
+            X = self.preprocess_fn(X_ref[0:min(X_ref.shape[0], n_infer)])
+            self.n_features = X.reshape(X.shape[0], -1).shape[-1]
```
jklaise, line 107: I think this also can use `self.X_ref` as the preprocessing would already be done in that attribute?

alibi_detect/cd/base.py
```diff
@@ -0,0 +1,215 @@
...
+        Returns
+        -------
+        Preprocessed reference data and new instances.
+        """
+        if isinstance(self.preprocess_fn, Callable):  # type: ignore
+            X = self.preprocess_fn(X)
+            X_ref = self.X_ref if self.preprocess_X_ref else self.preprocess_fn(self.X_ref)
+            return X_ref, X
+        else:
+            return self.X_ref, X
```
jklaise, line 134: I find it slightly hard to follow this logic. What is the use case for the `preprocess_X_ref` flag in `__init__` being set to `False`? If we could avoid having it as a parameter it would make the code easier to read.

alibi_detect/cd/base.py
jklaise: Ah I see in this case `X_ref` is not yet pre-processed.

alibi_detect/cd/base.py
```diff
@@ -0,0 +1,215 @@
...
+        return_distance
+            Whether to return the test statistic between the features of the new batch and reference data.
+
+        Returns
+        -------
+        Dictionary containing 'meta' and 'data' dictionaries.
+        'meta' has the model's metadata.
+        'data' contains the drift predictions and both feature and batch level drift scores.
+        'data' contains the drift prediction and optionally the feature level p-values,
+         threshold after multivariate correction if needed and test statistics.
```
jklaise, line 182: One of these lines seem redundant?

alibi_detect/cd/tabular.py
```diff
@@ -0,0 +1,130 @@
...
+                               self.categories_per_feature.items()}
+            else:
+                X_ref_count = self.X_ref_count
+        p_val = np.zeros(self.n_features, dtype=np.float32)
+        dist = np.zeros_like(p_val)
+        for f in range(self.n_features):
+            if f in list(self.categories_per_feature.keys()):
+                dist[f], p_val[f] = chisquare(X_ref_count[f], f_exp=X_count[f])
+            else:
+                dist[f], p_val[f] = ks_2samp(X_ref[:, f], X[:, f], alternative='two-sided', mode='asymp')
```
jklaise, line 129: Does it make sense to "bake in" these options here as opposed to `KSDrift` which exposes these to the user?

doc/source/methods/tabulardrift.ipynb
```diff
@@ -0,0 +1,140 @@
+{
```
jklaise, line 1: <p>Rouge <code>cat</code>  in the constructor example!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/162/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

alibi_detect/cd/base.py
arnaudvl: ` error: Argument 2 to "isinstance" has incompatible type "_SpecialForm"; expected "Union[type, Tuple[Union[type, Tuple[Any, ...]], ...]]"`

alibi_detect/cd/base.py
```diff
@@ -0,0 +1,215 @@
...
+from alibi_detect.utils.statstest import fdr
+
+logger = logging.getLogger(__name__)
+
+
+class BaseUnivariateDrift(BaseDetector):
+
+    def __init__(self,
+                 p_val: float = .05,
+                 X_ref: Union[np.ndarray, list] = None,
```
jklaise, line 18: Can this actually be a `list` (same with the child classes)? If not, suggest changing to `np.ndarray` and then we can remove some of the `type: ignore` directives when getting the shape of `X_ref` below.

alibi_detect/cd/base.py
```diff
@@ -0,0 +1,215 @@
...
+            Shape of input data.
+        data_type
+            Optionally specify the data type (tabular, image or time-series). Added to metadata.
+        """
+        super().__init__()
+
+        if p_val is None:
+            logger.warning('No p-value set for the drift threshold. Need to set it to detect data drift.')
+
+        if isinstance(preprocess_kwargs, dict):  # type: ignore
```
jklaise, line 69: Just checked, this first ignore we can remove.

alibi_detect/cd/base.py
jklaise: Upon further investigating, `Callable` is not a valid option to `isinstance` resulting in the `mypy` errors. Solution is to check `callable(preprocess_fn)` which returns a `bool`.

alibi_detect/cd/base.py
arnaudvl: Yes, for text.

alibi_detect/cd/base.py
arnaudvl: Sounds good.

alibi_detect/cd/base.py
arnaudvl: The main reason is that some preprocessing methods/steps might require access to all the reference data (eg using mean/var), so I wanted to keep things flexible.

alibi_detect/cd/base.py
jklaise: Disregard above as the issue lies with mypy and the current check is idiomatic.

alibi_detect/cd/tabular.py
jklaise: Yea I think I agree, maybe some disclaimer in the docs about how to use this feature.

alibi_detect/cd/base.py
jklaise: Is there something we can add to the docstrings about this? Is it still needed for e.g. Tabular? If no, we should remove it from the signature, if yes, we should check with `isinstance` and do the right thing which would also get rid of the `# type: ignore` directives when calling `shape`.

alibi_detect/cd/base.py
jklaise: I see, maybe worth explaining reasoning in the docs (unless I missed it?), can be a separate PR.

arnaudvl: Yes, good idea

alibi_detect/cd/tabular.py
arnaudvl: Agreed.

PR #170, opened 2021-02-15
Upper bound dependencies, add dependabot integration

jklaise:
Motivation is the same as in https://github.com/SeldonIO/alibi/pull/349 .

The only dependency I'm uncertain of is `dask`. First, not sure what the lower bound should be (I set it from what I had in my environment), second, they switched to calver (https://github.com/dask/community/issues/100) citing in part the ability to have breaking changes in releases. Now, this doesn't mean we should have an upper bound for every version as most likely there will be no breaking changes, but do we have an arbitrary upper bound (e.g. year 2022 as proposed) or just not have an upper bound for `dask`?

Finally, I cleaned up some entries in `setup.py` which are not relevant as we don't run tests using `python setup.py test` anymore (same behaviour as `alibi`).


jklaise:
We could also upper bound tensorflow at `3.0.0` if we trust them not breaking between minor versions.

PR #197, opened 2021-03-31
Add margin density drift method

ryanpmccaffrey:
**Reference Issues/PRs:**
None.

**What does this implement/fix? Explain your changes.**
Implemented new concept drift technique, MarginDensityDrift, based on measuring the percentage of total predictions made at the decision boundary of a TF/Keras probabilistic binary classifier model. 
Source: https://www.researchgate.net/publication/282542797_Don't_Pay_for_Validation_Detecting_Drifts_from_Unlabeled_data_Using_Margin_Density

**Brief summary of changes:**

- margindensity.py -- actual implementation of margin density technique
- test_margindensity.py -- unit test for new technique
- saving.py -- update saving and loading methods to accommodate margin density technique
- margindensitydrift.ipynb -- technique overview notebook
- algorithms.md/getting_started.md/README.md -- added references to new margin density technique
- cd_md3_winequality.ipynb -- example notebook showing technique on wine quality sample dataset

**Any other comments?**
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Change Log
 
+## [v0.5.1](https://github.com/SeldonIO/alibi-detect/tree/v0.5.1) (2021-03-05)
+[Full Changelog](https://github.com/SeldonIO/alibi-detect/compare/v0.5.0...v0.5.1)
+
+This is a bug fix release.
+
+### Fixed
+- The order of the reference and test dataset for the `TabularDrift` and `ChiSquareDrift` was reversed leading to incorrect test statistics
+- The implementation of `TabularDrift` and `ChiSquareDrift` were not accounting for the different sample sizes between reference and test datasets leading to incorrect test statistics
+- Bumped required `scipy` version to `1.3.0` as older versions were missing the `alternative` keyword argument for `ks_2samp` function 
+
 ## [v0.5.0](https://github.com/SeldonIO/alibi-detect/tree/v0.5.0) (2021-02-18)
 [Full Changelog](https://github.com/SeldonIO/alibi-detect/compare/v0.4.4...v0.5.0)
 ### Added
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ The following tables show the advised use cases for each algorithm. The column *
 | Chi-Squared              | ✔       | ✘     |  ✘          |  ✘    |  ✔                   |  ✘     |  ✔            |
 | Mixed-type tabular data  | ✔       | ✘     |  ✘          |  ✘    |  ✔                   |  ✘     |  ✔            |
 | Classifier               | ✔       | ✔     |  ✔          |  ✔    |  ✔                   |  ✘     |  ✘            |
-
+| Margin Density           | ✔       | ✔     |  ✘          |  ✘    |  ✘                   |  ✘     |  ✘            |
 
 ### Reference List
 
--- a/README.md
+++ b/README.md
@@ -264,8 +264,8 @@ BibTeX entry:
   title = {{Alibi-Detect}: Algorithms for outlier and adversarial instance detection, concept drift and metrics.},
   author = {Van Looveren, Arnaud and Vacanti, Giovanni and Klaise, Janis and Coca, Alexandru},
   url = {https://github.com/SeldonIO/alibi-detect},
-  version = {0.5.0},
-  date = {2021-02-18},
+  version = {0.5.1},
+  date = {2021-03-05},
   year = {2019}
 }
 ```
--- a/alibi_detect/base.py
+++ b/alibi_detect/base.py
@@ -1,5 +1,6 @@
 from abc import ABC, abstractmethod
 import copy
+import json
 import numpy as np
 from typing import Dict
 
@@ -87,3 +88,29 @@ class ThresholdMixin(ABC):
     @abstractmethod
     def infer_threshold(self, X: np.ndarray) -> None:
         pass
+
+
+class NumpyEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(
+                obj,
+                (
+                        np.int_,
+                        np.intc,
+                        np.intp,
+                        np.int8,
+                        np.int16,
+                        np.int32,
+                        np.int64,
+                        np.uint8,
+                        np.uint16,
+                        np.uint32,
+                        np.uint64,
+                ),
+        ):
+            return int(obj)
+        elif isinstance(obj, (np.float_, np.float16, np.float32, np.float64)):
+            return float(obj)
+        elif isinstance(obj, (np.ndarray,)):
+            return obj.tolist()
+        return json.JSONEncoder.default(self, obj)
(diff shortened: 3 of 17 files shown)


jklaise:
@ryanpmccaffrey thank you for the PR! We will set some time aside to review it and see if we have any follow-up questions.

Out of interest, have you used this method for any specific use cases?

ryanpmccaffrey:
My pleasure.  

Most of the drift detection methods I've come across seem to monitor P(X), the distribution of feature data.  Moreover, many of the methods that monitor P(Y|X), the distribution of labels given the feature data, seem to use performance monitoring techniques that require a feedback loop with labels (e.g., statistical process control theory).  In many cases a "closed-loop system" is not immediately possible, especially in the financial services industry (where I'm from).  For example, if we have a model that predicts whether or not someone will default on a loan, we may not have an answer to that question for 5-10 years or longer.  Even fraud cases could take weeks to adjudicate.  Meanwhile, we still want to monitor these models for drift.  This is what prompted us to explore model monitoring techniques on unlabeled data.  Furthermore, an approach that was statistical in nature (non-ML based) was actually preferred by us due to the burdens associated with re-training, internal model review processes, etc.  So, this simple approach seemed to fit the bill.  On the sample datasets and models I've applied this to (mostly non-DL models, by the way), the technique seems to be more robust to false positives as compared to some of the other virtual drift techniques out there.

Here's another paper that could be worth checking out:
https://arxiv.org/pdf/1704.00023.pdf

ojcobb:
Thanks for this @ryanpmccaffrey -- indeed it makes a lot of sense to interest ourselves in drift into and out of areas of model uncertainty as this is precisely the type of drift that is likely to cause degradation in model performance. 

However we are unsure on a few of the details of the proposed implementation and MD3 in general. In particular we believe the approach can be generalised to encapsulate multiclass classifiers and (some) regressors and also performed in a manner which returns p-values in a manner consistent with the rest of our detectors (except currently for the classifier detector but that's being updated #205) which can be thresholded more interpretably. 

We propose to instead allow users to specify how the model's uncertainty of each instance should be determined, compute the uncertainty of the model on held out validation instances and then perform a test for discrepancy in the distributions of uncertainties on the validation and test sets. 

An approach very similar to MD3 would then start by defining the notion of uncertainty as a binary flag indicating whether the model's prediction is within some margin of 0.5. A chi-squared test can then be performed to test whether a higher proportion of instances were flagged in the test set than the validation set and the corresponding p-value can be returned. 

We think a perhaps preferable alternative however might be to define the notion of uncertainty to be the entropy of the model's prediction (note how this extends to multiclass classifiers) and instead perform a KS-test on these continuous notions of uncertainty. Similarly a KS-test could be performed for regressors that return variance estimates alongside their predictions.

We believe these MD3-like generalisations can be implemented as existing detectors (KS, chi-squared etc) with custom model-dependent preprocessing steps. We have created issue #204 to implement the corresponding preprocessing functions and wrappers around detectors that call them, which I will work on next week. It would be great to hear your thoughts on whether this is likely to meet your use case. Thanks again.

ryanpmccaffrey:
Hi @ojcobb thanks for the feedback.  I'm a big fan of Seldon so it's great to see the team tackling drift that's likely to cause model performance degradation!  I like many of your proposed ideas (e.g., the use of entropy, extension to multi-class and regression problems, consistency with the other methods, etc.) and some of them we've actually thought about too.  However, it will require some further testing to truly know whether or not the proposed ideas/solutions will ultimately work for our use case.

In general, the only thing that potentially concerns me is the generation of false positives.  We have previously tried assessing drift on continuous notions of uncertainty and run into an issue with false positives.  For example, if my confidence in class A goes from 0.90 to 0.80, that may be lead to drift being detected even though my model classification performance hasn't been impacted at all (and, obviously, depending on the use case the detection of that drift might be warranted).  Here is one of the papers I was working off of:
https://arrow.tudublin.ie/cgi/viewcontent.cgi?article=1028&context=scschcomart

The other obvious potential false positive concern is not unique to the proposed (modified) MD3 implementation, but is related to the use of p-values more generally -- with large enough batches pretty much everything becomes "statistically significant", even if the effect size is negligible.  To overcome that, one suggestion could be to also implement an optional statistical measure of effect size (e.g., Cohen's d or h or something) as an additional hurdle to cross in order flag something as drift.  Otherwise I suppose just limiting the data size of each batch could also address this concern.

ojcobb:
Hi @ryanpmccaffrey -- thanks again for more thoughtful feedback!

Regarding your first point I think you are suggesting that you would only like to be sensitive to changes in model uncertainty if it leads to the model flipping its prediction. The new detectors we have implemented (#229) do in fact facilitate this by allowing the 0-1 margin notion of uncertainty with a margin width that can be specified. However, we believe changes in model uncertainty away from the decision boundary can also hamper model performance. For example suppose that for the majority of points the model predicts probabilities in [0.9, 1.0] because of strong class imbalance and suddenly a change in the feature distribution is such that suddenly the majority of predictions are falling in [0.8,0.9]. The model still predicts class 1 for these points but assumedly more of them are actually class 0 and therefore the model's performance has deteriorated and we'd like to be able to detect that.

Would I be right to summarise your second point on p-values in general terms by saying that you desire the ability to test the null d(P,Q) <= epsilon against the alternative d(P,Q) > epsilon for epsilon values other than 0 (which is what we offer)? Whilst this might be useful it certain contexts in the general case it is a more difficult test to compute p-values for (e.g. permutation tests don't apply, some test stats would become infinite mixtures). So the test statistics would have to compared directly to user-chosen thresholds, but given p-values are monotonic in test-statistics the same test can be achieved by user-chosen p-values (although the p-value interpretation would of course no longer hold if you're interested in the test for eps>0). If the test-statistic is of inherent interest note that we provide the option to return it along with the p-value. 

Let me know if I'm missing your points somehow or can provide any further feedback :slightly_smiling_face:

ryanpmccaffrey:
> However, we believe changes in model uncertainty away from the decision boundary can also hamper model performance. For example suppose that for the majority of points the model predicts probabilities in [0.9, 1.0] because of strong class imbalance and suddenly a change in the feature distribution is such that suddenly the majority of predictions are falling in [0.8,0.9]. The model still predicts class 1 for these points but assumedly more of them are actually class 0 and therefore the model's performance has deteriorated and we'd like to be able to detect that.

Agreed.  This flexibility I think will be great for users.    

I guess where I was going on the second point is by using something like a chi-squared test to generate p-values to determine statistical significance, an approach like that is highly dependent on sample size, right?  With a large enough sample (i.e., production batch in our case), even trivial relationships can appear to be statistically significant, which again would generate false positives.

Anyway, it seems like your team has already made a lot of progress in implementing some of these new ideas in a manner that's more flexible, generalizable and consistent with your other methods -- so I'd be happy to close this PR if you want to.

arnaudvl:
Hi @ryanpmccaffrey , does @ojcobb 's [answer](https://github.com/SeldonIO/alibi-detect/pull/197#issuecomment-824692715) not help for your second point? In summary: technically it is only really a false positive if you're testing for some `eps > 0`. We'll close the PR but are happy to exchange further ideas in our dedicted [slack channel](https://app.slack.com/client/T8YUF9D1Q/C01TA3HM1U5).

PR #198, opened 2021-04-01
Flexible TensorFlow and PyTorch backend for drift detectors

arnaudvl:
Allow drift detectors to work with both TensorFlow and PyTorch backends, both for the implementation of the drift detection test where applicable (eg ClassifierDrift or MMDDrift) and the preprocessing steps (eg HiddenOutput or TransformerEmbedding). This is a first step towards providing flexible backends for drift, outlier and adversarial detectors. The proposed logic can also be easily extended to other frameworks like scikit-learn or XGBoost for eg the ClassifierDrift detector. An example for a PyTorch MMD drift detector would be the following:

```python
from functools import partial
from alibi_detect.cd import MMDDrift

cd = MMDDrift(x_ref, backend='pytorch', p_val=.05)
preds = cd.predict(x_test)
```
For the TensorFlow implementation this simply becomes:

```python
cd = MMDDrift(x_ref, backend='tensorflow', p_val=.05)
```
We can also add preprocessing steps for both frameworks which would look as follows for a randomly initialized image encoder in PyTorch:

```python
import torch
import torch.nn as nn
from alibi_detect.cd.pytorch import preprocess_drift

device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')

# define encoder
encoder_net = nn.Sequential(
    nn.Conv2d(3, 64, 4, stride=2, padding=0),
    nn.ReLU(),
    nn.Conv2d(64, 128, 4, stride=2, padding=0),
    nn.ReLU(),
    nn.Conv2d(128, 512, 4, stride=2, padding=0),
    nn.ReLU(),
    nn.Flatten(),
    nn.Linear(2048, 32)
).to(device).eval()

# define preprocessing function
preprocess_fn = partial(preprocess_drift, model=encoder_net, device=device, batch_size=512)

cd = MMDDrift(x_ref, backend='pytorch', p_val=.05, preprocess_fn=preprocess_fn)
```
The same functionality is supported in TensorFlow and the main difference is that you would import `from alibi_detect.cd.tensorflow import preprocess_drift`. Other preprocessing steps such as the output of hidden layers of a model or extracted text embeddings using transformer models can be used in a similar way in both frameworks. TensorFlow example for the hidden output:

```python
from alibi_detect.cd.tensorflow import HiddenOutput, preprocess_drift

clf = # TensorFlow classifier model
preprocess_fn = partial(preprocess_drift, model=HiddenOutput(clf, layer=-1), batch_size=128)

# initialise drift detector
cd = MMDDrift(x_ref, backend='tensorflow', p_val=.05, preprocess_fn=preprocess_fn)
```
Example for text drift detection using PyTorch and transformer embeddings:

```python
import torch
import torch.nn as nn
from transformers import AutoTokenizer
from alibi_detect.cd.pytorch import preprocess_drift
from alibi_detect.models.pytorch import TransformerEmbedding

device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')

model_name = 'bert-base-cased'
tokenizer = AutoTokenizer.from_pretrained(model_name)

embedding_type = 'hidden_state'
layers = [5, 6, 7]
embed = TransformerEmbedding(model_name, embedding_type, layers)

model = nn.Sequential(embed, nn.Linear(768, 256), nn.ReLU(), nn.Linear(256, enc_dim)).to(device).eval()

preprocess_fn = partial(preprocess_drift, model=model, tokenizer=tokenizer, max_len=512, batch_size=32)

# initialise drift detector
cd = MMDDrift(x_ref, backend='pytorch', p_val=.05, preprocess_fn=preprocess_fn)
```
Again the same functionality is supported in TensorFlow but with `from alibi_detect.cd.tensorflow import preprocess_drift` and `from alibi_detect.models.pytorch import TransformerEmbedding` imports.

One more example for the classifier drift detector:

```python
import tensorflow as tf
from tensorflow.keras.layers import Conv2D, Dense, Flatten, Input
from alibi_detect.cd import ClassifierDrift

model = tf.keras.Sequential(
  [
      Input(shape=(32, 32, 3)),
      Conv2D(8, 4, strides=2, padding='same', activation=tf.nn.relu),
      Conv2D(16, 4, strides=2, padding='same', activation=tf.nn.relu),
      Conv2D(32, 4, strides=2, padding='same', activation=tf.nn.relu),
      Flatten(),
      Dense(2, activation='softmax')
  ]
)

cd = ClassifierDrift(x_ref, model, threshold=.55, backend='tensorflow')
preds = cd.predict(x_test)
```
while a PyTorch version would look like this:

```python
import torch
import torch.nn as nn

device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')

model = nn.Sequential(
    nn.Conv2d(3, 8, 4, stride=2, padding=0),
    nn.ReLU(),
    nn.Conv2d(8, 16, 4, stride=2, padding=0),
    nn.ReLU(),
    nn.Conv2d(16, 32, 4, stride=2, padding=0),
    nn.ReLU(),
    nn.Flatten(),
    nn.Linear(128, 2)
)

cd = ClassifierDrift(x_ref, model, threshold=.55, backend='pytorch')
preds = cd.predict(x_test)
```

The changes require duplication of functionality for both TensorFlow and PyTorch in various places of the package, adding different `tensorflow` and `pytorch` modules:

* Drift

```python
alibi_detect.cd.pytorch
alibi_detect.cd.tensorflow
```
contains the backend specific implementations of the MMDDrift, ClassifierDrift and various supported preprocessing steps. MMDDrift and ClassifierDrift which are imported as `from alibi_detect.cd import MMDDrift, ClassifierDrift` dispatch to the relevant backend.

* Models

```python
alibi_detect.models.pytorch
alibi_detect.models.tensorflow
```
contains backend specific training and model architecture functionality among other things. For instance the transformer embeddings, autoencoder architectures, PixelCNN++ etc are here as well. PyTorch has not reached parity with TensorFlow here since it contains a lot of outlier detection specific functionality.

* Utils

```python
alibi_detect.utils.pytorch
alibi_detect.utils.tensorflow
```
contains backend specific utlity functions such as distance computations, kernels and batch prediction functions for PyTorch and TensorFlow models, incl. for transformers.

**TO DO**

* [ ] update save and load functionality for both backends
* [ ] update tests for TensorFlow and add for PyTorch
* [ ] make sure nothing broke due to the changes, incl. all examples
* [ ] update docs and README, add flag which methods support which backend


cliveseldon:
Looks good. Initial questions:
 * Can the preprocess function backend differ from the man drift detector or do they need to match?
 * Related to above would one install `alibi-detect[pytorch]` or `alibi-detect[tensorflow]` to just get one of the two frameworks or will the package always have both?

arnaudvl:
> * Can the preprocess function backend differ from the man drift detector or do they need to match?

Although I don't imagine this to be a very likely use case, they could differ. So for instance for the MMDDrift detector you could use a TensorFlow preprocess step and a PyTorch MMD implementation. In that case you could do something like

```python
from alibi_detect.cd.tensorflow import HiddenOutput, preprocess_drift

model = # TensorFlow model; tf.keras.Model or tf.keras.Sequential
preprocess_fn = partial(preprocess_drift, model=HiddenOutput(model, layer=-1), batch_size=128)

# initialise drift detector
cd = MMDDrift(x_ref, backend='pytorch', p_val=.05, preprocess_fn=preprocess_fn)
```

> * Related to above would one install `alibi-detect[pytorch]` or `alibi-detect[tensorflow]` to just get one of the two frameworks or will the package always have both?

Currently most of the non-drift related functionality (outlier and adversarial detection) uses TensorFlow. As a result, it would make sense to keep it as a default requirement for now. Need to discuss with @jklaise whether `pip install alibi-detect[pytorch]` would be an acceptable current solution. As we get closer to feature parity in both frameworks I imagine we can provide `alibi-detect[pytorch]` and `alibi-detect[tensorflow]` options.

ojcobb:
Nice work! Some minor comments:

- I'm currently unsure about passing `sigma` and `infer_sigma` to `MMDDrift` objects which in general take kernels that aren't the GaussianRBF. Can't we instead pass `kernel_hps` and `infer_kernel_hps` or similar and for the `GaussianRBF` the `hp` to be inferred would indeed be sigma? If left as is it might at least be worth making clearer that those arguments are only relevant when `kernel=GaussianRBF`.
- Separately from the above, I'm unsure that `infer_sigma` implements the behaviour users might expect. If I understand correctly `infer_sigma=True` configures the kernel using the reference data only whereas `infer_sigma=False` configures the kernel using both reference and test data for each test. In both cases sigma is being inferred and users desiring the latter might accidentally end up with the former.
- When passing multiple sigma values can we take an average rather than sum? It obviously won't make a different in most contexts but there are contexts in which it's useful to know the maximum value the kernel can take (for bounds) and this would change depending on how many sigma values are passed.
- Is there a rationale for the switch from X notation to x notation in some places and not others?

jklaise:
Related to @ojcobb comment on renaming `X -> x`, we should double check that there is consistency between different methods on how arguments are passed, e.g. allowing the use of positional arguments for required arguments they should always be in the same order. Additionally, for optional kwargs worth considering if we want to make them kwargs only so they can't be specified positionally (and hence wouldn't depend on the order).

alibi_detect/utils/pytorch/prediction.py
```diff
@@ -0,0 +1,104 @@
+import numpy as np
+import torch
+import torch.nn as nn
+from typing import Callable, Union
+
+
+def predict_batch(x: Union[np.ndarray, torch.Tensor], model: Union[nn.Module, nn.Sequential],
+                  device: torch.device = None, batch_size: int = int(1e10),
+                  dtype: type = np.float32) -> Union[np.ndarray, torch.Tensor]:
```
jklaise, line 9: I don't think `type` is the correct typehint, because `torch.float32` and `tf.float32` are not of type `type`, can't we use `Union[np.float32, torch.float32]` (can make a type alias at the top to make it more succint if more floating point types are to be supported).

alibi_detect/utils/tensorflow/prediction.py
```diff
@@ -0,0 +1,81 @@
+import numpy as np
+import tensorflow as tf
+from typing import Union
+
+
+def predict_batch(x: Union[np.ndarray, tf.Tensor], model: tf.keras.Model, batch_size: int = int(1e10),
+                  dtype: type = np.float32) -> Union[np.ndarray, tf.Tensor]:
```
jklaise, line 7: Ditto as for `pytorch/prediction.py`.

alibi_detect/utils/tests/test_prediction.py
```diff
@@ -3,7 +3,7 @@
 import tensorflow as tf
 from tensorflow.keras.layers import Dense, InputLayer
 from alibi_detect.utils.prediction import predict_batch
-from alibi_detect.models.autoencoder import AE
+from alibi_detect.models.tensorflow.autoencoder import AE
```
jklaise, line 6: I believe this file is a duplicate of `tensorflow/test_prediction.py` and can be deleted.

alibi_detect/utils/tests/test_distance.py
```diff
@@ -1,10 +1,8 @@
-import dask.array as da
 import numpy as np
 from scipy.spatial.distance import cityblock
 from itertools import product
 import pytest
-from alibi_detect.utils.distance import (pairwise_distance, maximum_mean_discrepancy, abdm,
-                                         cityblock_batch, mvdm, multidim_scaling, relative_euclidean_distance)
+from alibi_detect.utils.distance import pairwise_distance, abdm, cityblock_batch, mvdm, multidim_scaling
```
jklaise, line 5: Are the `dask` array tests not really required anymore?

arnaudvl: No, `dask` is no longer supported.

alibi_detect/utils/statstest.py
```diff
@@ -1,17 +1,9 @@
...
 
-def permutation_test(x: Union[np.ndarray, da.array],
-                     y: Union[np.ndarray, da.array],
-                     n_permutations: int = 1000,
-                     metric: Callable = maximum_mean_discrepancy,
-                     return_distance: bool = False,
-                     return_permutation_distance: bool = False,
-                     **kwargs) \
-        -> Union[np.float, Tuple[np.float, np.float], Tuple[np.float, np.float, np.ndarray]]:
+def permutation_test(x: np.ndarray, y: np.ndarray, metric: Callable, n_permutations: int = 100,
```
jklaise, line 5: Why remove `dask` type hints? More generally, if it's true that all of our `numpy` functions would work seamlessly with `dask` arrays, we could not bother with `dask` typehints, especially if users never explicitly deal with `dask` arrays.

alibi_detect/utils/tests/test_prediction.py
arnaudvl: It's not a duplication, test another prediction function used within the outlier detectors. I want to address this in a follow up PR to unify both.

jklaise:
@arnaudvl re `dask`, if this is no longer supported take a note of this for the changelog and update `setup.py`. There are some places where `dask` is still mentioned in the codebase (e.g. in type hints), so I recommend running text search on the codebase for `dask`.

alibi_detect/cd/classifier.py
```diff
@@ -1,215 +1,72 @@
...
-            Optionally specify the data type (tabular, image or time-series). Added to metadata.
-        """
+if has_pytorch:
+    from alibi_detect.cd.pytorch.classifier import ClassifierDriftTorch
+
+if has_tensorflow:
+    from alibi_detect.cd.tensorflow.classifier import ClassifierDriftTF
+
+
+class ClassifierDrift:
```
jklaise, line 13: We should probably still make this inherit from `BaseDetector` to adhere to a well-defined interface, in particular it should have a `meta` property (which can just dispatch to the `_detector.meta`. I note that the `score` method in `BaseDetector` may not apply here (as the backend takes care of it), does it make sense to remove `score` from `BaseDetector` entirely then?

alibi_detect/cd/mmd.py
```diff
@@ -1,173 +1,70 @@
...
-        n_permutations
-            Number of permutations used in the permutation test.
-        chunk_size
-            Chunk size if dask is used to parallelise the computation.
-        input_shape
-            Shape of input data.
-        data_type
-            Optionally specify the data type (tabular, image or time-series). Added to metadata.
-        """
+class MMDDrift:
```
jklaise, line 15: Ditto regarding inheritance as for `ClassifierDrift` comment.

alibi_detect/cd/mmd.py
```diff
@@ -1,173 +1,70 @@
...
+            update_x_ref: Optional[Dict[str, int]] = None,
+            preprocess_fn: Optional[Callable] = None,
+            kernel: Callable = None,
+            sigma: Optional[np.ndarray] = None,
+            infer_sigma: bool = True,
+            n_permutations: int = 100,
+            device: Optional[str] = None,
+            input_shape: Optional[tuple] = None,
+            data_type: Optional[str] = None
+    ) -> None:
```
jklaise, line 32: Add docstring.

alibi_detect/cd/classifier.py
```diff
@@ -1,215 +1,72 @@
...
+            optimizer: Optional = None,
+            learning_rate: float = 1e-3,
+            compile_kwargs: Optional[dict] = None,
+            batch_size: int = 32,
+            epochs: int = 3,
+            verbose: int = 0,
+            train_kwargs: Optional[dict] = None,
+            device: Optional[str] = None,
+            data_type: Optional[str] = None
+    ) -> None:
```
jklaise, line 38: Add docstring.

jklaise:
Leaving a note that test models seem duplicated in several test files, probably best to factor these out into `conf.py` as fixtures in a follow-up PR.

examples/cd_chi2ks_adult.ipynb
```diff
@@ -25,7 +25,17 @@
    "cell_type": "code",
```
jklaise, line 25: <p>Nitpick: would be good to rerun these sorts of cells without <code>prophet</code>  installed to remove the ugly error in our docs.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

examples/cd_clf_cifar10.ipynb
```diff
@@ -10,6 +10,10 @@
     "\n",
```
jklaise, line 10: <p>Markdown issue rendering title.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

examples/cd_text_imdb.ipynb
```diff
@@ -6,20 +6,23 @@
    "source": [
```
jklaise, line 6: <p><code>The method works...</code>  sentence rendered as a title currently.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

arnaudvl:
> Nice work! Some minor comments:
> 
> * I'm currently unsure about passing `sigma` and `infer_sigma` to `MMDDrift` objects which in general take kernels that aren't the GaussianRBF. Can't we instead pass `kernel_hps` and `infer_kernel_hps` or similar and for the `GaussianRBF` the `hp` to be inferred would indeed be sigma? If left as is it might at least be worth making clearer that those arguments are only relevant when `kernel=GaussianRBF`.

Yes, `kernel_hps` or `kernel_kwargs` sounds good as a more general solution but with a caveat (see next answer).

> * Separately from the above, I'm unsure that `infer_sigma` implements the behaviour users might expect. If I understand correctly `infer_sigma=True` configures the kernel using the reference data only whereas `infer_sigma=False` configures the kernel using both reference and test data for each test. In both cases sigma is being inferred and users desiring the latter might accidentally end up with the former.

Not sure if it is entirely separate. `infer_sigma` is a kwarg from the kernel computation step (from the kernel object's `__call__`). As a result any custom kernel would need to implement this as well. Therefore `infer_kernel_hps` might be redundant and rephrased as `configure_kernel_from_x_ref` (or something more concise...). If set to True, sigma is inferred from the reference data, otherwise from `(x_ref, x_test)`. Of course `infer_sigma` is not general for all kernels, but then we need to make some choices on the form a custom kernel can take. This could be addressed in a follow up PR.

> * When passing multiple sigma values can we take an average rather than sum? It obviously won't make a different in most contexts but there are contexts in which it's useful to know the maximum value the kernel can take (for bounds) and this would change depending on how many sigma values are passed.

Sounds good.

> * Is there a rationale for the switch from X notation to x notation in some places and not others?

It will be changed everywhere.

alibi_detect/cd/classifier.py
arnaudvl: Agreed to address in follow up PR.

arnaudvl:
@jklaise Came across some tricky type annotations, would be good to check the `type: ignore`'s in the [latest commit](https://github.com/SeldonIO/alibi-detect/pull/198/commits/d55a7e4af63b6884261cdccaa0634cfbd4f49962)

README.md
```diff
@@ -82,35 +81,70 @@ The following tables show the advised use cases for each algorithm. The column *
...
+cd = MMDDrift(x_ref, backend='tensorflow', p_val=.05)
+preds = cd.predict(x)
+```
+
+The same detector in PyTorch:
+
+```python
+cd = MMDDrift(x_ref, backend='pytorch', p_val=.05)
+preds = cd.predict(x)
+```
```
jklaise, line 129: Add a sentence saying that `pytorch` should be installed separately, same for examples when talking about `pytorch`.

README.md
```diff
@@ -82,35 +81,70 @@ The following tables show the advised use cases for each algorithm. The column *
...
-| Mahalanobis Distance  | ✔       | ✘     |  ✘          |  ✘    |  ✔                   |  ✔     |  ✘            |
-| AE                    | ✔       | ✔     |  ✘          |  ✘    |  ✘                   |  ✘     |  ✔            |
-| VAE                   | ✔       | ✔     |  ✘          |  ✘    |  ✘                   |  ✘     |  ✔            |
-| AEGMM                 | ✔       | ✔     |  ✘          |  ✘    |  ✘                   |  ✘     |  ✘            |
-| VAEGMM                | ✔       | ✔     |  ✘          |  ✘    |  ✘                   |  ✘     |  ✘            |
-| Likelihood Ratios     | ✔       | ✔     |  ✔          |  ✘    |  ✔                   |  ✘     |  ✔            |
-| Prophet               | ✘       | ✘     |  ✔          |  ✘    |  ✘                   |  ✘     |  ✘            |
-| Spectral Residual     | ✘       | ✘     |  ✔          |  ✘    |  ✘                   |  ✔     |  ✔            |
-| Seq2Seq               | ✘       | ✘     |  ✔          |  ✘    |  ✘                   |  ✘     |  ✔            |
+| Isolation Forest      | ✔       |      |            |      |  ✔                   |       |              |
```
jklaise, line 84: Check the errors for all tables from the docs building process as there are some mismatches of row lengths it seems.

doc/source/overview/algorithms.md
```diff
@@ -6,30 +6,35 @@ The following tables summarize the advised use cases for the current algorithms.
...
-|[Mahalanobis Distance](../methods/mahalanobis.ipynb)|✔|✘|✘|✘|✔|✔|✘|
-|[AE](../methods/ae.ipynb)|✔|✔|✘|✘|✘|✘|✔|
-|[VAE](../methods/vae.ipynb)|✔|✔|✘|✘|✘|✘|✔|
-|[AEGMM](../methods/aegmm.ipynb)|✔|✔|✘|✘|✘|✘|✘|
-|[VAEGMM](../methods/vaegmm.ipynb)|✔|✔|✘|✘|✘|✘|✘|
-|[Likelihood Ratios](../methods/llr.ipynb)|✔|✔|✔|✘|✔|✘|✔|
-|[Prophet](../methods/prophet.ipynb)|✘|✘|✔|✘|✘|✘|✘|
-|[Spectral Residual](../methods/sr.ipynb)|✘|✘|✔|✘|✘|✔|✔|
-|[Seq2Seq](../methods/seq2seq.ipynb)|✘|✘|✔|✘|✘|✘|✔|
+|[Isolation Forest](../methods/iforest.ipynb)|✔||||✔|||
```
jklaise, line 9: Check length of rows of these tables as per sphinx errors when building docs.

jklaise (approved):

Issue #193, opened 2021-03-26
Categorical drift detection with unseen categories


arnaudvl:
#191

arnaudvl:
Addressed by #222

PR #203, opened 2021-04-07
Switch from GOF test to homogeneity test when using chi-squared test statistic

ojcobb:
Currently for each categorical feature the chi-squared drift detector performs a goodness of fit test of the observed data against the empirical distribution of the reference data. This biases p-values towards extremes because it doesn't allow for the possibility that an underlying distribution somewhere between the empirical distributions of the observed and reference data generated both. Instead a test for homogeneity should be performed. This can be thought of as instead identifying the distribution most likely to have generated both and performing a goodness-of-fit test on that. 

The difference should only be slight for large reference sets.

### DIFF
--- a/alibi_detect/cd/chisquare.py
+++ b/alibi_detect/cd/chisquare.py
@@ -1,6 +1,6 @@
 import logging
 import numpy as np
-from scipy.stats import chisquare
+from scipy.stats import chi2_contingency
 from typing import Callable, Dict, Optional, Tuple, Union
 from alibi_detect.cd.base import BaseUnivariateDrift
 
@@ -112,8 +112,8 @@ class ChiSquareDrift(BaseUnivariateDrift):
         p_val = np.zeros(self.n_features, dtype=np.float32)
         dist = np.zeros_like(p_val)
         for f in range(self.n_features):  # apply Chi-Squared test
-            n_ref, n_obs = X_ref_count[f].sum(), X_count[f].sum()
-            dist[f], p_val[f] = chisquare(X_count[f], f_exp=X_ref_count[f] * n_obs / n_ref)
+            contingency_table = np.vstack((X_ref_count[f], X_count[f]))
+            dist[f], p_val[f], _, _ = chi2_contingency(contingency_table)
         return p_val, dist
 
     def _get_counts(self, X: np.ndarray) -> Dict[int, np.ndarray]:
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ __pycache__/
 # pycharm
 .idea/
 
+# vscode
+.vscode
+
 # C extensions
 *.so
 
--- a/alibi_detect/cd/tabular.py
+++ b/alibi_detect/cd/tabular.py
@@ -86,7 +86,7 @@ class TabularDrift(BaseUnivariateDrift):
             # infer number of possible categories for each categorical feature from reference data
             if None in list(categories_per_feature.values()):
                 X_flat = self.X_ref.reshape(self.X_ref.shape[0], -1)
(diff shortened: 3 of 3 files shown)


jklaise (approved):

alibi_detect/cd/chisquare.py
```diff
@@ -112,8 +112,8 @@ def feature_score(self, X_ref: np.ndarray, X: np.ndarray) -> Tuple[np.ndarray, n
         p_val = np.zeros(self.n_features, dtype=np.float32)
         dist = np.zeros_like(p_val)
         for f in range(self.n_features):  # apply Chi-Squared test
-            n_ref, n_obs = X_ref_count[f].sum(), X_count[f].sum()
-            dist[f], p_val[f] = chisquare(X_count[f], f_exp=X_ref_count[f] * n_obs / n_ref)
+            contingency_table = np.vstack((X_ref_count[f], X_count[f]))
+            dist[f], p_val[f], _, _ = chi2_contingency(contingency_table)
```
arnaudvl, line 116: Leaving `lambda_=None` applies Pearson's Chi^2 test. Do we think it is worth allowing for other tests as well by specifying `lambda_` at init? See https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.chi2_contingency.html.

jklaise (approved):

alibi_detect/cd/chisquare.py
ojcobb: Just looked into this a little and after looking at [this paper](https://hal.archives-ouvertes.fr/hal-00514341/document) I'm not sure that it's worth the added complexity. Pearson's generally considered best and also the conditions under which it suitable and accurate is best understood. Speaking of which, I wonder if it might be worth adding warnings when such conditions (see p2 of aforementioned paper) are not met?

ojcobb:
I believe also this fix partially addresses #193.

I believe previous divide-by-0 issues were a result of wrongly applying the GOF test rather than test of homogeneity. In the latter the expected counts are based on observations in both the reference and test sets and therefore any category observed in the reference and/**or** test sets can be included in the stat without introducing a divide-by-0. In fact such observations are strongly suggestive of drift and therefore important to handle correctly. 

A change to the implementation would still be required for #193 though to add a 0 count for the newly observed category to the reference counts ahead of the test. Would you like me to address this @arnaudvl?

alibi_detect/cd/chisquare.py
arnaudvl: Good idea. Might even be worth returning it as well with the drift prediction? Could be helpful for downstream applications.

arnaudvl:
> I believe also this fix partially addresses #193.
> 
> I believe previous divide-by-0 issues were a result of wrongly applying the GOF test rather than test of homogeneity. In the latter the expected counts are based on observations in both the reference and test sets and therefore any category observed in the reference and/**or** test sets can be included in the stat without introducing a divide-by-0. In fact such observations are strongly suggestive of drift and therefore important to handle correctly.
> 
> A change to the implementation would still be required for #193 though to add a 0 count for the newly observed category to the reference counts ahead of the test. Would you like me to address this @arnaudvl?

Yes it would nice to kill 2 birds with 1 stone.

alibi_detect/cd/chisquare.py
jklaise: The warning sounds like a nice idea (provided we link to justification in the docs) and agree that makes sense to include this info in the return too. I would make it a separate PR though to keep things more atomic.

A more general conversation we can have in the future is if we should support more tests as there are a plethora of frequentist tests in the literature.

jklaise:
@ojcobb @arnaudvl I would implement the functionality to update reference with unseen categories in a new PR to keep the history atomic. Otherwise I think we've addressed everything this PR was set out to do.

Issue #205, opened 2021-04-09
Transition domain classifier-detector from thresholds to p-values


ojcobb:
Rather than allowing arbitrary performance metrics and comparing against a specified threshold we should restrict to either accuracy (hard) or logit (soft) performance indicators and perform either Binomial (as in [Failing Loudly](https://arxiv.org/abs/1810.11953)) or KS tests respectively to return a p-value.

PR #206, opened 2021-04-09
Transition classifier drift-detector from thresholds to p-values

ojcobb:
Addresses #205.

Instantiation now looks like
```python
cd = ClassifierDrift(p_val=.05, model=model, X_ref=X_ref, train_size=.75, epochs=1)
```
which by default performs a KS-test on the classifier's probs or log-probs (either fine) directly. If for some reason a user wishes to binarise the model predictions and apply a binomial test (as in [Failing Loudly](https://arxiv.org/abs/1810.11953)) they can do this by passing the argument `soft_preds=False`. I can only see this lowering the power of the test however so I don't recommend it and haven't demonstrated it in the notebook.


alibi_detect/cd/classifier.py
```diff
@@ -188,41 +182,60 @@ def score(self, X: Union[np.ndarray, list]) -> float:
...
-        idx_oof = np.concatenate(idx_oof, axis=0)
-        drift_metric = self.metric_fn(y[idx_oof], preds_oof)
-        return drift_metric
-
-    def predict(self, X: Union[np.ndarray, list], return_metric: bool = True) \
-            -> Dict[Dict[str, str], Dict[str, Union[int, float]]]:
+            preds_oof_list.append(preds)
+            idx_oof_list.append(idx_te)
+        preds_oof = np.array(np.concatenate(preds_oof_list, axis=0)[:, 1])
+        idx_oof = np.array(np.concatenate(idx_oof_list, axis=0))
```
jklaise, line 195: Would replace `np.array` with `np.asarray` as the latter will not make a copy if possible, e.g. https://stackoverflow.com/questions/14415741/what-is-the-difference-between-numpys-array-and-asarray-functions

alibi_detect/cd/classifier.py
```diff
@@ -188,41 +182,60 @@ def score(self, X: Union[np.ndarray, list]) -> float:
...
+
+        if self.soft_preds:
+            log_losses_ref = preds_oof[y[idx_oof] == 0]
+            log_losses_cur = preds_oof[y[idx_oof] == 1]
+            dist, p_val = ks_2samp(log_losses_ref, log_losses_cur, alternative='greater')
+        else:
+            baseline_accuracy = max(X_ref.shape[0], X.shape[0]) / (X_ref.shape[0] + X.shape[0])  # exp under null
+            n_oof = idx_oof.shape[0]
+            n_correct = (y[idx_oof] == preds_oof.round()).sum()
+            p_val = binom_test(n_correct, n_oof, baseline_accuracy, alternative='greater')
```
jklaise, line 205: Is `greater` the correct alternative to use? e.g. accuracy may either increase or decrease in face of drift.

ojcobb: Yeah 'greater' is the correct direction. Note here that we're not looking at the accuracy of a label classifier (which indeed might go up or down under drift) but of a domain classifier trained to classify ref/test instances. If it does this worse than the baseline, that it random chance, not because of drift. Only doing better than a baseline (i.e. there was something available to learn) is indicative of drift.

alibi_detect/cd/classifier.py
```diff
@@ -188,41 +182,60 @@ def score(self, X: Union[np.ndarray, list]) -> float:
...
+            accuracy = n_correct/n_oof
+            # relative error reduction, in [0,1]
+            # e.g. (90% acc -> 99% acc) = 0.9, (50% acc -> 59% acc) = 0.18
+            dist = 1 - (1 - accuracy)/(1-baseline_accuracy)
+            dist = max(0, dist)  # below 0 = no evidence for drift
```
jklaise, line 210: Related to the above, what if accuracy decreases wrt the baseline accuracy?

alibi_detect/cd/classifier.py
ojcobb: Oh yeah I think that can be removed now actually as I took another route to satisfy mypy.

alibi_detect/cd/classifier.py
jklaise: ok makes sense, I guess the underlying assumption is that the classifier is always at least as good as a dummy one.

alibi_detect/cd/classifier.py
```diff
@@ -188,41 +182,60 @@ def score(self, X: Union[np.ndarray, list]) -> float:
...
         Predict whether a batch of data has drifted from the reference data.
 
         Parameters
         ----------
         X
             Batch of instances.
-        return_metric
-            Whether to return the drift metric from the detector.
+        return_p_val
+            Whether to return the p-value of the test.
```
jklaise, line 224: Don't all drift detectors return a p-value by default @arnaudvl ? If so we can dispense with this additional kwarg.

alibi_detect/cd/classifier.py
ojcobb: I'm not sure we even need to assume that. Only that the classifier is capable of outperforming a dummy one when it is theoretically possible (i.e. when drift has occured).

jklaise: Oh yes that's what I meant.

examples/cd_clf_cifar10.ipynb
```diff
@@ -18,9 +18,17 @@
   },
```
jklaise, line 18: <p>Any reason to do just one type of corruption? Should update the accompanying text and delete the commented line if so.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/206/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

jklaise, line 18: <p>Missing end of first sentence!</p><p> </p><p>Also need to change "accuracy threshold set at 0.55" to "a p-value/significance level set at 0.05" or similar.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/206/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

jklaise, line 18: <p>I think this whole section is now defunct and should be removed?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/206/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

jklaise:
@ojcobb nice work! I think there's only a few things that need to be cleaned up in the example notebook, I left comments there. Otherwise @arnaudvl will need to check what other changes will be needed once #198 lands.

examples/cd_clf_cifar10.ipynb
ojcobb: Will fix.

ojcobb: Sorry I've just realised the wrong notebook is currently uploaded for some reason.. Hence the errors and things not changed. Let me just push the one that should be there

ojcobb: Ok pushed now. I hadn't clicked 'save notebook' before.

ojcobb:
I've merged in the flexible backend changes @arnaudvl

alibi_detect/cd/pytorch/classifier.py
```diff
@@ -149,9 +145,24 @@ def score(self, x: np.ndarray) -> float:
...
-            idx_oof.append(idx_te)
-        preds_oof = np.concatenate(preds_oof, axis=0).argmax(axis=1)
-        idx_oof = np.concatenate(idx_oof, axis=0)
-        drift_metric = self.metric_fn(y[idx_oof], preds_oof)
-        return drift_metric
+            preds_oof_list.append(preds)
+            idx_oof_list.append(idx_te)
+        preds_oof = np.concatenate(preds_oof_list, axis=0).argmax(axis=1)
+        idx_oof = np.concatenate(idx_oof_list, axis=0)
+
```
arnaudvl, line 152: From here on it looks like the functionality is framework-agnostic and the same as e.g. the TensorFlow one. Could be a good idea to add this as a function of the base method?

alibi_detect/cd/tensorflow/classifier.py
```diff
@@ -120,22 +115,38 @@ def score(self, x: np.ndarray) -> float:
...
-        idx_oof = np.concatenate(idx_oof, axis=0)
-        drift_metric = self.metric_fn(y[idx_oof], preds_oof)
-        return drift_metric
+            preds_oof_list.append(preds)
+            idx_oof_list.append(idx_te)
+        preds_oof = np.concatenate(preds_oof_list, axis=0)[:, 1]
+        idx_oof = np.concatenate(idx_oof_list, axis=0)
+
+        if self.soft_preds:
+            log_losses_ref = preds_oof[y[idx_oof] == 0]
```
arnaudvl, line 138: Maybe need to be careful here as, contrary to PyTorch, TensorFlow models often return softmax output instead of logits.

alibi_detect/cd/pytorch/classifier.py
```diff
@@ -149,9 +145,24 @@ def score(self, x: np.ndarray) -> float:
...
+        if self.soft_preds:
+            log_losses_ref = preds_oof[y[idx_oof] == 0]
+            log_losses_cur = preds_oof[y[idx_oof] == 1]
+            dist, p_val = ks_2samp(log_losses_ref, log_losses_cur, alternative='greater')
+        else:
+            baseline_accuracy = max(x_ref.shape[0], x.shape[0]) / (x_ref.shape[0] + x.shape[0])  # exp under null
+            n_oof = idx_oof.shape[0]
+            n_correct = (y[idx_oof] == preds_oof.round()).sum()
```
arnaudvl, line 160: Don't think this needs to be rounded as we already take the argmax earlier?

alibi_detect/cd/pytorch/classifier.py
```diff
@@ -149,9 +145,24 @@ def score(self, x: np.ndarray) -> float:
             train_args = [model, nn.CrossEntropyLoss(), dl_tr, self.device]
```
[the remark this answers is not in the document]
ojcobb: Could allow it to be specified?

alibi_detect/cd/tensorflow/classifier.py
ojcobb: The KS-test stat will remain the same for any monotonic transformation of the data (so doesn't matter if probs, log-probs, logits etc) but maybe I should use more general name to avoid confusion.

alibi_detect/cd/pytorch/classifier.py
ojcobb: I don't think we do?

ojcobb: The mistake is in taking the argmax earlier. I'd removed it for tf but not pt. Thanks for catching.

alibi_detect/cd/tensorflow/classifier.py
ojcobb: Actually I'm not sure if it's a great idea to apply it directly to logits. I'm going to change to ensure it's applied to probs

alibi_detect/cd/pytorch/classifier.py
arnaudvl: Don't you need that argmax for the soft_preds?

alibi_detect/cd/tests/test_classifier.py
```diff
@@ -40,6 +40,7 @@ def clfdrift_params(request):
 
 @pytest.mark.parametrize('clfdrift_params', list(range(n_tests)), indirect=True)
 def test_clfdrift(clfdrift_params):
+    print(clfdrift_params)
```
ojcobb, line 43: I need to undo this

alibi_detect/cd/pytorch/classifier.py
```diff
@@ -134,24 +133,31 @@ def score(self, x: np.ndarray) -> float:
...
         # iterate over folds: train a new model for each fold and make out-of-fold (oof) predictions
-        preds_oof, idx_oof = [], []
+        preds_oof_list, idx_oof_list = [], []
         for idx_tr, idx_te in splits:
             x_tr, y_tr, x_te = x[idx_tr], y[idx_tr], x[idx_te]
             ds_tr = TensorDataset(torch.from_numpy(x_tr), torch.from_numpy(y_tr))
             dl_tr = DataLoader(ds_tr, **self.dl_kwargs)  # type: ignore
             model = deepcopy(self.model)
-            train_args = [model, nn.CrossEntropyLoss(), dl_tr, self.device]
+            loss_fn = nn.CrossEntropyLoss() if (self.preds_type == 'logits') else nn.NLLLoss()
```
arnaudvl, line 151: Small comment: could this be placed in init, similar to tensorflow and set a `self.loss_fn` attribute?

Issue #191, opened 2021-03-23
what to do if ref and test data have different categories in chisquare?


AsiehH:
I have a question that I think should have been taken into account in this library but I can't find the solution. 

Currently if the reference data has a category feature that is different from that of the test data, we will get an error when we call the predict method in `TabularDrift` or `ChiSquareDrift`. 
I created `categories_per_feature` on the whole data but the way I split the data, one of the features of my reference data has categories from 0 to 11, and  0 to 12 for test data. 
The error I get is 
`operands could not be broadcast together with shapes (13,) (12,) `
This error comes from `chisquare` function under the hood. 

I think this is not a rare incident and it is probable that the reference data does not have all the categories of the test data for one or more features.

jklaise:
This is definitely a use case we want to cover more carefully. One thing to note is that if you know *a priori* how many categories are to be expected per feature then you should put them all in `categories_per_feature`, i.e. in your case you know there are 13 even if the reference set has 12.

More generally, of course, new categories may appear at test time which you did not know about. There are a couple of options we could explore in this case, two that come to mind:
- Explicitly add an extra "unknown" category per feature and provide some way of pre-processing the test data to ensure that all unseen categories are put into the "unknown" bucket
- When new categories are encountered, make sure the reference dataset and the `categories_per_feature` metadata are update to reflect this new reality.

@arnaudvl would be good to hear your thoughts?

arnaudvl:
Hi @AsiehH , for the vanilla Chi^2 test you can't have 0 values for a category in the reference dataset as this leads to divide by 0 issues, the test statistic goes to `inf` and p-value equals 0. The following formula ([from wikipedia](https://en.wikipedia.org/wiki/Chi-squared_test)) applies:

![image](https://user-images.githubusercontent.com/13081602/112618115-b8c44b80-8e1d-11eb-92b0-db5b5e8785d3.png),

where _mi_ refers to the observations of category _i_ in the reference data. We could by default return the `inf` statistic and 0 p-value when new categories are observed with a warning. But beyond that we would want to add methods which deal with this in a more robust way.

AsiehH:
Hi @arnaudvl,

Thanks for the response. 
Yes, in fact what I did was to use the whole dataset to create `categories_per_feature` before splitting. This solves the problem. 
[...]

PR #222, opened 2021-04-14
Allow unseen categories for the Chi^2 and tabular drift detectors

arnaudvl:
Allows for categorical drift detection with unseen categories in the test batches compared to the reference data and addresses #191 #193 . So for instance the reference data could contain categories `[0, 1, 2, 3]` for the first feature and the test data categories `[0, 2, 5, 6]`. The test of homogeneity is then performed on the union of categories between the reference and test data.


doc/source/methods/tabulardrift.ipynb
```diff
@@ -35,7 +35,7 @@
     "\n",
     "* `p_val`: p-value used for significance of the K-S and Chi-Squared test across all features. If the FDR correction method is used, this corresponds to the acceptable q-value.\n",
     "\n",
-    "* `categories_per_feature`: Dictionary with as keys the column indices of the categorical features and optionally as values the number of possible categorical values for that feature. E.g.: *{0: 5, 1: 9, 2: 7}* or *{0: None, 1: None, 2: None}*. If this argument is not specified, no features are assumed to be categorical.\n",
+    "* `categories_per_feature`: Dictionary with as keys the column indices of the categorical features and optionally as values the number of possible categorical values for that feature or a list with the possible values. E.g.: *{0: 4, 1: 2, 2: 3}* or *{0: None, 1: None, 2: None}* or *{0: [0, 1, 2, 3], 1: [0, 1], 2: [0, 1, 2]}*. If this argument is not specified, no features are assumed to be categorical.\n",
```
jklaise, line 38: I think we need more clear documentation on exactly which format is appropriate for which use case, e.g. `None` stuff for just identifying categorical columns, the first one for identifying number of of categories (all assumed to be in `[0, ..., n-1]` ?) and the last one for identifying categories by value explicitly (i.e. not assumed to be in `[0, ..., n-1]`).

doc/source/methods/chisquaredrift.ipynb
```diff
@@ -35,7 +35,7 @@
     "\n",
     "* `p_val`: p-value used for significance of the Chi-Squared test for. If the FDR correction method is used, this corresponds to the acceptable q-value.\n",
     "\n",
-    "* `categories_per_feature`: Optional dictionary with as keys the feature column index and as values the number of possible categorical values for that feature. E.g.: *{0: 5, 1: 9, 2: 7}*. If it is not specified, `categories_per_feature` is inferred from `x_ref`.\n",
+    "* `categories_per_feature`: Optional dictionary with as keys the feature column index and as values the number of possible categorical values for that feature. E.g.: *{0: 4, 1: 2, 2: 3}*. Alternatively, for each feature the possible categories can be provided in a list. E.g.: *{0: [0, 1, 2, 3], 1: [0, 1], 2: [0, 1, 2]}*. If it is not specified, `categories_per_feature` is inferred from `x_ref`.\n",
```
jklaise, line 38: Same as comment on `tabulardift.ipynb`, expand documentation and make it consistent between the two notebooks.

alibi_detect/cd/chisquare.py
```diff
@@ -64,16 +64,20 @@ def __init__(
...
+        # construct categories from the user-specified dict
+        if isinstance(categories_per_feature, dict):
+            vals = list(categories_per_feature.values())
+            if all(isinstance(v, (int, np.int16, np.int32, np.int64)) for v in vals):
+                # categories_per_feature = Dict[int, int]
+                categories_per_feature = {f: list(np.arange(v))  # type: ignore
+                                          for f, v in categories_per_feature.items()}
+            elif not all(isinstance(v, list) for v in vals):
+                raise NotImplementedError  # categories_per_feature not Dict[int, list]
+        else:  # infer number of possible categories for each feature from reference data
```
jklaise, line 76: This branch is also run if some values are floats instead of integers, might be better/more explicit to raise an error if non-ints are present rather than silently doing inference.

alibi_detect/cd/chisquare.py
```diff
@@ -64,16 +64,20 @@ def __init__(
             input_shape=input_shape,
             data_type=data_type
         )
-        if categories_per_feature is None:  # infer number of possible categories for each feature from reference data
+        # construct categories from the user-specified dict
+        if isinstance(categories_per_feature, dict):
+            vals = list(categories_per_feature.values())
+            if all(isinstance(v, (int, np.int16, np.int32, np.int64)) for v in vals):
```
jklaise, line 70: What happens if some values are floats instead of integers, looks like might result in an error later on?? Might be . rather than silently doing inference.

arnaudvl: It would result in an error which is the desired behaviour I believe. I could make the `NotImplementedError` a bit more informative.

jklaise: I'm not sure this is correct in the case that it is a `dict` of `list`s but the contents of the list are not integers.

alibi_detect/cd/chisquare.py
```diff
@@ -64,16 +64,21 @@ def __init__(
...
+                # categories_per_feature = Dict[int, int]
+                categories_per_feature = {f: list(np.arange(v))  # type: ignore
+                                          for f, v in categories_per_feature.items()}
+            elif not all(isinstance(v, list) for v in vals):
```
jklaise, line 74: This seems to prevent the use of `{0: None, ...}` as in the docs? I think it would be best if we have a test for each type of input `categories_per_feature` can take and check that the processing is done correctly on some dummy categorical data.

jklaise: Ah because this is `chi-square` I guess we don't have this as a valid input as all columns are categorical.

jklaise: Test would still be welcome for both detectors though...

arnaudvl: I will extend the current tests for both detectors.

alibi_detect/cd/chisquare.py
```diff
@@ -30,9 +30,13 @@ def __init__(
...
         categories_per_feature
-            Dict with as keys the feature column index and as values the number of possible categorical
-            values `n` for that feature. Eg: {0: 5, 1: 9, 2: 7}. If `None`, the number of categories is inferred
-             from the data. Categories are assumed to take values in the range `[0, 1, ..., n]`.
+            Optional dictionary with as keys the feature column index and as values the number of possible
+            categorical values for that feature or a list with the possible values. If you know how many
+            categories are present for a given feature you could pass this in the `categories_per_feature` dict
+            in the Dict[int, int] format, e.g. {0: 3, 3: 2}. If you pass N categories this will assume the
+            possible values for the feature are [0, ..., N-1]. You can also explicitly pass the possible categories
+            in the Dict[int, List[int]] format, e.g. {0: [0, 1, 2], 3: [0, 55]}. Not that the categories can be
```
jklaise, line 38: Not --> Note

alibi_detect/cd/tabular.py
```diff
@@ -32,11 +32,16 @@ def __init__(
...
-            inferred from the data. Categories are assumed to take values in the range `[0, 1, ..., n]`.
+            Dictionary with as keys the column indices of the categorical features and optionally as values
+            the number of possible categorical values for that feature or a list with the possible values.
+            If you know which features are categorical and simply want to infer the possible values of the
+            categorical feature from the reference data you can pass a Dict[int, NoneType] such as
+            {0: None, 3: None} if features 0 and 3 are categorical. If you also know how many categories are
+            present for a given feature you could pass this in the `categories_per_feature` dict in the
+            Dict[int, int] format, e.g. *{0: 3, 3: 2}*. If you pass N categories this will assume the possible
+            values for the feature are [0, ..., N-1]. You can also explicitly pass the possible categories in the
+            Dict[int, List[int]] format, e.g. {0: [0, 1, 2], 3: [0, 55]}. Not that the categories can be
```
jklaise, line 43: Not --> Note

alibi_detect/cd/chisquare.py
```diff
@@ -67,11 +71,13 @@ def __init__(
...
-            if all(isinstance(v, (int, np.int16, np.int32, np.int64)) for v in vals):
+            int_types = (int, np.int16, np.int32, np.int64)
+            if all(isinstance(v, int_types) for v in vals):
                 # categories_per_feature = Dict[int, int]
                 categories_per_feature = {f: list(np.arange(v))  # type: ignore
                                           for f, v in categories_per_feature.items()}
-            elif not all(isinstance(v, list) for v in vals):
+            elif not all(isinstance(val, list) for val in vals) and \
+                    all(isinstance(v, int_types) for val in vals for v in val):  # type: ignore
                 raise NotImplementedError('categories_per_feature needs to be None or one of '
```
jklaise, line 81: `ValueError` is probably more appropriate.

alibi_detect/cd/tabular.py
```diff
@@ -75,16 +80,19 @@ def __init__(
...
-            elif all(isinstance(v, (int, np.int16, np.int32, np.int64)) for v in vals):
+            elif all(isinstance(v, int_types) for v in vals):
                 # categories_per_feature = Dict[int, int]
                 categories_per_feature = {f: list(np.arange(v))  # type: ignore
                                           for f, v in categories_per_feature.items()}
-            elif not all(isinstance(v, list) for v in vals):
-                raise NotImplementedError  # categories_per_feature not Dict[int, list]
+            elif not all(isinstance(v, list) for v in vals) and \
+                    all(isinstance(v, int_types) for val in vals for v in val):  # type: ignore
+                raise NotImplementedError('categories_per_feature needs to be None or one of '
```
jklaise: NotImplementedError -> ValueError

jklaise: I don't think the default `None` case is actually handled (unlike in chisquare.py)?

jklaise: Oh I think see why, for `TabularDrift` the `None` case means "no categorical features" so no handling required, whereas for `ChiSquareDrift` the `None` case is more meaningful (infer all categories).

doc/source/methods/chisquaredrift.ipynb
```diff
@@ -35,7 +35,7 @@
...
     "\n",
-    "* `categories_per_feature`: Optional dictionary with as keys the feature column index and as values the number of possible categorical values for that feature. E.g.: *{0: 4, 1: 2, 2: 3}*. Alternatively, for each feature the possible categories can be provided in a list. E.g.: *{0: [0, 1, 2, 3], 1: [0, 1], 2: [0, 1, 2]}*. If it is not specified, `categories_per_feature` is inferred from `x_ref`.\n",
+    "* `categories_per_feature`: Optional dictionary with as keys the feature column index and as values the number of possible categorical values for that feature or a list with the possible values. If you know how many categories are present for a given feature you could pass this in the `categories_per_feature` dict in the *Dict[int, int]* format, e.g. *{0: 3, 3: 2}*. If you pass N categories this will assume the possible values for the feature are [0, ..., N-1]. You can also explicitly pass the possible categories in the *Dict[int, List[int]]* format, e.g. *{0: [0, 1, 2], 3: [0, 55]}*. Not that the categories can be arbitrary *int* values. If it is not specified, `categories_per_feature` is inferred from `x_ref`.\n",
```
jklaise, line 38: Not -> Note

doc/source/methods/tabulardrift.ipynb
```diff
@@ -35,7 +35,7 @@
...
+    "* `categories_per_feature`: Dictionary with as keys the column indices of the categorical features and optionally as values the number of possible categorical values for that feature or a list with the possible values. If you know which features are categorical and simply want to infer the possible values of the categorical feature from the reference data you can pass a *Dict[int, NoneType]* such as *{0: None, 3: None}* if features 0 and 3 are categorical. If you also know how many categories are present for a given feature you could pass this in the `categories_per_feature` dict in the *Dict[int, int]* format, e.g. *{0: 3, 3: 2}*. If you pass N categories this will assume the possible values for the feature are [0, ..., N-1]. You can also explicitly pass the possible categories in the *Dict[int, List[int]]* format, e.g. *{0: [0, 1, 2], 3: [0, 55]}*. Not that the categories can be arbitrary *int* values.\n",
```
jklaise, line 38: Not -> Note

alibi_detect/cd/tabular.py
arnaudvl: Yes, it should be clear in the docstrings / documentation.

jklaise (approved):

Issue #204, opened 2021-04-09
Preprocessing functions and detector wrappers for common and principled use cases


ojcobb:
Various drift detection strategies can be formulated as a combination of an existing drift detector (e.g ks or chi-squared) and a custom preprocessing step (often dependent on the deployed model of interest). We could avoid a lot of duplication and effort on behalf of users by providing implementations of these preprocessing steps and wrappers that call the corresponding preproc-test combinations. 

This issue is motivated by #197 and will encapsulate it by providing a wrapper for an MD3-like detector that accounts for drift into and out of areas of model uncertainty (the type of drift likely to cause degradation in model performance).

arnaudvl:
This will be closely related to #220

arnaudvl:
Similar to #220 is not exactly aligned anymore with our current thinking and can be closed.

PR #229, opened 2021-04-21
MD3-like detectors which consider regions of model uncertainty.

ojcobb:
Implements #204, which was motivated by and encompasses #197.

Wraps around KS or Chi2 detectors with custom preprocessing stages mapping instances onto (scalar) model uncertainties.  Separate `ClassifierUncertaintyDrift` and `RegressorUncertaintyDrift` classes to be used with classifiers and regressors respectively. The latter is limited to either regressors with dropout layers or a model outputting a vector corresponding to an ensemble of regressor predictions.

See the `cd_model_unc_cifar10.ipynb` for example usage and a new regression drift example problem using UCI's wine quality dataset.

Saving and loading not yet implemented.


alibi_detect/cd/preprocess.py
```diff
@@ -25,3 +29,146 @@ def pca(X: np.ndarray, n_components: int = 2, svd_solver: str = 'auto') -> np.nd
...
+    elif preds_type == 'logits':
+        probs = softmax(preds, axis=-1)
+    else:
+        raise NotImplementedError("Only prediction types 'probs' and 'logits' supported.")
+
+    if uncertainty_type == 'entropy':
+        uncertainties = entropy(probs, axis=-1)
+    elif uncertainty_type == 'margin':
+        top_2_probs = -np.partition(-probs, kth=1, axis=-1)[:, :2]
+        diff = np.abs(top_2_probs[:, 0] - top_2_probs[:, 1])
```
arnaudvl, line 101: Very minor comment: `np.abs` should not be necessary since the `top_2_probs` are already ordered since the partition is on the second element.

arnaudvl: Nice use of `np.partition` btw. :)

alibi_detect/cd/tests/test_model_uncertainty.py
```diff
@@ -0,0 +1,172 @@
...
+        return x
+
+
+def dumb_model(x, n_labels, softmax=False):
+    x = np.stack([np.mean(x*(i+1), axis=-1) for i in range(n_labels)], axis=-1)
+    if softmax:
+        x = scipy.special.softmax(x, axis=-1)
+    return x
+
+
```
arnaudvl, line 48: Get rid of commented section below?

alibi_detect/cd/utils.py
```diff
@@ -39,3 +43,67 @@ def update_reference(X_ref: np.ndarray,
             raise KeyError('Only `reservoir_sampling` and `last` are valid update options for X_ref.')
     else:
         return X_ref
+
+
+def activate_train_mode_for_dropout_layers(model: Union[nn.Module, tf.keras.Model], backend: str) -> Callable:
+    # TODO: Figure a way to do this properly for tensorflow models.
```
arnaudvl, line 49: Out of scope for current PR? If so, need to make sure to add an issue later.

alibi_detect/cd/utils.py
```diff
@@ -39,3 +43,67 @@ def update_reference(X_ref: np.ndarray,
...
+    device: Optional[str] = None,
+    tokenizer: Optional[Callable] = None,
+    max_len: Optional[int] = None,
+) -> np.ndarray:
+
+    backend = backend.lower()
+    model_kwargs = {
+        'model': model, 'batch_size': batch_size, 'tokenizer': tokenizer, 'max_len': max_len
+    }
+    if backend == 'tensorflow':
```
arnaudvl, line 84: For backend specification, best to have checks in place, see eg https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/cd/classifier.py#L98. You could practically copy the check linked over without the ClassifierDrift reerence and in a follow up PR we could implement a `_check_backend` function that you would use everywhere where dispatching occurs. @jklaise ?

alibi_detect/cd/utils.py
```diff
@@ -39,3 +43,67 @@ def update_reference(X_ref: np.ndarray,
...
+        from alibi_detect.cd.tensorflow.preprocess import preprocess_drift
+    elif backend == 'pytorch':
+        from alibi_detect.cd.pytorch.preprocess import preprocess_drift  # type: ignore
+        model_kwargs['device'] = device
+    else:
+        raise NotImplementedError(f'{backend} not implemented. Use tensorflow or pytorch instead.')
+    model_fn = partial(preprocess_drift, **model_kwargs)
+
```
arnaudvl, line 92: If I understand correctly up until here the `get_preds` function only initialises the same `model_fn` function over and over for each function call. I therefore believe it would be better to move it up to initialization of the detector and pass it as input to `partial(classifier_uncertainty, model_fn=model_fn, ..., )`. This would avoid conditional importing and initialization for each batch at runtime and could make serialization a bit easier as well.

alibi_detect/cd/utils.py
jklaise: Yea we should factor out `_validate_backend` or similar, can you open an issue?

alibi_detect/cd/utils.py
ojcobb: Yeah, no obvious way to do it. I'll add a note somewhere to be careful with using tf models and add an issue later.

alibi_detect/cd/model_uncertainty.py
```diff
@@ -0,0 +1,229 @@
...
+
+        Parameters
+        ----------
+        x
+            Batch of instances.
+        return_p_val
+            Whether to return the p-value of the test.
+        return_distance
+            Whether to return the corresponding test statistic (K-S for 'entropy', Chi2 for 'margin').
+        # TODO: Offer to return difference in uncertainty
```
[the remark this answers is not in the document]
ojcobb: It was trickier than anticipated so thought I'd defer to a separate issue, but just thought of a pretty simple way of implementing so will add to this PR.

alibi_detect/cd/utils.py
```diff
@@ -39,3 +43,67 @@ def update_reference(X_ref: np.ndarray,
             raise KeyError('Only `reservoir_sampling` and `last` are valid update options for X_ref.')
     else:
         return X_ref
+
+
+def activate_train_mode_for_dropout_layers(model: Union[nn.Module, tf.keras.Model], backend: str) -> Callable:
```
arnaudvl, line 48: There could be a few issues here:
- Since PyTorch is not a dependency, it might crash at import. Check https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/cd/base.py#L12 and https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/cd/base.py#L339 how this is handled in the base class.
- More importantly, this function contains backend specific functionality but is located in `alibi_detect.cd` and not e.g. `alibi_detect.cd.pytorch` or `alibi_detect.cd.tensorflow`. We should try to keep anything backend-specific in the relevant directories. As a result, the `activate_train_mode_for_dropout_layers` could be removed here and the backend-specific functionality could be invoked at initialization of the detector in `alibi_detect.cd.model_uncertainty.py` because it only needs to be done once. The backend-specific dropout activation could then simply live in `alibi_detect.utils.pytorch.utils.py` and `alibi_detect.utils.tensorflow.utils.py` since the functionality is not drift detector specific.

alibi_detect/cd/model_uncertainty.py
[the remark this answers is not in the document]
ojcobb: Actually I can't think of a simple way that wouldn't require changes to the KS and Chi2 detectors. Will defer to a separate issue.

doc/source/methods/modeluncdrift.ipynb
```diff
@@ -0,0 +1,151 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Model-uncertainty\n",
+    "\n",
+    "## Overview\n",
+    "\n",
+    "Model-uncertainty drift detectors aim to directly detect drift that's likely to effect the performance of a model of interest. The approach is to test for change in the number of instances falling into regions of the input space on which the model is uncertain in its predictions. For each instance in the reference set the detector obtains the model's prediction and some associated notion of uncertainty. For example for a classifier this may be the entropy of the predicted label probabilities or for a regressor with dropout layers dropout Monte Carlo can be used to provide a notion of uncertainty. The same is done for the test set and if significant differences in uncertainty are detected (via a Kolmogorov-Smirnoff test) then drift is flagged. The detector's reference set should be disjoint from the model's training set (on which the model's confidence may be higher).\n",
+    "\n",
```
arnaudvl, line 19: @ojcobb Could you add a quick sentence clarifying the difference for the classification model + KS drift when looking at entropy for uncertainty and classification model + Chi^2 for the margin method?

examples/cd_model_unc_cifar10.ipynb
```diff
@@ -0,0 +1,752 @@
...
+   ],
+   "source": [
+    "make_predictions(cd, X_h0, X_c, corruption)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As expected, drift was only detected on the corrupted datasets. The p-value is well above the $0.05$ threshold for *H0* and below for the corrupted datasets. This confirms that the corruptions impacted the certainty with which the model was able to predict, as we would expect."
```
arnaudvl, line 428: Mention here that no drift was detected on brightness as performance didn't really drop much.

examples/cd_model_unc_cifar10.ipynb
```diff
@@ -0,0 +1,752 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Model-uncertainty based drift detection on CIFAR-10\n",
```
arnaudvl, line 7: Change title to make clear both classifier uncertainty drift on CIFAR10 and regression uncertainty drift on Wine datasets are illustrated.

examples/cd_model_unc_cifar10.ipynb
ojcobb: Ahh I had done this but my notebooks often don't save all the changes for some reason. Will redo

Issue #230, opened 2021-04-22
Factor out backend (eg PyTorch, TensorFlow) validation.


arnaudvl:
Add `_validate_backend` functionality instead of checking on the fly as in https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/cd/classifier.py#L98.

amanjha8100:
I would like to solve this issue. I am new to open source therefore If I make any mistakes please do guide me.

jklaise:
@amanjha8100 for sure! The most appropriate place for this function seems to be https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/utils/frameworks.py. The function should take `backend` string provided by the user and validate it pretty much the same way as these lines: https://github.com/SeldonIO/alibi-detect/blob/13b2664dca601aebc25c774fd6814bd5d380d6e7/alibi_detect/cd/classifier.py#L97-L102 It would also be useful to define a `BACKENDS=['tensorflow', 'pytorch']` constant in this module.

@arnaudvl we may need additional functionality in the future, e.g. if some methods implement one backend but not the other.

amanjha8100:
I will try to solve this issue. Thank you for the guidance.

ascillitoe:
This issue will be solved by #537.

PR #233, opened 2021-04-22
Factor out backend issue 230

amanjha8100:
Please do review the PR. I did something very basic just created a function in frameworks.py and then called it from classifiers.py.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,6 +101,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+vir/
 
 # Spyder project settings
 .spyderproject
--- a/alibi_detect/cd/classifier.py
+++ b/alibi_detect/cd/classifier.py
@@ -1,6 +1,6 @@
 import numpy as np
 from typing import Callable, Dict, Optional, Union
-from alibi_detect.utils.frameworks import has_pytorch, has_tensorflow, validate_backend
+from alibi_detect.utils.frameworks import has_pytorch, has_tensorflow, _validate_backend
 
 if has_pytorch:
     from alibi_detect.cd.pytorch.classifier import ClassifierDriftTorch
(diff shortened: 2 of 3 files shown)


jklaise:
@amanjha8100 thanks for this! A few small changes:

 - You can delete the commented out text in `classifier.py`
 - Move the `BACKENDS` constant outside the function so that it is a module level variable
 - Rename the function to `_validate_backend` as that's an internal utility function.
 - Make the  `ImportError` message general (i.e. remove the reference to `ClassifierDrift`)

amanjha8100:
Ok thank you for the suggestions.

.gitignore
```diff
@@ -101,6 +101,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+vir/
```
ascillitoe, line 104: Is this needed?

alibi_detect/cd/classifier.py
```diff
@@ -94,12 +94,7 @@ def __init__(
         """
         super().__init__()
 
-        backend = backend.lower()
-        if backend == 'tensorflow' and not has_tensorflow or backend == 'pytorch' and not has_pytorch:
-            raise ImportError(f'{backend} not installed. Cannot initialize and run the '
-                              f'ClassifierDrift detector with {backend} backend.')
-        elif backend not in ['tensorflow', 'pytorch']:
-            raise NotImplementedError(f'{backend} not implemented. Use tensorflow or pytorch instead.')
+        _validate_backend(backend)
```
ascillitoe, line 97: This would need to be done for all the detectors. Running:

`grep -r "{backend} not implemented. Use tensorflow or pytorch instead" alibi_detect/*` will show where else it needs to be added.

ascillitoe (changes requested):
Apologies @amanjha8100, this PR seems to have slipped through the net. 

I've left two comments but in general, it seems like a nice idea to me. @jklaise do you think we still want to add this? It would be nice to have as we move towards increasing modularity (#388 ), although that might also require a rethink of this...

ascillitoe:
Related:
- #230

alibi_detect/utils/frameworks.py
```diff
@@ -9,3 +9,14 @@
     has_pytorch = True
 except ImportError:
     has_pytorch = False
```
ascillitoe, line 11: This could do with being updated to use `enum`'s, similar to how it is done in alibi:

https://github.com/SeldonIO/alibi/blob/master/alibi/utils/frameworks.py

ascillitoe:
Closing this as hasn't been touched in a while, and will be addressed by @mauicv's work on modularisation (see https://github.com/SeldonIO/alibi/pull/583# and #388).

PR #238, opened 2021-05-06
update transformers dependency

arnaudvl:
Update transformers dependency and functionality (preprocessing, saving/loading, example) for TensorFlow and PyTorch.

### DIFF
--- a/alibi_detect/cd/tensorflow/preprocess.py
+++ b/alibi_detect/cd/tensorflow/preprocess.py
@@ -6,34 +6,57 @@ from typing import Dict, Optional, Union
 from alibi_detect.utils.tensorflow.prediction import predict_batch, predict_batch_transformer
 
 
-class UAE(tf.keras.Model):
+class _Encoder(tf.keras.Model):
     def __init__(
             self,
-            encoder_net: tf.keras.Sequential = None,
-            input_layer: Union[tf.keras.layers.Layer, tf.keras.Model] = None,
-            shape: Optional[tuple] = None,
-            enc_dim: int = None
+            input_layer: Union[tf.keras.layers.Layer, tf.keras.Sequential, tf.keras.Model],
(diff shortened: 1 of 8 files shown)


alibi_detect/cd/tensorflow/preprocess.py
```diff
@@ -6,34 +6,57 @@
 from alibi_detect.utils.tensorflow.prediction import predict_batch, predict_batch_transformer
 
 
+class _Encoder(tf.keras.Model):
+    def __init__(
+            self,
+            input_layer: Union[tf.keras.layers.Layer, tf.keras.Sequential, tf.keras.Model],
```
jklaise, line 12: Technically every `tf.keras.Sequential` model is an instance of `tf.keras.Model`, so we could simplify all type hints to just `tf.keras.Model`? Hopefully people wouldn't take it as `tf.keras.Sequential` is not allowed.

alibi_detect/utils/pytorch/prediction.py
```diff
@@ -90,7 +90,7 @@ def predict_batch_transformer(x: Union[np.ndarray, torch.Tensor], model: Union[n
         for i in range(n_minibatch):
             istart, istop = i * batch_size, min((i + 1) * batch_size, n)
             tokens = tokenizer.batch_encode_plus(  # type: ignore
```
jklaise, line 92: The transformer docs say `batch_encode_plus` is deprecated and `__call__` should be used: https://huggingface.co/transformers/internal/tokenization_utils.html?highlight=batch_encode_plus#transformers.tokenization_utils_base.PreTrainedTokenizerBase.batch_encode_plus

alibi_detect/cd/tensorflow/preprocess.py
arnaudvl: Good point, I will raise this in a separate issue since both are used across the library.

Issue #261, opened 2021-06-07
Support updating reference data for LSDD detector


ojcobb:
During initialisation the lsdd detector (`alibi_detect/cd/lsdd.py`) performs normalisation of the reference data `x_ref`. When `update_x_ref` is not `None` unnormalized data is being integrated alongside normalized data which is incorrect and is causing problems. The integration of the data takes part within the base class which doesn't allow for additional normalisation steps. 

There may be a workaround involving overriding `self.preprocess_fn` with a new function `lambda x: self._normalize(self.preprocess_fn(x))`.

arnaudvl:
#263

PR #263, opened 2021-06-08
Lsdd updates

ojcobb:
A few bug fixes:

- Ensure when updating reference data it gets normalised if necessary. (#261)
- Catch overflow/underflow caused by too high dimensionality.
- Catch when error is caused by too many repeat samples. (And don't test using repeat samples)

### DIFF
--- a/alibi_detect/utils/pytorch/distance.py
+++ b/alibi_detect/utils/pytorch/distance.py
@@ -121,12 +121,6 @@ def permed_lsdds(
     the unpermed LSDD estimate.
     """
 
-    # Check for overflow/underflow caused by too high dimensionality
-    if (~torch.isfinite(H)).any():
-        raise ValueError(
-            "Overflow or underflow occured. Try reducing dimensionality or trying "
-            "MMD-based detection instead")
-
     # Compute (for each bootstrap) the average distance to each kernel center (Eqn 7)
     k_xc_perms = torch.stack([k_all_c[x_inds] for x_inds in x_perms], 0)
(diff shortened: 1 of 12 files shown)


alibi_detect/utils/pytorch/distance.py
```diff
@@ -120,6 +120,13 @@ def permed_lsdds(
     Vector of B LSDD estimates for each permutation, H_lam_inv which may have been inferred, and optionally
     the unpermed LSDD estimate.
     """
+
+    # Check for overflow/underflow caused by too high dimensionality
+    if (~torch.isfinite(H)).any():
+        raise ValueError(
+            "Overflow or underflow occured. Trying reducing dimensionality or trying "
```
arnaudvl, line 127: small typo -> "try" instead of trying 2x

alibi_detect/utils/tensorflow/distance.py
```diff
@@ -149,6 +149,13 @@ def permed_lsdds(
     Vector of B LSDD estimates for each permutation, H_lam_inv which may have been inferred, and optionally
     the unpermed LSDD estimate.
     """
+
+    # Check for overflow/underflow caused by too high dimensionality
+    if tf.reduce_any(~tf.math.is_finite(H)):
+        raise ValueError(
+            "Overflow or underflow occured. Trying reducing dimensionality or trying "
```
arnaudvl, line 156: Also try instead of trying

alibi_detect/utils/pytorch/distance.py
ojcobb: Ah yeah thanks. Fixed.

Issue #253, opened 2021-05-14
Update type hints for tensorflow functions that we pass numpy inputs


ojcobb:
Because tensorflow functions accept numpy inputs we often don't bother to make the conversions but then the type hints are misleading. So we should update type hints to acknowledge fact numpy inputs are also fine.

tim5go:
@ojcobb 
Can I give a try on this issue?

ojcobb:
Hi @tim5go, that would be great.

An example of what I was referring to is 
https://github.com/SeldonIO/alibi-detect/blob/a2fc6125eaf8d15068a3069e64d8fbc9cd1b41cb/alibi_detect/cd/tensorflow/mmd.py#L84-L86
where you can see `x` and `y` may be numpy arrays but the default kernel `GaussianRBF` is typed to operate only on tensorflow tensors.

There will likely be many places with similar inconsistencies. Let me know if I can advise further.

tim5go:
@ojcobb 
Thanks, your example is very clear!
Let me work on it. I will create a PR very soon.

ojcobb:
This problem wasn't as widespread as first thought. On reflection the benefit of updating the type hints for the small number of tensorflow functions we pass numpy inputs probably doesn't outweigh the cost of either breaking consistency of type hints between different tensorflow functions or using `Union[tf.Tensor, np.ndarray]` everywhere.

PR #268, opened 2021-06-11
Update type hints for Tensorflow functions that we pass numpy inputs

tim5go:
Regarding to issue #253, we would like to allow type hints to accept `numpy` inputs when we call `tensorflow` functions

### DIFF
--- a/alibi_detect/utils/pytorch/kernels.py
+++ b/alibi_detect/utils/pytorch/kernels.py
@@ -1,11 +1,11 @@
 import numpy as np
 import torch
 from . import distance
-from typing import Optional
+from typing import Optional, Union
 
 
 class GaussianRBF:
-    def __init__(self, sigma: Optional[torch.Tensor] = None) -> None:
+    def __init__(self, sigma: Optional[Union[np.ndarray, torch.Tensor]] = None) -> None:
         """
         Gaussian RBF kernel: k(x,y) = exp(-(1/(2*sigma^2)||x-y||^2). A forward pass takes
         a batch of instances x [Nx, features] and y [Ny, features] and returns the kernel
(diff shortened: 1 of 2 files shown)


ojcobb:
Hi @tim5go,

Thank you for having a crack at this issue for us. However I apologise as perhaps I should have been clearer in my initial description.

We don't necessarily want to change the type hints of every tensorflow function that _could_ operate on numpy inputs. The hint might then obscure a class's intended use rather than help demonstrate it. For example `EncoderLSTM` is intended as as internal LSTM and hinting at numpy inputs obscures this a little.

Instead we specifically wish to change the type hints for those that we actually do pass numpy inputs. For example, in the example I previously gave the fact that `alibi_detect.utils.tensorflow.kernels.GaussianRBF` is hinted to operate exclusively on tensorflow tensors is problematic because we most commonly apply it directly to numpy arrays.

I realise these example are hard to find without familiarity with the library, so no worries if you don't fancy digging them up.

tim5go:
@ojcobb 
I see. So I guess I should revert most of the changes (as they all live inside a Tensorflow Model)
This type hint is more applicable on the Utility function case, am I right?

ojcobb:
@tim5go yes, sorry about that. Indeed it's more for utility functions. 

There might not be that many actually -- `GaussianRBF` is probably the most important one.

tim5go:
@ojcobb 
PR updated, please check.
At the same time, I am still checking if there's any similar case as `GaussianRBF`

ojcobb:
This problem wasn't as widespread as first thought. On reflection the benefit of updating the type hints for the small number of tensorflow functions we pass numpy inputs probably doesn't outweigh the cost of either breaking consistency of type hints between different tensorflow functions or using Union[tf.Tensor, np.ndarray] everywhere.

PR #278, opened 2021-06-28
allow for custom preprocessing steps for batch predictions

arnaudvl:
Accomodates various data types (e.g. graphs) to be passed to the drift detectors.

- [x] Remove redundant `predict_batch` function in `alibi_detect.utils.prediction`
- [x] Change all old `predict_batch` usages across outlier and adversarial detectors
- [x] Add multi-output prediction support
- [x] Allow for flexible input types via a preprocessing step in `predict_batch`
- [x] Update transformers batch predict function
- [x] Update tests


alibi_detect/utils/pytorch/prediction.py
```diff
@@ -1,12 +1,14 @@
...
 from typing import Callable, Union
+from alibi_detect.utils.prediction import tokenize_transformer
 
 
-def predict_batch(x: Union[np.ndarray, torch.Tensor], model: Union[nn.Module, nn.Sequential],
-                  device: torch.device = None, batch_size: int = int(1e10),
-                  dtype: Union[np.float32, torch.dtype] = np.float32) -> Union[np.ndarray, torch.Tensor]:
+def predict_batch(x: Union[list, np.ndarray, torch.Tensor], model: Union[Callable, nn.Module, nn.Sequential],
+                  device: torch.device = None, batch_size: int = int(1e10), preprocess_fn: Callable = None,
+                  dtype: Union[np.float32, torch.dtype] = np.dtype) -> Union[np.ndarray, torch.Tensor, tuple]:
```
jklaise, line 11: I don't think these `dtype` annotations are doing what you mean/want...

alibi_detect/utils/pytorch/prediction.py
```diff
@@ -21,37 +23,48 @@ def predict_batch(x: Union[np.ndarray, torch.Tensor], model: Union[nn.Module, nn
...
+                    preds = tuple([] for _ in range(len(preds_tmp)))
+                for j, p in enumerate(preds_tmp):
+                    if device.type == 'cuda' and isinstance(p, torch.Tensor):
+                        p = p.cpu()
+                    preds[j].append(p if not return_np or isinstance(p, np.ndarray) else p.numpy())
+            else:
+                if device.type == 'cuda' and isinstance(preds_tmp, torch.Tensor):
+                    preds_tmp = preds_tmp.cpu()
+                preds.append(preds_tmp if not return_np or isinstance(preds_tmp, np.ndarray)  # type: ignore
+                             else preds_tmp.numpy())
```
jklaise, line 61: Probably can condense the whole thing without condiditions by doing `preds_tmp = tuple(preds_tmp)` ? Not a big deal though.

alibi_detect/utils/tensorflow/prediction.py
```diff
@@ -1,10 +1,13 @@
...
-from typing import Union
+from typing import Callable, Union
+from alibi_detect.utils.prediction import tokenize_transformer
 
 
-def predict_batch(x: Union[np.ndarray, tf.Tensor], model: tf.keras.Model, batch_size: int = int(1e10),
-                  dtype: Union[np.float32, tf.DType] = np.float32) -> Union[np.ndarray, tf.Tensor]:
+def predict_batch(x: Union[list, np.ndarray, tf.Tensor], model: Union[Callable, tf.keras.Model],
+                  batch_size: int = int(1e10), preprocess_fn: Callable = None,
+                  dtype: Union[np.dtype, tf.DType] = np.dtype) -> Union[np.ndarray, tf.Tensor, tuple]:
```
jklaise, line 10: Not sure about these dtype annotations as above for PT.

alibi_detect/utils/pytorch/prediction.py
arnaudvl: The type of `preds` in `if` is `Tuple[List[Union[np.ndarray, torch.Tensor]]]` and in `else` `List[Union[np.ndarray, torch.Tensor]]`. So your suggestion would then be to do `preds_tmp = tuple(preds_tmp)` and then add an if statement to check if `len(preds)==1`, and if so, to just return `preds[0]` at the end?

jklaise: Yea, suggestion was to convert to `tuple` so we treat both cases in the same way. Good point on the length check as I assume we want to unpack tuples of length 1?

jklaise: Is the implicit assumption that multi-output models always return a `Tuple` instead of a `List` ?

arnaudvl: Yes, the assumption is that a `tuple` is returned. I believe this is common practice as I can't think of a practical example where models return lists of outputs (you always could if you really wanted of course...). By extending to list types on line 50 and not returning a tuple at the end I believe the list multi-output case would also be covered if needed.

With the length check and added `tuple(preds_tmp)` I think the difference in code length will be minimal though.

alibi_detect/utils/pytorch/prediction.py
arnaudvl: Unsure if even `dtype: Union[np.dtype, torch.dtype] = np.float32` would be correct. Thoughts?

alibi_detect/utils/pytorch/prediction.py
jklaise: ok then fine to leave as is. btw the `else` branch doesn't consider `List`, assumes only a single tensor is returned?

alibi_detect/utils/pytorch/prediction.py
jklaise: Unlikely, if in doubt, check with `isinstance`. We had this conversation previously, but I forgot what the conclusion was...

jklaise: I.e. `np.float32` is not an instance of `np.dtype`, there is something subtle going on there.

alibi_detect/utils/pytorch/prediction.py
arnaudvl: Yes, can add a check that the right type is returned by the model (`Union[np.ndarray, torch.Tensor]`) and throw an error that other return types are not covered.

alibi_detect/utils/pytorch/prediction.py
jklaise: Probably needs a separate issue to look at how to type this properly (and if it's even possible).

arnaudvl: I can't seem to find a nice way to cover all numpy types without explicitly listing them in `dtype`. It's a little bit confusing because `isinstance(x.dtype, np.dtype)=True` but `type(np.float32)=type` and `isinstance(np.float32, np.dtype)=False` but e.g. `isinstance(torch.float32, torch.dtype)=True`.

jklaise: Both `np.dtype` and `torch.dtype` are of type `type` (because they are classes), so `dtype: type = np.float32` is valid, but not very useful as it doesn't narrow it down. E.g. see https://github.com/SeldonIO/alibi-detect/blob/5be1ed9f240112089b492d3f98905b837b3298cf/alibi_detect/utils/prediction.py#L27

arnaudvl: Yes, even if strictly not correct, at least in this PR it should be clear that you either pass a numpy or torch type. So leave as is and open issue?

jklaise: Yea I guess so

arnaudvl: #281

jklaise (approved):
lgtm

PR #280, opened 2021-06-29
update licenses

arnaudvl:


jklaise:
@arnaudvl licenses are built and checked on Python 3.8 only (https://github.com/SeldonIO/alibi-detect/blob/3a3e84bb3e15287d31b11862d160efa6a9dd2f9a/.github/workflows/ci.yml#L69-L83) is it possible you're using a different Python version locally which had that extra dependency making the check fail?

Issue #275, opened 2021-06-25
Extend uncertainty drift detectors to data types beyond numpy arrays


arnaudvl:

arnaudvl:
#282

PR #282, opened 2021-06-30
allow list inputs for classifier and regressor uncertainty drift

arnaudvl:
Addresses #275 

- [x] Classifier list inputs
- [x] Regression list inputs
- [x] Update tests
- [x] Add graph example
- [x] Update docs


arnaudvl:
Note: I am tempted to rename the `preprocess_batch_fn` and `preprocess_fn` in `predict_batch` to `collate_fn` to be more in line with PyTorch. @ojcobb

examples/cd_mol.ipynb
```diff
@@ -0,0 +1,1071 @@
...
+    "\n",
+    "preprocess_fn = partial(predict_batch, model=model, device=device, preprocess_fn=batch_fn, batch_size=32)\n",
+    "dd = KSDrift(x_ref, p_val=.05, preprocess_fn=preprocess_fn)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Since the dataset is heavily imbalanced, we will test the detectors on a sample which oversamples from the minority class (molecules which inhibit HIV virus replication):"
```
ojcobb, line 757: This implies our detectors do not work on class-imbalanced data. I assume that's not the case?

ojcobb: I think I misunderstood this. The correction of the class imbalance is introduced as a manifestation of drift? I think it might help to make this clearer.

examples/cd_mol.ipynb
```diff
@@ -0,0 +1,1071 @@
...
+   ],
+   "source": [
+    "make_predictions(dd, xs)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Although we didn't pick up drift in the GIN model prediction distribution for the test sample, we can see that the model is less certain about the predictions on the test set, illustrated by the lower ROC-AUC.\n",
```
ojcobb, line 878: It looks to me like drift is picked up for the test sample?

ojcobb (commented):
Looks good! 

Only minor comment is that I think now with various preprocessing function abstractions each doing subtly different things I think we should make them each as clear as possible. In particular I think we should now update the `predict_batch` functions to be clearer (especially now that it's being directly leveraged in examples).

I think many would assume `predict_batch` would return predictions for a single (mini) batch of data. This is reinforced by the fact that the `x` arg is described as `Batch of instances`. But this use of 'batch' is different to that of the `batch_size` arg. 

I think `predict_in_batches` or even `batch_predict` would be clearer and the description of `x` should be changed to something that makes clearer that it contains multiple batches worth of data.

examples/cd_mol.ipynb
```diff
@@ -0,0 +1,1071 @@
...
+    "\n",
+    "\n",
+    "def train(loader: DataLoader, verbose: bool = False) -> None:\n",
+    "    dl = tqdm(loader, total=len(loader)) if verbose else loader\n",
+    "    model.train()\n",
+    "    for data in dl:\n",
+    "        data = data.to(device)\n",
+    "        optim.zero_grad()\n",
+    "        y_hat = model(data)\n",
+    "        is_labeled = data.y == data.y\n",
```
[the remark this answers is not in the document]
arnaudvl: Not for nan values, which is the behaviour tested here (not all instances have values for all targets). Example:

```python
import torch
x = torch.tensor([1, 2, 3]).float()
x[x == 2] = float('nan')
x == x
```
`tensor([ True, False,  True])`

examples/cd_mol.ipynb
```diff
@@ -0,0 +1,1071 @@
...
+    "from alibi_detect.utils.pytorch import predict_batch\n",
+    "from functools import partial\n",
+    "\n",
+    "\n",
+    "def batch_fn(data: List[Data]) -> Batch:\n",
+    "    return Batch().from_data_list(data)\n",
+    "\n",
+    "\n",
+    "preprocess_fn = partial(predict_batch, model=model, device=device, preprocess_fn=batch_fn, batch_size=32)\n",
+    "dd = KSDrift(x_ref, p_val=.05, preprocess_fn=preprocess_fn)"
```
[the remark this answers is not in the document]
arnaudvl: I think it's probably the other way around by now. Initially we started off with the idea of doing concept drift (cd) but really it evolved towards doing data drift (dd).

examples/cd_mol.ipynb
arnaudvl: Yes that's correct, I will clarify this sentence.

examples/cd_mol.ipynb
arnaudvl: No, this refers to the prediction distribution drift in the previous section, not the prediction uncertainty drift in this section. But it is clearly unclear so I'll clarify the sentence. :)

arnaudvl:
> Looks good!
> 
> Only minor comment is that I think now with various preprocessing function abstractions each doing subtly different things I think we should make them each as clear as possible. In particular I think we should now update the `predict_batch` functions to be clearer (especially now that it's being directly leveraged in examples).
> 
> I think many would assume `predict_batch` would return predictions for a single (mini) batch of data. This is reinforced by the fact that the `x` arg is described as `Batch of instances`. But this use of 'batch' is different to that of the `batch_size` arg.
> 
> I think `predict_in_batches` or even `batch_predict` would be clearer and the description of `x` should be changed to something that makes clearer that it contains multiple batches worth of data.
> 
> Also is the additional layer of abstraction introduced by the `preprocess_drift` function really necessary? Can we not just have `predict_batch` and `predict_batch_transformer`?

Some good points. We can address consistency across different functions and their clarity in a separate issue.

Issue #247, opened 2021-05-13
Change PCA preprocessing step


arnaudvl:
Fit on reference data, project test batch.

ojcobb:
I'd consider the recommended workflow for PCA preprocessing to be fitting a PCA object on a split of reference data which then has a `call` method that projects new data onto the PCs. This object could then be passed as the `preprocess_fn` argument.

However this can be implemented as:
```
from sklearn.decomposition import PCA
pca = PCA(n_components)
pca.fit(x_split)
preprocess_fn = pca.transform
```

The only possible way I can see of simplifying this any further is providing a wrapper `cd.PCA` that has a `__call__()` method that calls `transform()`. The above could then be implemented as:
```
from alibi_detect.preprocess import PCA
pca = PCA(n_components)
pca.fit(x_split)
preprocess_fn = pca
```

I'm not sure this is worth implementing. I think we'd be better off just providing an example of how to use `sklearn.decomposition.PCA`? @arnaudvl

arnaudvl:
Sounds good, feel free to add as an example to one of the existing example notebooks. The PCA preprocessing step is mentioned in various locations across the notebooks, so best to update those mentions as well as it doesn't come with the library anymore.

arnaudvl:
#284

PR #284, opened 2021-07-01
Remove PCA preprocessing function

ojcobb:
Addresses #247.

An example of how to use PCA as a preprocessing step is now contained within the `cd_online_wine.ipynb` example. References to including PCA-specific functionality have been adapted.


arnaudvl:
Minor comment: is it possible to suppress the local CUDA UserWarning from the online wine drift example?

PR #286, opened 2021-07-09
allow List[Any] input to classifierdrift

arnaudvl:
- [x] PyTorch
- [x] TensorFlow
- [x] Instance level feedback
- [x] Example
- [x] Update TensorFlow trainer to allow for custom `tf.keras.utils.Sequence` datasets
- [x] Tests: resolve issue `update_x_ref`
- [x] Docs


examples/cd_text_amazon.ipynb
```diff
@@ -0,0 +1,553 @@
...
+    "from alibi_detect.cd.pytorch import preprocess_drift\n",
+    "from alibi_detect.models.pytorch import TransformerEmbedding\n",
+    "from functools import partial\n",
+    "from transformers import AutoTokenizer\n",
+    "\n",
+    "emb_type = 'hidden_state'  # pooler_output, last_hidden_state or hidden_state\n",
+    "# layers to extract hidden states from for the embedding used in drift detection\n",
+    "# only relevant for emb_type = 'hidden_state'\n",
+    "n_layers = 8\n",
+    "layers = [-_ for _ in range(1, n_layers + 1)]\n",
```
ojcobb, line 157: Is the `-` sign here intentional?

examples/cd_text_amazon.ipynb
```diff
@@ -0,0 +1,553 @@
...
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "Some layers from the model checkpoint at distilbert-base-uncased were not used when initializing TFDistilBertModel: ['vocab_transform', 'vocab_projector', 'activation_13', 'vocab_layer_norm']\n",
+      "- This IS expected if you are initializing TFDistilBertModel from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).\n",
+      "- This IS NOT expected if you are initializing TFDistilBertModel from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).\n",
+      "All the layers of TFDistilBertModel were initialized from the model checkpoint at distilbert-base-uncased.\n",
+      "If your task is similar to the task the model of the checkpoint was trained on, you can already use TFDistilBertModel for predictions without further training.\n"
```
ojcobb, line 391: Lots of warnings showing around this part of the notebook.

alibi_detect/models/tensorflow/trainer.py
```diff
@@ -1,21 +1,25 @@
...
-            callbacks: tf.keras.callbacks = None) -> None:  # TODO: incorporate callbacks + LR schedulers
+def trainer(
+        model: tf.keras.Model,
+        loss_fn: tf.keras.losses,
+        x_train: np.ndarray,
+        y_train: np.ndarray = None,
+        dataset: tf.keras.utils.Sequence = None,
+        optimizer: tf.keras.optimizers = tf.keras.optimizers.Adam(learning_rate=1e-3),
+        loss_fn_kwargs: dict = None,
+        preprocess_fn: Callable = None,
```
ojcobb, line 15: Can we now rename this arg `preprocess_batch_fn` given that it acts on batches and reserve `preprocess_fn` for functions that are expected to include their own batching logic if necessary? Would also be more consistent as at the moment we have `preprocess_fn=preprocess_batch_fn` in places.

alibi_detect/models/pytorch/trainer.py
```diff
@@ -12,14 +12,39 @@ def trainer(
         device: torch.device,
         optimizer: Callable = torch.optim.Adam,
         learning_rate: float = 1e-3,
+        preprocess_fn: Callable = None,
```
ojcobb, line 15: See above.

alibi_detect/cd/classifier.py
```diff
@@ -26,12 +29,14 @@ def __init__(
...
             optimizer: Optional[Callable] = None,
             learning_rate: float = 1e-3,
-            compile_kwargs: Optional[dict] = None,
             batch_size: int = 32,
+            preprocess_batch_fn: Optional[Callable] = None,
             epochs: int = 3,
             verbose: int = 0,
             train_kwargs: Optional[dict] = None,
             device: Optional[str] = None,
+            dataset: Optional[Callable] = None,
```
ojcobb, line 38: Is it definitely necessary to pass this `dataset` argument? If so I think it would be helpful to make description clearer as to when it is useful and that's it's just a wrapper rather something containing actual data.

alibi_detect/cd/pytorch/classifier.py
```diff
@@ -134,30 +149,35 @@ def score(self, x: np.ndarray) -> Tuple[float, float]:
...
             preds_oof_list.append(preds)
             idx_oof_list.append(idx_te)
         preds_oof = np.concatenate(preds_oof_list, axis=0)
         probs_oof = softmax(preds_oof, axis=-1) if self.preds_type == 'logits' else preds_oof
         idx_oof = np.concatenate(idx_oof_list, axis=0)
         y_oof = y[idx_oof]
-
         p_val, dist = self.test_probs(y_oof, probs_oof, n_ref, n_cur)
-        return p_val, dist
+        return p_val, dist, y_oof, probs_oof[:, 1]
```
ojcobb, line 183: Simply returning `y_oof` and `probs_oof` doesn't seem particularly useful to me. It is a mix of reference and test instances ordered in a way that doesn't correspond to either the reference or test window? So even if high scores are observed there's no way to connect the score to the corresponding instances?

examples/cd_text_amazon.ipynb
```diff
@@ -0,0 +1,553 @@
...
+    "from alibi_detect.cd import ClassifierDrift\n",
+    "from alibi_detect.utils.prediction import tokenize_transformer\n",
+    "\n",
+    "\n",
+    "def make_predictions(model, backend: str, ref_size: int, test_size: int, n_sample: int, stars_h1: int = 4) -> None:\n",
+    "    \"\"\" Create drift Classifier detector, init, sample data and make predictions. \"\"\"\n",
+    "    \n",
+    "    # batch_fn tokenizes each batch of instances of the reference and test set during training\n",
+    "    b = 'pt' if backend == 'pytorch' else 'tf'\n",
+    "    batch_fn = partial(tokenize_transformer, tokenizer=tokenizer, max_len=max_len, backend=b)\n",
```
ojcobb, line 313: Can we always do this and get rid of the `preprocess_drift` function?

ojcobb (commented):
Looks good! Some minor comments.

examples/cd_text_amazon.ipynb
arnaudvl: Yes, using the last `n_layers` hidden layers.

examples/cd_text_amazon.ipynb
arnaudvl: Yes, internal transformers library warnings unfortunately. Tried with warnings in the notebook explicitly switched off but that didn't help.

alibi_detect/cd/pytorch/classifier.py
arnaudvl: `y_oof` contains the labels (0=reference, 1=test) and `probs_oof` contains the scores in the same order, so it does correspond to reference and test window I think and could be used downstream to e.g. plot any way you see fit?

arnaudvl: Ah ok, I see, you meant to tie back to the order of the original reference and test batch, good point.

examples/cd_text_amazon.ipynb
arnaudvl: `batch_fn` is used for `preprocess_batch_fn` here is different than the function of `preprocess_drift` which is a utility function for `preprocess_fn`.

alibi_detect/cd/classifier.py
arnaudvl: It is necessary and should hopefully be clear from the docs as it's there with example. But let me know if it isn't and we'll clarify this.

alibi_detect/models/tensorflow/trainer.py
arnaudvl: Included this in #298

examples/cd_text_amazon.ipynb
arnaudvl: Again part of #298

alibi_detect/models/pytorch/trainer.py
arnaudvl: #298

alibi_detect/cd/classifier.py
```diff
@@ -148,15 +148,16 @@ def predict(self, x: Union[np.ndarray, list],  return_p_val: bool = True,
         return_distance
             Whether to return a notion of strength of the drift.
             K-S test stat if binarize_preds=False, otherwise relative error reduction.
-        return_score
-            Whether to return the instance level classifier scores and labels (0=reference data, 1=test data).
+        return_probs
+            Whether to return the instance level classifier probabilities for the test data
+            (0=reference data, 1=test data).
```
ojcobb, line 153: For the reference and test data?

PR #306, opened 2021-08-06
Interpretable detector and deep kernels detector

ojcobb:
For the implementation of an interpretable drift detector I found that the NME test statistic of Jitkrittum (https://arxiv.org/abs/1605.06796) didn't actually work very well. The NME test stat (for a single test location) is (z.mean()/z.std())^2 where z_i = k(x_i,w) - k(y_i, w) for a reference sample x_i and test sample y_i. The z.mean() in the numerator encourages a test location w to be learnt that pushes k(x_i,w) away from k(y_i,w) on average, but the z.std() term in the denominator counteracts this and encourages them all to be the same, resulting in poor test locations.

The z.std() denominator is to give the NME a known distribution under the null for testing. We can remove it and just perform a nonparametric test instead. 

However we still then have a problem of requiring paired reference and test samples. We can instead encourage the pushing apart by fitting a domain classifier of the form logit(prob_test) = a + b*k(x_i, w) where x_i may now be a reference or test instance. 

If we just learn test locations like this then the non-discriminative features in the test-loc can take any value and the test location ends up very noisy. So we instead fit a classifier logit(prob_test) = a + b*k(z_i, m+d) where m is the mean reference instance. We apply l1 regularisation to d to encourage sparsity. We can then interpret d as what should be added to an average instance to make it appear more (b>0) or less (b<0) similar to a test instance.

I put together a quick example of usage in `examples/cd_spot_the_diff.ipynb`. Sadly the quality of diffs and detection performance can be sensitive to learning rate (like any learnt approach) and also strength of regularisation.

Implementation of a DeepKernel detector coming up, as well as documentation and examples.


ojcobb:
Should note that there are some more general changes contained in this PR.

- Make `GaussianRBF` trainable pytorch and tensorflow components.
- Allow `ClassifierDrift` to continue training classifier from where it left off with previous call to `predict()`.
- Allow `ClassifierDrift` to return the trained model from `predict()`.
- Allow trainer to take a `reg_loss_fn` argument that defines a regularisation term to be added to the loss.
- `trainer` now prints moving averages when in verbose mode.

alibi_detect/models/pytorch/trainer.py
```diff
@@ -14,6 +14,7 @@ def trainer(
         learning_rate: float = 1e-3,
         preprocess_fn: Callable = None,
         epochs: int = 20,
+        reg_loss_fn: Callable = (lambda model: 0),
```
arnaudvl, line 17: add docstring

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -14,22 +18,76 @@ def __init__(self, sigma: Optional[torch.Tensor] = None) -> None:
...
+            Bandwidth used for the kernel. Needn't be specified if being inferred or trained.
+            Can pass multiple values to eval kernel with and then average.
+        trainable
+            Whether or not to track gradients w.r.t. sigma to allow it to be trained.
         """
         super().__init__()
-        self.sigma = sigma
+        if sigma is None:
+            self.log_sigma = nn.Parameter(torch.empty(1), requires_grad=trainable)
+            self.init_required = True  # TODO: Is this right now?
```
arnaudvl, line 29: Looks right, although the ValueError in the forward pass could already have been thrown at init if `trainable=True` and `sigma=None`. One alternative would be to infer a sigma first via the median heuristic and set trainable True afterwards for log_sigma. But that might be more of a non-priority edge case.

alibi_detect/cd/__init__.py
```diff
@@ -1,8 +1,10 @@
 from .chisquare import ChiSquareDrift
 from .classifier import ClassifierDrift
 from .ks import KSDrift
+from .learnt_kernel import LearntKernelDrift
```
arnaudvl, line 4: It becomes clear we are not a US-based drift detection library here. :)

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -14,22 +18,76 @@ def __init__(self, sigma: Optional[torch.Tensor] = None) -> None:
...
+class DeepKernel(nn.Module):
+    """"
+    Computes simmilarities as k(x,y) = (1-eps)*k_a(proj(x), proj(y)) + eps*k_b(x,y)
+    """
+    def __init__(
+        self,
+        proj: nn.Module,
+        kernel_a: nn.Module = GaussianRBF(trainable=True),
+        kernel_b: nn.Module = GaussianRBF(trainable=True),
+        eps: Union[float, str] = 'trainable'
```
arnaudvl, line 70: Is there a reason why we don't choose for 2 kwargs: `eps` and `eps_trainable` or something similar and instead set `eps` in the trainable case by default to .5?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -14,22 +18,76 @@ def __init__(self, sigma: Optional[torch.Tensor] = None) -> None:
...
+        elif eps == 'trainable':
+            self.logit_eps = nn.Parameter(torch.tensor(0.))
+        else:
+            raise NotImplementedError("eps should be 'trainable' or a float in (0,1)")
+
+    @property
+    def eps(self) -> torch.Tensor:
+        return self.logit_eps.sigmoid()
+
+    def forward(self, x: torch.Tensor, y: torch.Tensor):
```
arnaudvl, line 90: Add `torch.Tensor` return type.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -14,22 +18,76 @@ def __init__(self, sigma: Optional[torch.Tensor] = None) -> None:
...
 
         gamma = 1. / (2. * self.sigma ** 2)   # [Ns,]
         # TODO: do matrix multiplication after all?
         kernel_mat = torch.exp(- torch.cat([(g * dist)[None, :, :] for g in gamma], dim=0))  # [Ns, Nx, Ny]
         return kernel_mat.mean(dim=0)  # [Nx, Ny]
+
+
+class DeepKernel(nn.Module):
+    """"
+    Computes simmilarities as k(x,y) = (1-eps)*k_a(proj(x), proj(y)) + eps*k_b(x,y)
```
arnaudvl, line 63: Apologies in advance for the nitpicking: "similarities".

alibi_detect/utils/pytorch/distance.py
```diff
@@ -28,6 +31,70 @@ def squared_pairwise_distance(x: torch.Tensor, y: torch.Tensor, a_min: float = 1
...
+    with torch.no_grad():
+        k_is = []  # type: Union[list, tuple]
+        for i in range(n_batch):
+            istart, istop = i * batch_size, min((i + 1) * batch_size, n_z)
+            i_batch = z[istart:istop]
+            if isinstance(preprocess_fn, Callable):  # type: ignore
+                i_batch = preprocess_fn(i_batch)
+            i_batch = i_batch.to(device)
+            k_ijs = []
+            for j in range(n_batch):
```
arnaudvl, line 86: Might be worth to add a separate github issue to make this function a bit more efficient? E.g. by reusing `i_batch` when `istart, istop == jstart, jstop` and avoiding duplication of the computation for kxy and kyx by identifying beforehand where x stops and y starts?

alibi_detect/cd/base.py
```diff
@@ -203,8 +209,8 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray, n
         pass
 
     def predict(self, x: Union[np.ndarray, list],  return_p_val: bool = True,
-                return_distance: bool = True, return_probs: bool = True) \
-            -> Dict[Dict[str, str], Dict[str, Union[int, float]]]:
+                return_distance: bool = True, return_probs: bool = True, return_model: bool = True) \
```
arnaudvl, line 212: What is the reason for returning the model? To understand what model exactly resulted in the drift predictions?

alibi_detect/cd/learnt_kernel.py
```diff
@@ -0,0 +1,161 @@
...
+        if backend == 'tensorflow' and has_tensorflow:
+            pop_kwargs = ['device', 'dataloader']
+            [kwargs.pop(k, None) for k in pop_kwargs]
+            if dataset is None:
+                kwargs.update({'dataset': TFDataset})
+            self._detector = LearntKernelDriftTF(*args, **kwargs)  # type: ignore
+        else:
+            if dataset is None:
+                kwargs.update({'dataset': TorchDataset})
+                # kwargs.update({'dataset': TensorDataset})  # TODO: Ok?
```
arnaudvl, line 131: Don't think so with ideally allowed input types `List[Any]` as well as arrays/tensors.

alibi_detect/cd/base.py
```diff
@@ -252,6 +260,181 @@ def predict(self, x: Union[np.ndarray, list],  return_p_val: bool = True,
...
+        Preprocessed reference data and new instances.
+        """
+        if isinstance(self.preprocess_fn, Callable):  # type: ignore
+            x = self.preprocess_fn(x)
+            x_ref = self.x_ref if self.preprocess_x_ref else self.preprocess_fn(self.x_ref)
+            return x_ref, x
+        else:
+            return self.x_ref, x
+
+    def get_splits(self, x_ref: np.ndarray, x: np.ndarray) \
```
arnaudvl, line 351: Input types should be of type `Union[np.ndarray, list]`.

alibi_detect/cd/tensorflow/spot_the_diff.py
```diff
@@ -0,0 +1,203 @@
...
+        dataset
+            Dataset object used during training.
+        data_type
+            Optionally specify the data type (tabular, image or time-series). Added to metadata.
+        """
+
+        if kernel is None:
+            kernel = GaussianRBF(trainable=True)
+        if initial_diffs is None:
+            initial_diffs = np.random.normal(size=(n_diffs,) + x_ref.shape[1:]) * x_ref.std(0)
```
arnaudvl, line 110: So currently only works for arrays? It's fine to have a separate issue open for extensions to `List[Any]` if it is a non-trivial amount of extra work.

arnaudvl: Additionally, it is also assuming continuous numerical inputs which of course makes sense as default given the GaussianRBF kernel. It would however be interesting to see how this approach could work with e.g. text data after initially embedding the text if there is evidence for "king - man + woman = queen"-like properties of embedding vectors. This would require an additional preprocessing step before applying the kernel.

alibi_detect/cd/tensorflow/spot_the_diff.py
```diff
@@ -0,0 +1,203 @@
...
+        train_kwargs
+            Optional additional kwargs when fitting the classifier.
+        dataset
+            Dataset object used during training.
+        data_type
+            Optionally specify the data type (tabular, image or time-series). Added to metadata.
+        """
+
+        if kernel is None:
+            kernel = GaussianRBF(trainable=True)
```
arnaudvl, line 108: Seems to conflict with my previous comment where I stated that `trainable=True` and `self.init_required=True`. So does it train OK then...?

alibi_detect/cd/tensorflow/spot_the_diff.py
```diff
@@ -0,0 +1,203 @@
...
+            verbose: int = 0,
+            train_kwargs: Optional[dict] = None,
+            dataset: Callable = TFDataset,
+            data_type: Optional[str] = None
+    ) -> None:
+        """
+        Classifier-based drift detector with a classifier of form y = a + b_1*k(x,w_1) + ... + b_J*k(x,w_J),
+        where k is a kernel and w_1,...,w_J are learnable test locations. If drift has occured the test locations
+        learn to be more/less (given by sign of b_i) simmilar to test instances than reference instances.
+        The test locations are regularised to be close to the average reference instance such that the **difference**
```
arnaudvl, line 40: Make clear it is the average "feature-wise distance".

alibi_detect/cd/tensorflow/spot_the_diff.py
```diff
@@ -0,0 +1,203 @@
...
+    def __init__(
+            self,
+            x_ref: np.ndarray,
+            p_val: float = .05,
+            preprocess_x_ref: bool = True,
+            update_x_ref: Optional[Dict[str, int]] = None,
+            preprocess_fn: Optional[Callable] = None,
+            kernel: Optional[tf.keras.Model] = None,
+            n_diffs: int = 1,
+            initial_diffs: Optional[np.ndarray] = None,
```
arnaudvl, line 19: It's probably worth opening an issue highlighting best practice or providing intuition about the impact of both `n_diffs` and `initial_diffs` on the feature attributions obtained from the detector. Either in the docs or made clear via an example.

alibi_detect/utils/pytorch/kernels.py
ojcobb: Sorry I'm unsure on both your points here.

On first point -- having `trainable=True` and `sigma=None` is the default for a trainable `GaussianRBF` and we wouldn't want it to lead to an error?

On second point -- I don't think we'd want that to throw an error either? The trainable `GaussianRBF` is always called with `infer_sigma=False`.  We just temporarily need `init_required=True` to override this for the first call.

alibi_detect/cd/__init__.py
ojcobb: Haha I actually think I need to change every instance of 'learnt' to 'learned'. Turns out 'learnt' is standard outside US for the past participle but as an adjective should always be 'learned'.

alibi_detect/utils/pytorch/kernels.py
ojcobb: I guess we could. I thought this was less ambiguous as it is either trainable or the value provided, it can't be both.

And `logit_eps=0` is same as `eps=0.5`.

alibi_detect/cd/base.py
ojcobb: Was useful for implementing `SpotTheDiffDrift` detector as a wrapper around the `ClassifierDrift` detector, but thought could be useful more generally too.

alibi_detect/cd/tensorflow/spot_the_diff.py
ojcobb: Well we still support the `preprocess_batch_fn` etc but obvs we need full differentiability from `initial_diffs` to classifier output so they must be in a form that can be handled by torch/tensorflow. So the diffs would indeed be in the processed space rather than raw space that allows any object.

I guess we could allow the user to specify the inverse of `preprocess_batch_fn` to map the diffs back to the raw space. I guess this will often require some custom logic though such as for text we'd need to take an argmax or something to select words.

alibi_detect/cd/tensorflow/spot_the_diff.py
ojcobb: Indeed -- don't see the problem.

alibi_detect/cd/tensorflow/spot_the_diff.py
ojcobb: Yeah I'll touch on it in the example notebook but would require further investigation to nail down.

alibi_detect/utils/pytorch/kernels.py
arnaudvl: Yes it became clear when it is used in the spot the difference detector. So all good and fine to remove the TODO.

alibi_detect/utils/pytorch/kernels.py
arnaudvl: Yes, my point was more that now if you set it to 'trainable' `eps` is by default 0.5 but I am not sure if that's always a good setting in the trade off between the `kernel_a` and `kernel_b`.

alibi_detect/cd/tensorflow/spot_the_diff.py
arnaudvl: I think we don't really support anything other than arrays since `preprocess_batch_fn` would only be used after we already init the classifier with `x_ref` as an array without preprocessing. As a result adding a `preprocess_batch_fn` would lead to a mismatch in the way `x_ref` is used in `self.mean` and how the instances `x` are used in the forward pass of the classifier since they would have been preprocessed.

arnaudvl: So if there is a `preprocess_batch_fn` it should already be applied before init of `InterpretableClf` to `x_ref` passed to the classifier?

alibi_detect/utils/pytorch/kernels.py
ojcobb: Ah, I don't think it's a problem given that it's trainable. Here the init won't cause a 0 gradient problem like it does for `sigma` in `GaussianRBF`.

arnaudvl:
In order to make the `SpotTheDiffDrift` detector work with preprocessing step, I believe we should:

- Apply the preprocessing step to `x_ref` before initializing `initial_diffs` if `initial_diffs is None` and use the preprocessed data for `initial_diffs` and as input to the classifier model.
- For the `ClassifierDriftTorch` detector:
  - Pass the original `x_ref` data 
  - Set `update_x_ref` to False
  - Set `preprocess_x_ref` to False because `preprocess_batch_fn` is applied when using the trainer and prediction function (`self.predict_fn`) in the classifier detector
  - Set `preprocess_fn` to None
- As a result of the above defaults in `self._detector`:
  - Remove `update_x_ref`, `preprocess_x_ref`, `preprocess_fn` from `SpotTheDiffDrift`, `SpotTheDiffDriftTorch` and `SpotTheDiffDriftTF`.

alibi_detect/cd/tensorflow/learnt_kernel.py
```diff
@@ -0,0 +1,208 @@
...
+            h_mat = k_xx + k_yy - k_xy - tf.transpose(k_xy)
+
+            mmd2_est = (tf.reduce_sum(h_mat)-tf.linalg.trace(h_mat))/(len(x)*(len(x)-1))
+            var_est = (4*tf.reduce_sum(tf.reduce_sum(h_mat, axis=-1)**2)/(len(x)**3) -
+                       4*tf.reduce_sum(h_mat)**2/(len(x)**4))
+            reg_var_est = var_est + self.var_reg
+
+            return mmd2_est/tf.math.sqrt(reg_var_est)
+
+    def score(self, x: np.ndarray) -> Tuple[float, float, np.ndarray]:
```
arnaudvl: Change type to `Union[np.ndarray, list]`?

alibi_detect/cd/tensorflow/learnt_kernel.py
```diff
@@ -0,0 +1,208 @@
...
+                for _ in range(self.n_permutations)]
+        )
+        p_val = (mmd2 <= mmd2_permuted).mean()
+        return p_val, mmd2, mmd2_permuted
+
+    @staticmethod
+    def trainer(
+        j_hat: JHat,
+        datasets: Tuple[tf.keras.utils.Sequence, tf.keras.utils.Sequence],
+        optimizer: tf.keras.optimizers = tf.keras.optimizers.Adam,
```
arnaudvl: The default optimizer requires an initialized optimizer, e.g. `tf.keras.optimizers.Adam(learning_rate=1e-3)`.

alibi_detect/cd/pytorch/spot_the_diff.py
```diff
@@ -0,0 +1,213 @@
...
+        """
+        if preprocess_fn is not None and preprocess_batch_fn is not None:
+            raise ValueError("SpotTheDiffDrift detector only supports preprocess_fn or preprocess_batch_fn, not both.")
+
+        if preprocess_fn is not None:
+            x_ref_proc = preprocess_fn(x_ref)
+        elif preprocess_batch_fn is not None:
+            x_ref_proc = predict_batch(
+                x_ref, lambda x: x, preprocess_fn=preprocess_batch_fn,
+                device=torch.device('cpu'), batch_size=batch_size
```
arnaudvl, line 118: We can infer the device already and then pass `device.type` to the `ClassifierDriftTorch` detector.

alibi_detect/cd/tensorflow/learnt_kernel.py
ojcobb: Ok I'll change the default I think as I'm worried about having a single optimizer with continued state between separate calls to predict.

alibi_detect/cd/pytorch/spot_the_diff.py
ojcobb: But that's for applying the model whereas this is just for applying `lambda x: x`.

ojcobb: Given the preds are instantly returned to the cpu it seemed silly potentially transferring them to a gpu to apply the identity.

arnaudvl: Ah yes, I forgot that the device comes only in use after the preprocessing in predict batch, my bad!

examples/cd_spot_the_diff.ipynb
```diff
@@ -0,0 +1,810 @@
...
+    "\n",
+    "The `SpotTheDiffDrift` detector is designed to make this process straightforward. Like the `ClassifierDrift` detector, it uses a portion of the available data train a classifier to discriminate between reference and test instances. Letting $\\hat{p}_T(x)$ represent the probability assigned by the classifier that the instance $x$ is from the test set rather than reference set, the difference here is that we use a classifier of the form $$\\text{logit}(\\hat{p}_T) = b_0 + b_1 k(x,w_1) + ... + b_Jk(x,w_J),$$ where $k(\\cdot,\\cdot)$ is a kernel specifying a notion of similarity between instances, $w_i$ are learnable *test locations* and $b_i$ are learnable regression coefficients.\n",
```
arnaudvl, line 11: Typo: "to train"

examples/cd_spot_the_diff.ipynb
```diff
@@ -0,0 +1,810 @@
...
+    "%matplotlib inline"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "MNIST_PATH = '/home/oliver/Projects/alibi-detect/examples'\n",
```
arnaudvl, line 51: Set to generic path.

examples/cd_spot_the_diff.ipynb
```diff
@@ -0,0 +1,810 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import torch\n",
+    "import tensorflow as tf\n",
```
arnaudvl, line 33: Any tensorflow reference can be removed as I believe on pytorch implementation is used.

examples/cd_spot_the_diff_mnist_wine.ipynb
```diff
@@ -0,0 +1,810 @@
+{
```
arnaudvl, line 1: This notebook to be removed as it is a duplicate (but without NME reference) as the other notebook?

doc/source/methods/learnedkerneldrift.ipynb
```diff
@@ -0,0 +1,202 @@
...
+   "source": [
+    "## Usage\n",
+    "\n",
+    "### Initialize\n",
+    "\n",
+    "Arguments:\n",
+    "\n",
+    "* `x_ref`: Data used as reference distribution.\n",
+    "\n",
+    "* `kernel`: A differentiable **TensorFlow** or **PyTorch** module that takes two instances as input and returns a scalar notion of similarity os output.\n",
```
arnaudvl, line 37: Typo: 'os' -> 'as'

doc/source/methods/spotthediffdrift.ipynb
```diff
@@ -0,0 +1,221 @@
...
+    "\n",
+    "* `verbose`: Verbosity level during the training of the kernel. 0 is silent and 1 prints a progress bar.\n",
+    "\n",
+    "* `train_kwargs`: Optional additional kwargs for the built-in TensorFlow (`from alibi_detect.models.tensorflow import trainer`) or PyTorch (`from alibi_detect.models.pytorch import trainer`) trainer functions.\n",
+    "\n",
+    "* `dataset`: Dataset object used during training of the kernel. Defaults to `alibi_detect.utils.pytorch.data.TorchDataset` (an instance of `torch.utils.data.Dataset`) for the PyTorch backend and `alibi_detect.utils.tensorflow.data.TFDataset` (an instance of `tf.keras.utils.Sequence`) for the TensorFlow backend. For PyTorch, the dataset should only take the windows x_ref and x_test as input, so when e.g. *TorchDataset* is passed to the detector at initialisation, during training *TorchDataset(x_ref, x_test)* is used. For TensorFlow, the dataset is an instance of `tf.keras.utils.Sequence`, so when e.g. *TFDataset* is passed to the detector at initialisation, during training *TFDataset(x_ref, x_test, batch_size=batch_size, shuffle=True)* is used. x_ref and x_test can be of type np.ndarray or List[Any].\n",
```
arnaudvl, line 81: This is the description for LearnedKernelDrift. I think for SpotTheDiffDrift the description from ClassifierDrift should be used since we pass (x, y) to the dataset, not (x_ref, x_test).

doc/source/methods/spotthediffdrift.ipynb
```diff
@@ -0,0 +1,221 @@
...
+    "\n",
+    "The prediction takes the form of a dictionary with `meta` and `data` keys. `meta` contains the detector's metadata while `data` is also a dictionary which contains the actual predictions stored in the following keys:\n",
+    "\n",
+    "* `is_drift`: 1 if the sample tested has drifted from the reference data and 0 otherwise.\n",
+    "\n",
+    "* `diffs`: a numpy array containing the diffs used to discriminate reference from test instances.\n",
+    "\n",
+    "* `diff_ceoffs` a coefficient correspond to each diff where a coeffient greater than zero implies that the corresponding diff makes the average references instances *more* similar to a test instance on average and less than zero implies *less* similar.\n",
```
arnaudvl, line 170: Typo: 'drift_ceoffs' -> 'drift_coeffs'

doc/source/overview/getting_started.md
```diff
@@ -99,14 +101,18 @@ And **data drift**:
 * [Least-Squares Density Difference Drift Detector](../methods/lsdddrift.ipynb)
 
 * [Maximum Mean Discrepancy Drift Detector](../methods/mmddrift.ipynb)
+*
```
arnaudvl, line 104: Remove empty bullet point

doc/source/overview/getting_started.md
```diff
@@ -99,14 +101,18 @@ And **data drift**:
...
 * [Maximum Mean Discrepancy Drift Detector](../methods/mmddrift.ipynb)
+* 
+* [Learned Kernel MMD Drift Detector](../methods/learnedkerneldrift.ipynb)
 
 * [Chi-Squared Drift Detector](../methods/chisquaredrift.ipynb)
 
 * [Mixed-type Tabular Data Drift Detector](../methods/tabulardrift.ipynb)
 
 * [Classifier Drift Detector](../methods/classifierdrift.ipynb)
+*
```
arnaudvl, line 112: Remove empty bullet point and fix spot-the-diff link

alibi_detect/cd/base.py
```diff
@@ -79,6 +84,9 @@ def __init__(
         if preds_type not in ['probs', 'logits']:
             raise ValueError("'preds_type' should be 'probs' or 'logits'")
 
+        if n_folds is not None and n_folds > 1 and not retrain_from_scratch:
+            raise ValueError("If using multilpe folds the model must be retrained from scratch for each fold.")
```
arnaudvl, line 88: Typo: 'multiple'

PR #314, opened 2021-08-23
Allow for DeepKernel without kernel_b component.

ojcobb:
Is now possible to set `kernel_b=None`. 

This seemed preferable to allowing both a kernel_b to be specified and `eps=0`.

### DIFF
--- a/alibi_detect/utils/tensorflow/kernels.py
+++ b/alibi_detect/utils/tensorflow/kernels.py
@@ -87,7 +87,7 @@ class DeepKernel(tf.keras.Model):
         self,
         proj: tf.keras.Model,
         kernel_a: tf.keras.Model = GaussianRBF(trainable=True),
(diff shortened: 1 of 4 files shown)


alibi_detect/utils/tensorflow/kernels.py
```diff
@@ -77,9 +77,10 @@ class DeepKernel(tf.keras.Model):
         The kernel to apply to the projected inputs. Defaults to a Gaussian RBF with trainable bandwidth.
     kernel_b
         The kernel to apply to the raw inputs. Defaults to a Gaussian RBF with trainable bandwidth.
+        Set to None in order to use only the deep component (i.e. eps=0).
```
arnaudvl, line 80: Change type of `kernel_b` to `Optional[tf.keras.Model ]`

Issue #313, opened 2021-08-16
Fix various embedded images across notebooks


arnaudvl:
E.g. https://docs.seldon.io/projects/alibi-detect/en/latest/examples/cd_online_camelyon.html or https://docs.seldon.io/projects/alibi-detect/en/latest/examples/alibi_detect_deploy.html.

ascillitoe:
@arnaudvl I have a fix for this in the latest doc updates, I'll include it in the main commit for those updates.

PR #319, opened 2021-08-30
Minor fixes to docs, and docs builds added to CI

ascillitoe:
This PR contains a few minor fixes to the docs. Included in this are fixes to various images embedded across notebooks ( https://github.com/SeldonIO/alibi-detect/issues/313). See the gotcha's [here](https://github.com/SeldonIO/alibi/tree/master/doc) for more info on embedding images in notebooks.

Additionally, html and pdf doc builds have been added to CI. The docs are being built with the standard sphinx make commands. This should catch many docs errors, but tighter emulation of the readthedocs build process might be useful in the future.


.github/workflows/ci.yml
```diff
@@ -62,9 +62,17 @@ jobs:
         run: |
           sudo apt-get install pandoc
           pip install -r requirements/docs.txt
+          pip install -e .
```
jklaise, line 65: Is it necessary to install `alibi` and all its runtime dependencies here? We need to keep in mind also that on `readthedocs` the available dependencies are limited to `requirements/docs.txt` + `pandoc`.

ascillitoe: Thanks, removed this.

jklaise:
To recap what we discussed offline:

- We should make sure the build environments between Github Actions / RTD / local are as close as possible, e.g. current CI installs `pandoc==2.5` whilst RTD used `pandoc=1.19.2`, this results in discrepancies and false results.
- To this end, we will look to use the RTD `latest` docker image (`readthedocs/build:6.0`) for our GA build step
- This also requires an update to `.readthedocs.yml` to use `latest` image
- We also enabled RTD build on PR, however this is limited (not building PDF, not building on merge/release), so we're keen to keep our GA docs build, we just need to make an effort to have the environment on GA be as close (ideally identical) to RTD as possible

ascillitoe:
The latest commit is now performing the GA docs build on the `latest` RTD docker image. The builds pass, but importantly the below errors that were being missed during CI are now occurring (this is good!):
```
/__w/alibi-detect/alibi-detect/doc/source/examples/alibi_detect_deploy.nblink:137: WARNING: Undefined substitution referenced: "vae_lillog.png".
/__w/alibi-detect/alibi-detect/doc/source/examples/alibi_detect_deploy.nblink:439: WARNING: Undefined substitution referenced: "deploy_diagram.png".
``` 

Ideally it would be nice to turn warnings into errors (for the html build) so warnings like the above are automatically flagged, but we still need to fix the `WARNING: duplicate object description` first. For the pdf build there are far too many warnings to fix!

Also, @jklaise do we still need to explicitly set `pandoc` in `readthedocs.yml`? I can't see it hurting?

Issue #323, opened 2021-09-02
Example notebooks out of date


ascillitoe:
The recent api updates (i.e. separating `models`, `utils.prediction` etc into `tensorflow` and `pytorch`) mean a number of the example notebooks are now out-of-date. Some examples below:

- Many examples load in `predict_batch` as `from alibi_detect.utils.prediction import predict_batch` instead of `from alibi_detect.utils.tensorflow.prediction import predict_batch`. The arguments and use of `predict_batch` also needs updating i.e. `clf` is now the second argument, and the absence of `return_class` arg means these notebooks need reworking to convert prediction proba's to classes (or this functionality needs to be added back in internally).  e.g. see `examples/cd_distillation_cifar10.ipynb`.
- Some fetch methods (such as `fetch_vaegmm` used in `examples/od_aegmm_kddcup.ipynb` fail when attempting to load state from `meta.pickle` files. These files need to be updated to match the new api (or this can be fixed on the code side with sub-classes, aliases etc). 
- General updates to imports in notebooks needed.

The above is not exhaustive. A general sweep to check all the examples is needed.

jklaise:
We'll be looking into notebook testing (maybe a new issue?) as a longer-term initiative. Example of how this is done in Seldon Core:

- https://github.com/SeldonIO/seldon-core/blob/3cfe519ae918acc33690ed9da5390093135ca035/testing/scripts/seldon_e2e_utils.py#L645 (create script from notebook)
- https://github.com/SeldonIO/seldon-core/blob/master/notebooks/convert.tpl (conversion template, has a lot of sleep calls between cells because their examples make network requests that can time out, we don't need to do this)
- https://github.com/SeldonIO/seldon-core/blob/master/testing/scripts/test_notebooks.py (run script)

jklaise:
Many are fixed in this PR: #333.

The following list are outstanding:
 - `od_if_kddcup` - the saved version of the model must be using an older `sklearn` version which triggers a `ModuleNotFoundError: No module named 'sklearn.ensemble.iforest'` as it seems they shuffled around internal modules - needs updating the saved detector
 - `od_vae_adult.ipynb` - due to `load_outlier_detector=False`, the notebook trains and attempts to save a detector but fails because of #335 .
 - `od_aegmm_kddcup.ipynb` - outdated metatada files causing `fetch` to fail
 - `od_prophet_weather.ipynb` - outdated metadata files causing `fetch` to fail

[...]

Issue #318, opened 2021-08-29
No such file or directory meta.dill


matteoguarrera:
When testing the _**VAE outlier detection on CIFAR10**_ notebook I got this error downloading the pretrained model remotelly. No such file or directory meta.dill
Am I doing something wrong?

ascillitoe:
Hi @matteoguarrera, thanks for flagging this. It does look like it might be a minor bug with the `fetch_detector` method. We will address this ASAP but if you want to keep going with the notebook in the meantime, you could either manually create the `my_path/OutlierVAE` directory first, or set `load_outlier_detector = False` to train a new detector from scratch.

matteoguarrera:
The fastest way to overcome this issue is to fetch a previous commit. 
`pip install git+git://github.com/SeldonIO/alibi-detect.git@e58a3af0e5f983fdbcdd4259c3fd843732c0084f
`

ascillitoe:
Ah ha good spot. Seems like there are actually a few different issues here:

1. Use of `os.mkdir()` in `fetch_detector` can't create the nested directory i.e. relies on `my_path` already being created.
2. I am having trouble with `tf.keras.utils.get_file()` (used in `fetch_enc_dec`) when trying to set `fname` to an absolute path. i.e. 
`tf.keras.utils.get_file(fname='my_path/OutlierVAE/model/encoder_net.h5',
                  origin='https://storage.googleapis.com/seldon-models/alibi-detect/od/OutlierVAE/cifar10/model/encoder_net.h5')` 
                  
   results in:  
                  `FileNotFoundError: [Errno 2] No such file or directory: '/home/ascillitoe/.keras/datasets/my_path/OutlierVAE/model/encoder_net.h5'`

    but `tf.keras.utils.get_file(fname='encoder_net.h5',...)` works OK.
    
3. Since https://github.com/SeldonIO/alibi-detect/commit/65ed410d102c4c8f61de07c0c7c759fc609987b9 we are working with `meta.dill` and `<model>.dill`, but we are still fetching `.pickle` files from the google bucket with `fetch_detector`. This will throw up an error when `load_detector` is called in `fetch_detector`.

@matteoguarrera I presume you're actually only encountering the third issue? Hence why rolling back to a commit before https://github.com/SeldonIO/alibi-detect/commit/65ed410d102c4c8f61de07c0c7c759fc609987b9 works for you.

ascillitoe:
Re fixes:
[...]

PR #324, opened 2021-09-02
Fix fetching methods

ascillitoe:
While looking into https://github.com/SeldonIO/alibi-detect/issues/318 a number of other issues were found with the fetch methods:

1. Use of `os.mkdir()` in `fetch_detector` can't create the nested directory i.e. relies on `my_path` already being created.
2. `tf.keras.utils.get_file(fname=...,origin=...)` used internally doesn't save the fetched file in the desired directory unless `fname` is properly set as an absolute path.
3. Since https://github.com/SeldonIO/alibi-detect/commit/65ed410d102c4c8f61de07c0c7c759fc609987b9 we are working with `meta.dill` and `<model>.dill`, but we are still fetching `.pickle` files from the google bucket with `fetch_detector`. 

This PR fixes the above by:
1. Switching to `pathlib` for filesystem operations (this should also give better cross-OS support). `os` is still used for handling url's as `pathlib` can make things more complicated with url's. 
2. `pathlib.resolve()` is used to make sure absolute paths are always passed to `tf.keras.get_file()`.
3. When loading/fetching from the google bucket, `.dill` files are loaded if they exist, otherwise we revert to loading `.pickle.`

The above has been tested on a number of notebooks. Most of the fetch methods work without issue now, but one or two such as `fetch_vaegmm` used in `examples/od_aegmm_kddcup.ipynb` fail. Error here is `ModuleNotFoundError: No module named 'alibi_detect.models.autoencoder'` which occurs when attempting to load state from `meta.pickle`, so think this is more due to the metadata files being generated with the previous api i.e. before `models` was separated into `pytorch` and `tensorflow`. @jklaise @arnaudvl  - with this in mind I don't know whether we want to think about merging these fixes in now or wait until we update all of the example notebooks as a larger exercise (see https://github.com/SeldonIO/alibi-detect/issues/323).


jklaise:
Related, can we replace `cloudpickle` with `pickle` from `alibi_detect.datasets` as well @arnaudvl ? https://github.com/SeldonIO/alibi-detect/search?q=cloudpickle

jklaise:
I think it makes sense to merge this even if it doesn't fix all the notebook issues and we can pick up #323 as another PR.

alibi_detect/utils/saving.py
```diff
@@ -960,11 +961,19 @@ def load_detector(filepath: str, **kwargs) -> Data:
...
 
     # check if path exists
-    if not os.path.isdir(filepath):
+    filepath = Path(filepath)
+    if not filepath.is_dir():
         raise ValueError('{} does not exist.'.format(filepath))
 
+    # Check if dill files exist, otherwise read pickle files
+    files = list(filepath.glob('*.dill'))
+    if len(files) == 2:
```
jklaise, line 970: Where does the 2 come from?

jklaise (approved):
LGTM, left one comment. I would also swap out the one remaining `cloudpickle` usage in `alibi_detect.datasets` (check if it works post-change).

alibi_detect/utils/saving.py
ascillitoe: 2 as we expect a `meta.dill` and `<model>.dill` file in the directory. Admittedly this isn't particularly robust, but I'm assuming we have tight control over what exists in a given google bucket directory. I'll change this to check if `meta.dill` and `<model>.dill` exist, otherwise check if `.pickle` files exist, otherwise raise an error.

PR #325, opened 2021-09-02
Feature docs

ascillitoe:
Work in progress to update docs layout, including a background to drift detection page.


doc/source/overview/getting_started.md
```diff
@@ -165,4 +165,4 @@ preds = od.predict(X_test)
 The predictions are returned in a dictionary with as keys `meta` and `data`. `meta` contains the detector's metadata while `data` is in itself a dictionary with the actual predictions. It has either `is_outlier`, `is_adversarial` or `is_drift` (filled with 0's and 1's) as well as optional `instance_score`, `feature_score` or `p_value` as keys with numpy arrays as values.
```
arnaudvl, line 165: Mention that other relevant values can also be returned in the `data` dict as values.

doc/source/overview/algorithms.md
```diff
@@ -6,38 +6,38 @@ The following tables summarize the advised use cases for the current algorithms.
```
arnaudvl, line 6: Mention that the *Feature Level* column is also relevant for drift detection. As discussed we can improve the tables further in a followup PR.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "* [3. Online detection](#online)\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 0. What is drift? <a class=\"anchor\" id=\"what_is_drift\"></a>\n",
+    "\n",
+    "Although powerful, modern machine learning models can be sensitive. Seemingly subtle changes in a data distribution can destroy the performance of otherwise state-of-the art models, which can be especially problematic when ML models are deployed in production. Typically, ML models are tested on held out data in order to estimate their furture performance. Crucially, this assumes that the *process* underlying the input data $\\mathbf{X}$ and output data $\\mathbf{Y}$ remains constant. \n",
```
arnaudvl, line 27: Typo: furture -> future

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "*Drift* is said to occur when the process underlying $\\mathbf{X}$ and $\\mathbf{Y}$ during deployment differs from the process that generated the training data. In this case, we can no longer expect the model's performance during deployment to match that observed on held out training data. If a *ground truth* is available during deployment, *supervised drift detection* can be performed, with the model's predictive performance monitored directly. However, in many deployment scenarios, such as the binary classification example below, a ground truth is not available and *unsupervised drift detection* methods are required.\n",
```
arnaudvl, line 29: While deployment is a natural setting for us, I would probably refer to test time instead since drift is not strictly related to deployed models.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "![drift in deployment](images/drift_deployment.png) \n",
+    "\n",
+    "To explore the different types of drift, consider the common scenario where we have input data $\\mathbf{X}$ and output data $\\mathbf{Y}$, jointly distributed according to $P(\\mathbf{X},\\mathbf{Y})$, and a model $f: x \\mapsto y$ trained on training/reference data with distribution $P_{ref}(\\mathbf{X},\\mathbf{Y})$. *Drift* is said to have occured when $P(\\mathbf{X},\\mathbf{Y}) \\ne P_{ref}(\\mathbf{X},\\mathbf{Y})$. Writing the joint distribution as\n",
```
arnaudvl, line 33: This is actually an important nuance: the training data should be disjoint from the reference data for a pretrained preprocessing step (eg encoder, model used for the uncertainty detectors etc).

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "$$\n",
+    "P(\\mathbf{X},\\mathbf{Y}) = P(\\mathbf{Y}|\\mathbf{X})P(\\mathbf{X}) = P(\\mathbf{X}|\\mathbf{Y})P(\\mathbf{Y}),\n",
+    "$$\n",
+    "\n",
+    "we can categorise drift into a number of types:\n",
+    "\n",
+    "- **Covariate drift**: Also referred to as input drift, this occurs when $P(\\mathbf{Y}|\\mathbf{X})$ = $P_{ref}(\\mathbf{Y}|\\mathbf{X})$, but the distribution of the input data has shifted $P(\\mathbf{X}) \\ne P_{ref}(\\mathbf{X})$. In a mortgage risk classification task, an example of this would be if at inference, applications for properties with more bedrooms than the properties present in the training data were considered. The model may now be extrapolating, potentially resulting in unreliable predictions. \n",
```
arnaudvl, line 41: Consider changing "at inference" with eg "at test time" given the ambiguous nature of the usage of inference.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "- **Prior drift**: Also referred to as label drift, this occurs when $P(\\mathbf{X}|\\mathbf{Y})=P_{ref}(\\mathbf{X}|\\mathbf{Y})$, but the distribution of the outputs has shifted $P(\\mathbf{Y}) \\ne P_{ref}(\\mathbf{Y})$. Using the same example again, an increase in the proportion of applications classsed as \"approved\" would be an example of prior drift. The model could still be accurate, but accuracy metrics might be affected. \n",
+    "\n",
+    "- **Concept drift**: This occurs when the process generating $y$ from $x$ has changed, such that $P(\\mathbf{Y}|\\mathbf{X}) \\ne P_{ref}(\\mathbf{Y}|\\mathbf{X})$. An example of this could be if a change in the wider economic climate resulted in a change in the risk for a given mortgage application. It is possible that the model might no longer give a suitable approximation of the true process. \n",
+    "\n",
```
arnaudvl, line 46: In this section we might already want to mention that prior/concept drift require access to ground truth which is typically not available so we need to find proxies (eg prediction distribution shift for prior drift).

arnaudvl: In general I quite like the symptoms/disease example from the introduction of [Lipton et al.](https://arxiv.org/abs/1802.03916) to illustrate the *P(X,Y)* breakdown.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "It is relatively easy to spot drift by eyeballing these figures here. However, the task becomes considerably harder for high-dimensional real problems, especially since real-time ground truths are not typically available. "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 1. Detecting drift <a class=\"anchor\" id=\"drift_detection\"></a>\n",
+    "\n",
+    "[Alibi Detect](https://github.com/SeldonIO/alibi-detect) offers a wide array of methods for detecting drift (see [here](../overview/algorithms.md)), many of which are examined in the NeurIPS 2019 paper [Failing Loudly: An Empirical Study of Methods for Detecting Dataset Shift](https://arxiv.org/abs/1810.11953). Generally, these aim to determine whether the distribution $P(\\mathbf{x})$ has drifted from a reference distribution $P_{ref}(\\mathbf{x})$, where $\\mathbf{x}$ may represent input data $\\mathbf{X}$, true output data $\\mathbf{Y}$, or some form of model output, depending on what type of drift we wish to detect. \n",
```
arnaudvl, line 60: "many of which" -> "some of which" :)
We've implemented and tested much more than what is talked about in that paper.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "Due to natural randomness in the process being modelled, we don't necessarily expect observations $\\mathbf{x}_1,\\dots,\\mathbf{x}_N$ drawn from $P(\\mathbf{x})$ to be identical to $\\mathbf{x}^{ref}_1,\\dots,\\mathbf{x}^{ref}_M$ drawn from $P_{ref}(\\mathbf{x})$. To decide whether differences between $P(\\mathbf{x})$ and $P_{ref}(\\mathbf{x})$ are due to drift or just natural randomness in the data, *statistical two-sample hypothesis* testing is used, with the null hypothesis $P(\\mathbf{x})=P_{ref}(\\mathbf{x})$. If the p-value obtained is below a given threshold, the null is rejected and the alternative hypothesis $P(\\mathbf{x}) \\ne P_{ref}(\\mathbf{x})$ is accepted, suggesting drift is occuring. \n",
+    "\n",
+    "Since $\\mathbf{x}$ is often high-dimensional (even a 200 x 200 greyscale image has 40k dimensions!), performing hypothesis testing in the full-space is often computationally intractable. Instead, the pipeline below is often used, with dimension reduction as a pre-processing step.  \n",
```
arnaudvl, line 64: Intractable or unsuitable for the statistical test which is performed. E.g. a K-S test can pretty quickly be computed on relatively high dimensions, but this will not result in a great test because it just compares a lot of individual stats on the marginals and average them somehow.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "Hypothesis testing involves first choosing a *test statistic* $S(\\mathbf{x})$, which is expected to be small if the null hypothesis $H_0$ is true, and large if the alternative $H_a$ is true. For observed data $\\mathbf{x}$, $S(\\mathbf{x})$ is computed, followed by a p-value $\\hat{p} = P(\\text{such an extreme } S(\\mathbf{x}) | H_0)$. In other words, $\\hat{p}$ represents the probability of such an extreme value of $S(\\mathbf{x})$ occuring given that $H_0$ is true. Typically results are said to be *statistically significant*, and the null $P(\\mathbf{x})=P_{ref}(\\mathbf{x})$ is rejected, when $\\hat{p}\\le \\alpha$. Conveniently, the threshold $\\alpha$ represents the desired false positive rate. \n",
+    "\n",
+    "The *test statistics* available in [Alibi Detect](https://github.com/SeldonIO/alibi-detect) can be broadly split into two categories; univariate and multivariate tests.\n",
+    "\n",
+    "* Univariate:\n",
+    "    * [Chi-Squared](methods/chisquaredrift.ipynb) (for categorical data)\n",
+    "    * [Kolmogorov-Smirnov](methods/ksdrift.ipynb)\n",
+    "\n",
+    "* Multivariate:\n",
```
arnaudvl, line 78: You could add the classifier/uncertainty-based detectors here as well.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "Since the univariate tests examine the feature-wise marginal distributions, they may miss drift in cases where the joint distribution over all $d$ features has changed, but the marginals have not. The multivariate tests, which use permutation testing to extract p-values, offer a more powerful alternative at the cost of greater complexity."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Dimension reduction <a class=\"anchor\" id=\"dimension_reduction\"></a>\n",
+    "\n",
+    "Given an input dataset $\\mathbf{X}\\in \\mathbb{R}^{N\\times d}$, where $N$ is the number of observations and $d$ the number of dimensions, the aim is the reduce the data dimensionality from $d$ to $K$, where $K\\ll d$. The lower dimensional data $\\hat{\\mathbf{X}}\\in \\mathbb{R}^{N\\times K}$ is then monitored by the drift detector. The paper [Failing Loudly: An Empirical Study of Methods for Detecting Dataset Shift](https://arxiv.org/abs/1810.11953) presents a number of dimension reduction approaches, which can be broadly categorised under:\n",
```
arnaudvl, line 93: Typo: the reduce -> to reduce

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "1. Linear projections\n",
```
arnaudvl, line 95: Might be worth considering the following categories:
1. Linear projections
2. Non-linear projections, optionally learned (incl. both random and trained encoders)
3. Feature maps extracted from the model (incl. model output layer which would be the only thing to extract from a truly black box model which you can only query but don't know the internals of)

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "2. Auto-encoders\n",
+    "3. Black-box shift detectors\n",
+    "\n",
+    "[Alibi Detect](https://github.com/SeldonIO/alibi-detect) allows for a high degree of flexibility here, with a user's chosen dimension reduction technique able to be incorperated into their chosen detector via the `preprocess_fn` and `preprocess_x_ref` arguments. In the following sections, the three categories of technqiues are briefily introduced. For more details, see the [examples](examples.rst). \n",
```
arnaudvl, line 99: Typo: technqiues -> techniques

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "\n",
+    "#### Linear projections \n",
+    "\n",
+    "This includes dimension reduction techniques such as [principal component analysis (PCA)](https://en.wikipedia.org/wiki/Principal_component_analysis) and [sparse random projections (SRP)](https://doi.org/10.1145/1150402.1150436). These techniques involve using a transformation or projection matrix $\\mathbf{R}$ to reduce the dimensionality of a given data matrix $\\mathbf{X}$, such that $\\hat{\\mathbf{X}} = \\mathbf{XR}$. A straightforward way to include such techniques as a pre-processing stage is to pass them to the detectors via the `preprocess_fn` argument, for example for the `scikit-learn` library's `PCA` class: \n",
+    "\n",
+    "```python\n",
+    "pca = PCA(2)\n",
+    "pca.fit(X_train)\n",
+    "detector = MMDDrift(X_train, backend='tensorflow', p_val=.05, preprocess_fn=pca.transform)\n",
```
arnaudvl, line 109: Separate `x_ref` from `x_train`.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "```\n",
+    "\n",
+    "#### Auto-encoders \n",
+    "\n",
+    "These non-linear dimension reducing representations consist of an encoder function $\\phi : \\mathcal{X} \\mapsto \\mathcal{H}$ and a decoder function $\\psi : \\mathcal{H} \\mapsto \\mathcal{X}$, where the latent space $\\mathcal{H}$ has lower dimensionality than the input space $\\mathcal{X}$. The output of the encoder $\\hat{\\mathbf{X}} \\in \\mathcal{H}$ can then be monitored by the drift detector. Training involves learning both the encoding function $\\phi$ and the decoding function $\\psi$, in order to reduce the reconstruction loss: $\\phi, \\psi = \\text{arg} \\min_{\\phi, \\psi}\\, \\lVert \\mathbf{X}-(\\phi \\circ \\psi)\\mathbf{X}\\rVert^2$. However, either trained (TAE) or untrained autoencoders (UAE) can be used. A `tensorflow` autoencoder can be incorporated into a detector by packaging it as a callable function using `preprocess_drift` (from [alibi_detect.cd.tensorflow.preprocess](../api/alibi_detect.cd.tensorflow.preprocess.rst)) and `partial` (see [here](https://docs.python.org/3/library/functools.html#functools.partial)), for example:\n",
```
arnaudvl, line 114: Would highlight that MSE is just an example of a training loss used since there are multiple ways to train an AE. Also the similar comment for `X_train` as before, making sure that `X_ref` is used everywhere and should not overlap with `X_train` for pretrained preprocessing functions.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "![BBSD](images/BBSD.png) \n",
+    "\n",
+    "[Failing Loudly: An Empirical Study of Methods for Detecting Dataset Shift](https://arxiv.org/abs/1810.11953) finds BBSD's to be a very effective method, which is encouraging since BBSD's allow the user to repurpose existing black-box models for use as drift detectors. The syntax for incorporating BBSD's into drift detectors is similar to previous autoencoder example, with the added step of using `HiddenOutput` ([alibi_detect.cd.tensorflow.preprocess](../api/alibi_detect.cd.tensorflow.preprocess.rst) or [alibi_detect.cd.pytorch.preprocess](../api/alibi_detect.cd.pytorch.preprocess.rst)) to select the network layer to extract outputs from. The code snippet below is borrowed from [Maximum Mean Discrepancy drift detector on CIFAR-10](../examples/cd_mmd_cifar10.ipynb), where the softmax layer of the well-known [ResNet-32](https://arxiv.org/pdf/1512.03385.pdf) model is fed into an `MMDrift` detector.\n",
```
arnaudvl, line 130: Typo: MMDrift -> MMDDrift

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "```python\n",
+    "encoder_net = tf.keras.Sequential([...])\n",
+    "preprocess_fn = partial(preprocess_drift, model=encoder_net, batch_size=512)\n",
+    "detector = MMDDrift(X_train, backend='tensorflow', p_val=.05, preprocess_fn=preprocess_fn)\n",
+    "```\n",
+    "\n",
+    "Similar functionality can be achieved with `pytorch` models using the [alibi_detect.cd.pytorch.preprocess](../api/alibi_detect.cd.pytorch.preprocess.rst) utilities.\n",
+    "  \n",
+    "#### Black-box Shift Detectors\n",
```
arnaudvl, line 124: This section could be generalized to feature map extraction where the softmax output of a model (or its labels) could be an efficient example of it?

arnaudvl:
An additional section worth adding could be "input preprocessing" where we could illustrate that certain modalities cannot be processed straight away for effective drift detection. E.g. strings or tokens need to be embedded (+include illustration of built-in functionality for these embeddings), same for graphs. Although this would also be achieved via the preprocessing functionality like the dimensionality reduction techniques, the goal is different (just to get the data in a suitable format to apply a detector on).

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "    y = np.zeros(X.shape[0])\n",
+    "    y[idx] = 1\n",
+    "    return y"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The 2D input data $\\mathbf{x}=(x_1,x_2)$ is generated by sampling from a mixture of two standard normal distributions:\n",
```
arnaudvl, line 197: Standard normal implies *N(0,1)* so change to Normal or Gaussian distributions.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "For our model, we choose the well-known, but simple, decision tree classifier. As well as training the model, this is a good time to initialise and \"train\" the MMD detector by calling:\n",
+    "\n",
+    "```python\n",
+    "detector = MMDDrift(X_train, backend='tensorflow', p_val=.05)\n",
```
arnaudvl, line 288: Might be good to mix it up with a `pytorch` backend sometimes just to highlight that this is also possible.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "{'data': {'is_drift': 0,\n",
+       "  'distance': -0.004682348824606164,\n",
```
arnaudvl, line 401: A little bit annoying that the distance value is negative here for illustration purposes...

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+   ],
+   "source": [
+    "detector.predict(X_test)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As expected, since we are sampling from the distributions, the detector don't detect any covariate shift here. More specifically, the detector's p-value is above the threshold of $\\alpha=0.05$, indicating that no statistically significant drift has been detected."
```
arnaudvl, line 424: Typo: don't -> doesn't

arnaudvl: Mention that the distance (*S(x)* of the MMD detector) is below the threshold distance from the permutation test so no drift is flagged. It might be helpful to also add in a small paragraph on permutation testing.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "# Check for drift in covariates\n",
+    "pred = detector.predict(X_test)\n",
+    "print('Is drift? %s!' %labels[pred['data']['is_drift']])"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The test data has drifted into a previously unseen region of feature space, and the model is now misclassifying a number of test observations. If true test labels are avaiable, this is easily detectable by monitoring the test accuracy. However, labels aren't always avaiable during deployment, in which case a drift detector monitoring the covariates comes in handy. In this case, the trained MMD detector successfully detects the covariate drift.\n",
```
arnaudvl, line 491: Typo: avaiable -> available

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "In a similar manner, prior drift can be monitored by training a detector on training labels, and then feeding it a model's predicted labels:\n",
```
arnaudvl, line 493: Being a bit cautious here -> a proxy for prior drift + the usual `y_train` -> `y_ref`.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "- [Learned Kernel](methods/learnedkerneldrift.ipynb)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 3. Online drift detection <a class=\"anchor\" id=\"online\"></a>\n",
+    "\n",
+    "So far, we have discussed drift detection in an *offline* context, with the entire dataset $\\{\\mathbf{X}_i\\}_{i=1}^{N}$ compared to the reference dataset $\\{\\mathbf{X}^{ref}_i\\}_{i=1}^{M}$. However, during model deployment data usually arrives sequentially, and drift detection must be performed in an *online* fashion. \n",
```
arnaudvl, line 532: entire dataset -> maybe clarify to "entire test set"?

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "<div style=\"text-align: center\">\n",
+    "<img src=\"images/online.gif\" width=\"350\"/>\n",
+    "</div>\n",
+    "\n",
+    "One approach is to perform a test for drift every $W$ time-steps, using the $W$ samples that have arrived since the last test. In other words, that is to compare $\\{\\mathbf{X}_i\\}_{i=t}^{t+W}$ to $\\{\\mathbf{X}^{ref}_i\\}_{i=1}^{M}$. Such a strategy could be implemented using any of the offline detectors implemented in alibi-detect, but being both sensitive to slight drift and responsive to severe drift is difficult. If the window size $W$ is too large the delay between test-points hampers responsiveness to severe drift, but an overly small window is unreliable. This is demonstrated below, where the offline [MMD detector](methods/mmddrift.ipynb) is used to monitor drift in data sampled from a normal distribution $\\mathcal{N}\\left(\\mu,\\sigma^2 \\right)$, with the mean starting to drift from $\\mu=0$ to $\\mu=0.5$ at $t=40$.\n",
```
arnaudvl, line 538: "delay between test-points" -> something like "the increased time passed between consecutive statistical tests" since test-points could be confused with points of the test set

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "<img src=\"images/offline_drift_W20.gif\" width=\"700\"/>\n",
+    "</div>\n",
+    "\n",
+    "\n",
+    "### Online drift detectors\n",
+    "\n",
+    "An alternative strategy is to perform a test each time data arrives. However the usual offline methods are not applicable because the process for computing p-values is too expensive, and it doesn’t account for correlated test outcomes when using overlapping windows of test data. Instead, [Alibi Detect](https://github.com/SeldonIO/alibi-detect) offers specialist online drift detectors:\n",
+    "\n",
+    "- [Online Maximum Mean Discrepancy](methods/onlinemmddrift.ipynb)\n",
+    "- [Online Least-Squares Density Difference](methods/onlinemmddrift.ipynb)\n",
```
arnaudvl, line 551: Incorrect link

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "<div style=\"text-align: center\">\n",
+    "<img src=\"images/offline_drift_W2.gif\" width=\"700\"/>\n",
+    "<img src=\"images/offline_drift_W20.gif\" width=\"700\"/>\n",
+    "</div>\n",
+    "\n",
+    "\n",
+    "### Online drift detectors\n",
+    "\n",
+    "An alternative strategy is to perform a test each time data arrives. However the usual offline methods are not applicable because the process for computing p-values is too expensive, and it doesn’t account for correlated test outcomes when using overlapping windows of test data. Instead, [Alibi Detect](https://github.com/SeldonIO/alibi-detect) offers specialist online drift detectors:\n",
```
arnaudvl, line 548: Few things to mention here:
- "it doesn’t account for correlated test outcomes when using overlapping windows of test data" -> add something like "leading to miscalibrated detectors operating at an unknown False Positive Rate (FPR)" -> also make clear that well-calibrated FPR's are key to judge to severity of a drift detection and that in the absence of calibration, drift detection can be useless since there is no way of knowing what fraction are false positives
- Mention that the calibration procedure follows [Cobb et al.](https://arxiv.org/abs/2108.00883)

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "- [Online Maximum Mean Discrepancy](methods/onlinemmddrift.ipynb)\n",
+    "- [Online Least-Squares Density Difference](methods/onlinemmddrift.ipynb)\n",
+    "\n",
+    "The online detectors work by computing the test statistic $S(\\mathbf{x})$ once using the first $W$ data points. Then, during deployment, the test statistic can be updated sequentially at a low cost. When no drift has occured the test statistic fluctuates around its expected value, and once drift occurs the test statistic starts to drift upwards. When it exceeds some preconfigured threshold value, drift is detected. The online detectors are constructed in a similar manner to the offline detectors, for example for the online MMD detector:\n",
```
arnaudvl, line 553: For the initialization of the detector for the first *W* points a slightly different procedure is used I believe which leverages a subset of the reference data (which is then not used by the reference data). Check https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/cd/pytorch/mmd_online.py#L100 and/or the paper itself.

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "```python\n",
+    "online_detector = MMDDriftOnline(X_ref, ert, window_size, backend='tensorflow', n_bootstraps=5000)\n",
+    "```\n",
+    "\n",
+    "But, in addition to providing the detector with reference data, the expected run-time (see below), and size of the sliding window must also be defined. Another important difference is that the online detectors make prediction on single data instances:\n",
```
arnaudvl, line 559: Typo: "make prediction" -> "make predictions"

doc/source/cd/background.ipynb
```diff
@@ -0,0 +1,608 @@
...
+    "\n",
+    "Usually we would like the ERT to be large, however this results in insensitive detectors which are slow to respond when drift does occur. Hence, there is a tradeoff between the expected run time and the expected detection delay. To target the desired ERT, thresholds are configured during an initial configuration phase via simulation (`n_bootstraps` sets the number of boostrap simulations used here). This configuration process is only suitable when the amount of reference data is relatively large (ideally around an order of magnitude larger than the desired ERT). Configuration can be expensive (less so with a GPU) but allows the detector to operate at low-cost during deployment. For a more in-depth explanation, see [Drift Detection: An Introduction with Seldon](https://www.youtube.com/watch?v=JaPMFf0cKlM)."
```
arnaudvl, line 575: Clarify expected detection delay.

ascillitoe:
Still to do:

1. Merge in master. 
2. Look at how to deal with GIFs in latex build.

jklaise:
@ascillitoe what's the main reason for using `.rst` format here? Just conscious about trying to maintaing 3 different kinds of inputs for docs (`.md`, `.ipynb`, `.rst`).

ascillitoe:
> @ascillitoe what's the main reason for using `.rst` format here? Just conscious about trying to maintaing 3 different kinds of inputs for docs (`.md`, `.ipynb`, `.rst`).

Purely for this [image wildcard functionality](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#images). So that we can have sphinx pull in the `.gif`'s for the html build and `.png`'s for the latex build. Not ideal though I know...

ascillitoe:
P.s. opinionated statement here, but I tend to think `.rst` is still the most flexible/powerful out of the three e.g. @arnaudvl and I were discussing adding panels/tabs to better present the algo overview tables at some point, and that functionality does seem to be better supported in `.rst` (e.g. see [https://sphinx-tabs.readthedocs.io/](sphinx-tabs)). Although that being said, [sphinx-panels](https://sphinx-panels.readthedocs.io/) does seem to have some `.md` support. 

I will just do a quick test to see if there is any support for the wildcard functionality in `.md`, as that would solve a lot of our problems. Will also look at converting the other `rst` files I've added to `md`.

jklaise:
@ascillitoe that makes a lot of sense, I agree that for more dynamic views we may need to go with `.rst` anyway.

There is also `myst` which is closer to Markdown but aims to offer a similar level of dynamism, see https://github.com/SeldonIO/alibi/issues/348, although it would be a bigger task to see if we can/want to transition to that.

ascillitoe:
@jklaise ah yes good spot, the `md` support for [sphinx-panels](https://sphinx-panels.readthedocs.io/) was actually `myst` not `md`.

Just did a very quick test/hack to look at removing `m2r` and using `myst` files. The wildcard functionality seems to work nicely, so this could be a good compromise.

jklaise:
Wouldn't `myst `files require a different build process though?

ascillitoe:
For my very simple test simply adding `myst-parser` as a sphinx extension (and using a newer `sphinx` version) seems to be enough, but definitely need to look into it more carefully (and look at readthedocs build process).

doc/source/cd/background.rst
```diff
@@ -0,0 +1,843 @@
+Background
+==========
+
+.. contents:: :local:
+   :depth: 2
+
+0. What is drift?
```
jklaise, line 7: Do we want to start numbering from 0. instead of 1.?

ascillitoe: What is your preference? I was copying `examples/alibi_detect_deploy.html` which starts from 0. but happy to change if 1. if you prefer?

jklaise: I think you need to blame @arnaudvl for that! He has the final word...

arnaudvl: I am impartial, whatever reads the easiest.

ascillitoe:
The latest commit moves the docs to use MyST files. Following https://myst-parser.readthedocs.io/en/latest/sphinx/index.html, all files are named as `md` files, and a passed through the `myst-parser` (vanilla `md` files can still be used). 

This change allows us to remove `m2r`, meaning we are no longer pinned to `sphinx<3.0`. However, I've included `sphinx<=3.2` for now due to an unrelated issue with `autodoc` and the `torch.jit.script` decorator on newer versions. 

The docs builds are now passing, but I need to go through the RTD compiled docs and check them in more detail.

jklaise:
@ascillitoe does `myst-parser` require `sphinx>=3.0`? Just conscious that bumping major sphinx version could unearth a few other issues (though it seems you're on it!).

ascillitoe:
Unfortunately, the latest `myst-parser` version (`0.15.2`) depends on `>=3.1`. For now, I could go back and look at what is the latest `myst-parser` version which works with `sphinx<3.0`, and we could look at moving to a newer `sphinx` version in the future? Although so far it does look like the docs are actually OK with `sphinx>=3.0`?

jklaise:
I see no problem upgrading if the docs are fine, though I would pay special attention to the API docs.

jklaise:
Just a note that if we upgrade, it would be good to do the same changes for `alibi` straight after for consistency and so that we can keep using one local environment for the development of both libraries.

ascillitoe:
Sure thing, I will check for any issues with `alibi` too 👍🏻

Makefile
```diff
@@ -17,16 +17,20 @@ mypy: ## Run typeckecking according to mypy configuration in setup.cfg
...
 	# explicit cd here due to a bug in latexmk 4.41
-	python -m sphinx -b latex -d _build/doctrees -D language=en doc/source doc/_build/latex && \
+	python -m sphinx -b latex -d doc/_build/doctrees -D language=en doc/source doc/_build/latex && \
 	cd doc/_build/latex && \
 	latexmk -pdf -f -dvi- -ps- -jobname=alibi-detect -interaction=nonstopmode
 
+.PHONY: build_latex
+build_latex: ## Build the documentation into a pdf
+	$(MAKE) -C doc latexpdf
+
```
jklaise, line 33: This seem redundant (and overlaps the targets) given the above command.

ascillitoe: Oops must have screwed up that merge resolution. Will delete.

ascillitoe:
P.s. @jklaise as you predicted, the api docs are screwed up with `sphinx==3.2.0` (can't seem to get autodoc to process dostrings in `__init__` methods). Seems to be fixed on even newer `sphinx` versions, but I'm trying to see if I can fix that other issue with `torch.jit.script` here. As a plan B we can always fall back to an older `myst-parser` version.

doc/source/cd/background.md
```diff
@@ -0,0 +1,858 @@
...
+\mathbf{x} = \phi_1 \mathcal{N}\left(\left[-1,-1\right]^T, \sigma^2\mathbf{I} \right) + \phi_2 \mathcal{N}\left(\left[1,1\right]^T, \sigma^2\mathbf{I} \right)
+$$
+
+with the standard deviation set at $\sigma=0.8$, and the weights
+set to $\phi_1=\phi_2=0.5$.
+
+```ipython3
+sigma = 0.8
+ref_norm_0 = stats.multivariate_normal([-1,-1], np.eye(2)*sigma**2)
+ref_norm_1 = stats.multivariate_normal([ 1, 1], np.eye(2)*sigma**2)
```
jklaise, line 422: Should we try to make the code cells mostly self-contained? E.g. `stats` is a submodule of `scipy` (I assume), but it's not imported. I know some other code cells are example-only which are not even valid Python code, but would be great to have as much executable code as possible. (I assume testing this via execution if it was a Jupyter notebook would be ideal, but I assume we want to keep this as a markdown file?).

doc/source/cd/background.rst
ascillitoe: Although starting from 0. is nicely pythonic, I feel like 1.0 might be a bit more readable, so have gone with that. Updated `examples/alibi_detect_deploy.html` too.

doc/source/cd/background.md
ascillitoe: Added import to cell, and linked to stripped down runnable notebook. We should be able to include the notebook in tests too.

doc/source/cd/background.md
```diff
@@ -0,0 +1,862 @@
...
+wide array of methods for detecting drift (see
+[here](../overview/algorithms.md)), some of which are examined in the
+NeurIPS 2019 paper [Failing Loudly: An Empirical Study of Methods for
+Detecting Dataset Shift](https://arxiv.org/abs/1810.11953).
+Generally, these aim to determine whether the distribution
+$P(\mathbf{x})$ has drifted from a reference distribution
+$P_{ref}(\mathbf{x})$, where $\mathbf{x}$ may represent
+input data $\mathbf{X}$, true output data $\mathbf{Y}$, or
+some form of model output, depending on what type of drift we wish to
+detect.
```
ojcobb, line 109: I guess you prefer to keep using x rather than introduce a third variable such as Z but I think this could be a bit confusing give how lower case x usually denotes either constituents of a dataset X or realisations of a rv X.

Could be worth noting that this notation is unconventional, i.e. "...P_ref(x), where we abuse notation slightly to allow x to represent..."

doc/source/cd/background.md
```diff
@@ -0,0 +1,862 @@
...
+
+This can be seen in the animation below, where the online detector
+considers each incoming observation/sample individually, instead of
+considering a batch of observations like the offline detectors.
+
+```{image} images/online_drift.*
+:align: center
+:alt: Online detector
+:width: 700px
+```
```
ojcobb, line 839: Cool animation but I don't understand what n in the bottom left representing? Also I don't know how easy it would be but would be cool if the translucent window followed behind the line to showing how it slides continuously in contrast to the jumping ones above.

doc/source/cd/background.md
```diff
@@ -0,0 +1,862 @@
...
+```{image} images/online.*
+:align: center
+:alt: Online drift detection
+:width: 350px
+```
+
+One approach is to perform a test for drift every $W$ time-steps,
+using the $W$ samples that have arrived since the last test. In
+other words, that is to compare $\{\mathbf{X}_i\}_{i=t}^{t+W}$ to
+$\{\mathbf{X}^{ref}_i\}_{i=1}^{M}$. Such a strategy could be
```
ojcobb, line 762: Nitpick but could we use t-W+1 to t rather than t to t+W given those are the observations available at time t? 

If keeping as is the upper limit should be changed to t+W-1

doc/source/cd/background.md
```diff
@@ -0,0 +1,862 @@
...
+Carlo](http://proceedings.mlr.press/v48/gal16.pdf) can be used to
+provide a notion of uncertainty.
+
+## 4. Online drift detection
+
+So far, we have discussed drift detection in an *offline* context, with
+the entire test set $\{\mathbf{X}_i\}_{i=1}^{N}$ compared to the
+reference dataset $\{\mathbf{X}^{ref}_i\}_{i=1}^{M}$. However, at
+test time data sometimes arrives sequentially, meaning drift detection
+must be performed in an *online* fashion.
```
ojcobb, line 751: Perhaps better motivates online drift detection to say that we would *like* to detect drift in an online fashion in order to respond as quickly as possible and limit the damage it might cause, rather than we *must* because that's how it arrives.

ojcobb (commented):
Amazing work Ashley! The figures and animations are going to be so helpful. Left various comments, some fairly opinionated so obvs feel free to ignore if you don't agree.

ascillitoe:
> Amazing work Ashley! The figures and animations are going to be so helpful. Left various comments, some fairly opinionated so obvs feel free to ignore if you don't agree.

Thanks for these @ojcobb, very helpful, I'll have a look at addressing them where possible :)

doc/source/cd/background.md
```diff
@@ -0,0 +1,862 @@
...
+
+*Drift* is said to occur when the process underlying $\mathbf{X}$
+and $\mathbf{Y}$ at test time differs from the process that
+generated the training data. In this case, we can no longer expect the
+model’s performance on test data to match that observed on held out
+training data. If a *ground truth* is available at test time,
+*supervised drift detection* can be performed, with the model’s
+predictive performance monitored directly. However, in many scenarios,
+such as the binary classification example below, a ground truth is not
+available and *unsupervised drift detection* methods are required.
```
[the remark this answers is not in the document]
ascillitoe: Added clarification here.

doc/source/cd/background.md
```diff
@@ -0,0 +1,862 @@
...
+
+```{image} images/drift_deployment.png
+:align: center
+:alt: Drift in deployment
+```
+
+To explore the different types of drift, consider the common scenario
+where we have input data $\mathbf{X}$ and output data
+$\mathbf{Y}$, jointly distributed according to
+$P(\mathbf{X},\mathbf{Y})$, and a model
```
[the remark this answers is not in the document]
ascillitoe: Done.

doc/source/cd/background.md
```diff
@@ -0,0 +1,862 @@
...
+Note that a change in one of the conditional probabilities
+$P(\mathbf{X}|\mathbf{Y})$ and $P(\mathbf{Y}|\mathbf{X})$
+does not necessarily imply a change in the other. For example, consider
+the pneumonia prediction example from [Lipton et
+al.](https://arxiv.org/pdf/1802.03916.pdf), whereby a classification
+model $f$ is trained to predict $y$, the occurrence (or not)
+of pneumonia, given a list of symptoms $\boldsymbol{x}$. During a
+pneumonia outbreak, $P(\mathbf{Y}|\mathbf{X})$ (e.g. pneumonia
+given cough) might rise, but the manifestations of the disease
+$P(\mathbf{X}|\mathbf{Y})$ might not change.
```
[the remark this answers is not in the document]
ascillitoe: Done.

doc/source/cd/background.md
```diff
@@ -0,0 +1,862 @@
...
+controls for the probability of at least one false positive. The FDR
+correction on the other hand allows for an expected fraction of false
+positives to occur.
+
+Since the univariate tests examine the feature-wise marginal
+distributions, they may miss drift in cases where the joint distribution
+over all $d$ features has changed, but the marginals have not. The
+multivariate tests, which use permutation testing to extract
+$p$-values, offer a more powerful alternative at the cost of
+greater complexity.
```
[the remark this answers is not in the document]
ascillitoe: Thanks, interesting point. Have removed the comment about marginals remaining the same and talked about the above instead.

doc/source/cd/background.md
```diff
@@ -0,0 +1,862 @@
...
+the input space on which the model is uncertain in its predictions. For
+each instance in the reference set the detector obtains the model’s
+prediction and some associated notion of uncertainty. The same is done
+for the test set and if significant differences in uncertainty are
+detected (via a Kolmogorov-Smirnoff test) then drift is flagged. The
+model’s notion of uncertainty depends on the type of model. For a
+classifier this may be the entropy of the predicted label probabilities.
+For a regressor with dropout layers, [dropout Monte
+Carlo](http://proceedings.mlr.press/v48/gal16.pdf) can be used to
+provide a notion of uncertainty.
```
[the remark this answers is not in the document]
ascillitoe: I had similar thoughts. I was a little unsure about it going in dimension reduction as doesn't follow the "dimension reduction as a preprocessing stage" like the other detectors. But I see what you mean about it not fitting here. I've moved it to the dimension reduction section and have added a sentence noting these detectors don't follow quite the same pipeline as the others in that section.

doc/source/cd/background.md
ascillitoe: Also had exactly the same thought as this. I've changed to $\mathbf{z}$ now, and updated figures etc accordingly. Reads OK I think but I need to check again for consistency.

ascillitoe:
@jklaise, @arnaudvl, @ojcobb: I've made another round of changes (to Oliver's comments), and have gone through and checked api etc is OK. Please do let me know if you'd like any more changes. 🙂

jklaise:
LGTM!

doc/source/cd/background.md
```diff
@@ -0,0 +1,869 @@
...
+  $P(\mathbf{Y}|\mathbf{X})$ =
+  $P_{ref}(\mathbf{Y}|\mathbf{X})$. The model may now be
+  extrapolating, potentially resulting in unreliable predictions.
+- **Prior drift**: Also referred to as label drift, this occurs when
+  the distribution of the outputs has shifted
+  $P(\mathbf{Y}) \ne P_{ref}(\mathbf{Y})$, whilst
+  $P(\mathbf{X}|\mathbf{Y})=P_{ref}(\mathbf{X}|\mathbf{Y})$. The
+  model could still be accurate, but accuracy metrics might be
+  affected. Additionally, it is common for the optimal decision 
+  boundary to change under prior drift.
```
ojcobb, line 63: Seems to be three clauses here making the same point. Could we not compress them to something along the lines of "the change could affect the model decision boundary and performance metrics"

doc/source/cd/background.md
```diff
@@ -0,0 +1,869 @@
...
+```{image} images/bg_2d_drift.png
+:align: center
+:alt: 2D drift example
+```
+
+It is relatively easy to spot drift by eyeballing these figures here.
+However, the task becomes considerably harder for high-dimensional real
+problems, especially since real-time ground truths are not typically
+available. Some types of drift, such as prior and concept drift, are
+also difficult to detect without access to ground truths. As a
```
ojcobb, line 97: also -> especially?

doc/source/cd/background.md
```diff
@@ -0,0 +1,869 @@
...
+When applied to multidimensional data with dimension $d$, the
+univariate tests are applied in a feature-wise manner. The obtained
+$p$-values for each feature are aggregated either via the
+[Bonferroni](https://mathworld.wolfram.com/BonferroniCorrection.html)
+or the [False Discovery
+Rate](http://www.math.tau.ac.il/~ybenja/MyPapers/benjamini_hochberg1995.pdf)
+(FDR) correction. The Bonferroni correction is more conservative and
+controls for the probability of at least one false positive. The FDR
+correction on the other hand allows for an expected fraction of false
+positives to occur. The univariate tests assume each test (i.e. each
```
ojcobb, line 182: These corrections preserve the desired false positive rate assuming the tests (...) are independent.

ojcobb (commented):
Three v minor comments but otherwise LGTM!

doc/source/cd/background.md
```diff
@@ -0,0 +1,869 @@
...
+Rate](http://www.math.tau.ac.il/~ybenja/MyPapers/benjamini_hochberg1995.pdf)
+(FDR) correction. The Bonferroni correction is more conservative and
+controls for the probability of at least one false positive. The FDR
+correction on the other hand allows for an expected fraction of false
+positives to occur. The univariate tests assume each test (i.e. each 
+feature dimension in this case) are independent. Usually this is not 
+the case, resulting in false positive rates (FPR's) up to $d$-times 
+lower than desired, which becomes especially problematic when $d$ is 
+large. On the other hand, the multivariate tests allow us to keep our 
+false positive rates known, at the cost of greater complexity.
```
ojcobb, line 187: As @jklaise said, a final comment additionally noting that the univariate tests only consider marginal distributions and could miss changes to joint relationships would be useful.

PR #332, opened 2021-09-14
Fixing division by zero error in tensorflow and pytorch distance files

amirgholipour:
There were some errors related to division by zero in distance files which I fixed them. Also I put some type: ignore comment in cd/utilis.py and cd/tabular.py.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-repos:
+ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.4.0
     hooks:
@@ -7,3 +7,4 @@ repos:
     rev: v0.740
     hooks:
       - id: mypy
+        args: [--no-strict-optional, --ignore-missing-imports]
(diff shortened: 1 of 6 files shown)


amirgholipour:
Please review and accept the changes.

arnaudvl:
Thanks for your interest in the repo and the contribution. There are however a few reasons why we cannot merge it in master. Drift detectors compare distributions of data which assumes here the setting of `n, m > 1` and have little power for just a single sample. If you wish to compare a single instance to a distribution, then we advice to use one of the outlier detectors.

amirgholipour:
> Thanks for your interest in the repo and the contribution. There are however a few reasons why we cannot merge it in master. Drift detectors compare distributions of data which assumes here the setting of `n, m > 1` and have little power for just a single sample. If you wish to compare a single instance to a distribution, then we advice to use one of the outlier detectors.

I tried to address the issue with batch training. after that modification you are able to use Autoencoder based outlier detection with any batch size.

the other issue was division by zero error. which is happening when you would like to use the LearnedKernelDrift with tensorflow and pytorch.
At least, please accept division by zero correction one.

amirgholipour:
Please accept correction with division by zero

PR #338, opened 2021-09-17
Test build to update docs to sphinx 4.x

ascillitoe:
This is a draft PR to update the docs to use `sphinx>=4.0`. There is currently a problem documenting classes that inherit mocked classes (e.g. `alibi_detect.utils.pytorch.kernels.DeepKernel`), which is sensitive to the `sphinx` version:

- `sphinx<3.0`: Everything is OK here. 
- `sphinx>=3.0 < 3.4.2`: Classes that inherit mocked classes are not documented. (see [sphinx #8164](https://github.com/sphinx-doc/sphinx/issues/8164)). This is fixed in `sphinx 3.4.2`. 
- `sphinx>=3.4.2`: The previously missing classes are now documented, but there is a problem with the "Bases" section in the docs. Example: In `alibi_detect.utils.pytorch.kernels.DeepKernel` "Bases: `torch.nn.`" instead of "Bases: `torch.nn.Module`". 

See [here](https://seldon--338.org.readthedocs.build/projects/alibi-detect/en/338/api/alibi_detect.utils.pytorch.html) for an example readthedocs build.


ascillitoe:
Screenshot with `sphinx==4.2`:
sphinx_problem

Screenshot with `sphinx<3.0`:
sphinx_working

ascillitoe:
Update: looks like a fix for this might be on its way in `sphinx 4.3.0`, due approx. early November. Hopefully, we can go ahead and update the docs at that point. See https://github.com/sphinx-doc/sphinx/pull/9658.

ascillitoe:
Looks like `sphinx v4.3.0` might take a little longer than expected to be released. Merging this now so that we can build docs for the new non-parametric detectors (#359). I will leave https://github.com/SeldonIO/alibi/issues/505 open as a reminder for us to check the issue with `Bases` is fixed by `v4.3.0` when it is released.

PR #339, opened 2021-09-17
Add version to metadata

### LINKED ISSUE #236: Add alibi-detect version to metadata

jklaise:
Fixes #236. Takes the same approach as https://github.com/SeldonIO/alibi/pull/481.

### DIFF
--- a/alibi_detect/ad/tests/test_adae.py
+++ b/alibi_detect/ad/tests/test_adae.py
@@ -6,6 +6,7 @@ import tensorflow as tf
 from tensorflow.keras.layers import Dense, InputLayer
 from tensorflow.keras.utils import to_categorical
 from alibi_detect.ad import AdversarialAE
+from alibi_detect.version import __version__
 
 threshold = [None, 5.]
 w_model = [1., .5]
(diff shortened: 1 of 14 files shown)


jklaise:
It appears that due to the current tests of comparing the `meta` attribute directly to a literal dictionary, the literal dict would need to include `"version": __version__` for every test.

It might be better to factor out these `meta` dictionaries and use them when initializing the detectors, though this would still necessitate including `"version": __version__` for each one if we keep the same tests.

jklaise:
Latest commit adds `__version__` entry to the metadata tests. This seemed to be the easiest way to fix the tests without doing any refactoring.

ascillitoe (approved):

ascillitoe:
@jklaise, @arnaudvl:  I've merged this, but a minor follow-up thought, do you think we should move the save/load instructions from "Getting started" to a separate "Saving and loading" page like in alibi? For consistency with alibi but also because I wonder if we could do with adding a warning about different alibi-detect versions like in alibi, and this would start to clutter the "Getting started" page?

jklaise:
Sounds like a good idea.

PR #341, opened 2021-09-21
Install notebook-specific dependencies via shell from the notebooks.

jklaise:
This makes the notebooks more self-contained and enables us to run
notebook tests with extra dependencies. This avoids having to maintain a
dedicated "extras_require" sections for "examples".


jklaise:
This approach may also be lacking in some cases:
 - `pytorch-geometric` is actually a set of libraries, usually the user installs binaries specific to their hardware wheareas just calling `!pip install...` will fetch the sources and trigger a build process which takes time
 - There appear to be some issues with permissions on CI once some data is downloaded onto the home directory

jklaise:
This notebook explicitly references a folder under `/home` which is a relic of how we store and share data internally which is less desirable for user-facing notebooks and for tests:

 - `cd_online_camelyon.ipynb`: `WILDS_PATH = '/home/data/wilds'`

jklaise:
@ascillitoe this is ready for review. Installing the dependencies via shell escape worked for a few notebooks, for others there are issues with downloading large datasets for which I've added some warnings (these will eventually be excluded from notebook tests). We will deal with the `ipywidgets` thing later...

ascillitoe (approved):
@jklaise looks good. I like the idea of also pip installing `alibi` itself, as this should make the notebooks truly self-contained.

One thought with this, do you think we should add a note/warning above where all the examples are listed in the docs (i.e. the new "Examples" page) advising users to run examples within a virtual environment? Just thinking some users might not be happy if they grab one of our now nicely standalone notebooks and it screws with their existing pip installs without them expecting it...

ascillitoe:
Yes agreed one would hope people would be aware! I can add a small warning into the upcoming `feature_docs` PR in any case.

PR #343, opened 2021-09-22
Check for older artifacts without version key in metadata

jklaise:
This fixes the issue of not being able to load older artefacts that don't have a `version` key. Tested on a notebook fetching a pre-trained detector.


ascillitoe (approved):

ascillitoe (changes requested):
flake warning:
```
alibi_detect/utils/saving.py:988:121: E501 line too long (123 > 120 characters)
```

ascillitoe (approved):

PR #344, opened 2021-09-23
Update save utils to use `pathlib` instead of `os` for all path operations

### LINKED ISSUE #335: Replace `os.mkidr` in `utils.saving` so nested directories can be created
Similar to https://github.com/SeldonIO/alibi-detect/issues/318#issuecomment-909192132.

This causes issues in notebooks that train and save a model, effectively `save_detector` is broken if saving to a non-existing subdirectory.

ascillitoe:
Fixes #335. 

`os.path.join` and other `os.path` operations are replaced with their `pathlib` equivalents. `os.path.join` is still used when operating on url's, as `pathlib` complicates matters here.

### DIFF
--- a/alibi_detect/utils/saving.py
+++ b/alibi_detect/utils/saving.py
@@ -78,7 +78,7 @@ DEFAULT_DETECTORS = [
 ]
 
 
-def save_detector(detector: Data, filepath: str) -> None:
+def save_detector(detector: Data, filepath: Union[str, os.PathLike]) -> None:
     """
     Save outlier, drift or adversarial detector.
 
@@ -96,12 +96,14 @@ def save_detector(detector: Data, filepath: str) -> None:
     if detector_name not in DEFAULT_DETECTORS:
         raise ValueError('{} is not supported by `save_detector`.'.format(detector_name))
 
-    if not os.path.isdir(filepath):
+    # check if path exists
+    filepath = Path(filepath)
+    if not filepath.is_dir():
         logger.warning('Directory {} does not exist and is now created.'.format(filepath))
-        os.mkdir(filepath)
+        filepath.mkdir(parents=True, exist_ok=True)
 
     # save metadata
-    with open(os.path.join(filepath, 'meta.dill'), 'wb') as f:
+    with open(filepath.joinpath('meta.dill'), 'wb') as f:
         dill.dump(detector.meta, f)
 
     # save outlier detector specific parameters
@@ -140,7 +142,7 @@ def save_detector(detector: Data, filepath: str) -> None:
     elif detector_name == 'LLR':
         state_dict = state_llr(detector)
 
-    with open(os.path.join(filepath, detector_name + '.dill'), 'wb') as f:
+    with open(filepath.joinpath(detector_name + '.dill'), 'wb') as f:
         dill.dump(state_dict, f)
 
     # save outlier detector specific TensorFlow models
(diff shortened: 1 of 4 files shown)


alibi_detect/utils/saving.py
```diff
@@ -1089,21 +1073,20 @@ def load_tf_model(filepath: str,
...
-        model_dir = os.path.join(filepath, load_dir)
-    if not os.path.isdir(model_dir):
+    model_dir = Path(filepath).joinpath(load_dir)
+    # Check if path exists
+    if not model_dir.is_dir():
         logger.warning('Directory {} does not exist.'.format(model_dir))
         return None
-    if model_name + '.h5' not in [f for f in os.listdir(model_dir) if not f.startswith('.')]:
+    # Check if model exists
+    if model_name + '.h5' not in [f.name for f in model_dir.glob('[!.]*.h5')]:
```
jklaise, line 1082: Not sure this is correct, what are we matching?

alibi_detect/utils/saving.py
```diff
@@ -1120,144 +1103,159 @@ def load_tf_hl(filepath: str, model: tf.keras.Model, state_dict: dict) -> List[t
     -------
     List with loaded tf.keras models.
     """
-    model_dir = os.path.join(filepath, 'model')
+    model_dir = Path(filepath).joinpath('model')
     hidden_layer_kld = state_dict['hidden_layer_kld']
     if not hidden_layer_kld:
-        return None
+        return [None]
```
jklaise, line 1109: Why the change to list return?

alibi_detect/utils/saving.py
```diff
@@ -1120,144 +1103,159 @@ def load_tf_hl(filepath: str, model: tf.keras.Model, state_dict: dict) -> List[t
...
+        Saved model directory.
 
     Returns
     -------
     Loaded AE.
     """
-    model_dir = os.path.join(filepath, 'model')
-    if not [f for f in os.listdir(model_dir) if not f.startswith('.')]:
+    model_dir = Path(filepath).joinpath('model')
+    if not [f.name for f in model_dir.glob('[!.]*.h5')]:
```
jklaise, line 1132: Same comment as above.

alibi_detect/utils/saving.py
ascillitoe: `[f for f in os.listdir(model_dir) if not f.startswith('.')]` returned all files in `model_dir` that don't begin with `.`, and then we checked if `model_name + '.h5` existed within that list. 

`[f.name for f in model_dir.glob('[!.]*.h5')]` should also be returning all files that don't begin with `.`, but with the added restriction of only returning files ending in `.h5`. Idea behind the added restriction was that the `if model_name + '.h5' not in [...]` would be searching through a shorter list. I did test both in a quick colab notebook but guess further tests might be helpful here.

alibi_detect/utils/saving.py
ascillitoe: The output type hint is `-> List[tf.keras.Model]` i.e. `load_tf_hl` returns a list of models. Returning a list of `[None]` gets rid of the type hint warning here. Perhaps `return []` would actually be more suitable here?

alibi_detect/utils/saving.py
ascillitoe: Just putting a quick notebook together to test this further...

jklaise: Oh I was thinking this was regexp but different rules for globbing, so it does what we want!

alibi_detect/utils/saving.py
jklaise: I see, `[]` makes sense to me.

alibi_detect/utils/saving.py
ascillitoe: Ta da!

https://colab.research.google.com/drive/17Iw819bN-mwuorOpYw15x1bIdYa4OIuj?usp=sharing

jklaise (approved):

PR #345, opened 2021-09-23
Notebook testing v2

jklaise:


jklaise:
This should be considered ready for review. The latest GA run on executing all notebooks (apart from excluded ones) is here (still running): https://github.com/SeldonIO/alibi-detect/actions/runs/1270019197. The failures are mostly on notebooks with outdated saved detectors. The changes required are tracked here and can be done in a follow-up PR: https://github.com/SeldonIO/alibi-detect/issues/323#issuecomment-919169191

ascillitoe:
@jklaise perhaps a silly question, but is `pytest` meant to be executing `testing/test_notebooks.py` under the CI github actions? Won't this mean we get a "build failing" tag even if it's only the notebooks failing?

It doesn't look like the `alibi` version (https://github.com/SeldonIO/alibi/pull/484) exhibits this behavior?

jklaise:
Ah good catch, I updated `pytest` to point to `alibi` directory but didn't do likewise for detect.

ascillitoe (approved):

Issue #349, opened 2021-09-27
`fetch_state_dict` only looks for `.pickle` files and not `.dill`


jklaise:
This makes it currently impossible to update artefacts to their `.dill` counterparts and be able to fetch them.

ascillitoe:
Fixed by #350.

Issue #322, opened 2021-09-02
Separate out release vs nightly/PR CI tests


ascillitoe:
A number of the example notebooks have very minor bits of deprecated code. Testing the examples during CI would help pick up these issues. However, some of the notebooks involve expensive computations, so are not suitable for regular CI. To be able to test the examples we probably need to do one (or more) of the following:

- Try to reduce the computational complexity of examples as much as possible. 
- Use a GPU via a github actions runner (Importantly not for PR's https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories)
- Fetch as many pre-trained models as possible (this would also be a good test of the fetching methods). 
- Only test examples during release CI (not nightly and PR's). 

More delineation between nightly and release CI will also be useful in other areas e.g. see https://github.com/SeldonIO/alibi-detect/issues/315#issuecomment-911567056.

Issue #315, opened 2021-08-23
Improve error handling for URL requests in dataset fetch methods


ascillitoe:
The `requests.get()` calls in the fetch methods such as `fetch_nab` and `fetch_cifar10c` currently don't have any error handling, potentially resulting in unclear errors when the requested url is down. For example see:

https://github.com/SeldonIO/alibi-detect/blob/436e99efb88c922dc04cec97b0b52d88bf439d65/alibi_detect/datasets.py#L369-L371

The url requests could be wrapped in try/except blocks which log an exception message and raise an error, in a similar way to the alibi datasets methods:

https://github.com/SeldonIO/alibi/blob/3a589c4de86a89fbdb956427d0d3020de05bdfda/alibi/datasets.py#L105-L110

Logic is also needed in the tests to skip tests when the URL is down, for example:

https://github.com/SeldonIO/alibi/blob/3a589c4de86a89fbdb956427d0d3020de05bdfda/alibi/tests/test_datasets.py#L13-L17

ascillitoe:
Coming back to this. The only issue with skipping the test when a URL is down is that CI won't pick up cases where URL's are permanently dead or changed. Ideally we need to skip temporarily dead URL's during nightly and PR tests, but still check these less frequently i.e. during CI for releases.

ascillitoe:
Reopening this since we missed a utility function in `alibi_detect.datasets`; the `corruption_types_cifar10c` function causes a CI failure if `https://storage.googleapis.com/seldon-datasets` is down (see https://github.com/SeldonIO/alibi-detect/runs/6662544833). We should add the `try-except` logic added in #350 to any tests using this function.

PR #350, opened 2021-09-27
Fetch meta files as dill or pickle

ascillitoe:
Fix for #349.

`fetch_state_dict` looks for `dill` files, and then tries `pickle` files. This can be updated to `dill` only in a future PR, once all google bucket artefacts have been updated to `dill`.

### DIFF
--- a/alibi_detect/tests/test_datasets.py
+++ b/alibi_detect/tests/test_datasets.py
@@ -3,7 +3,7 @@ import pandas as pd
 import pytest
 from requests import RequestException
 from alibi_detect.datasets import fetch_kdd, fetch_ecg, corruption_types_cifar10c, fetch_cifar10c, \
-    fetch_attack, fetch_nab, get_list_nab, fetch_genome
+    fetch_attack, fetch_nab, get_list_nab
 from alibi_detect.utils.data import Bunch
 
 # KDD cup dataset
@@ -82,7 +82,7 @@ attacks = ['cw', 'slide']
 
 
 @pytest.mark.parametrize('return_X_y', [True, False])
-def test_fetch_cifar10c(return_X_y):
+def test_fetch_attack(return_X_y):
     dataset = list(np.random.choice(datasets, 1))[0]
     model = list(np.random.choice(models, 1))[0]
     attack = list(np.random.choice(attacks, 1))[0]
--- a/alibi_detect/tests/test_datasets.py
+++ b/alibi_detect/tests/test_datasets.py
@@ -1,8 +1,107 @@
 import numpy as np
 import pandas as pd
-from alibi_detect.datasets import fetch_nab, get_list_nab
+import pytest
+from requests import RequestException
+from alibi_detect.datasets import fetch_kdd, fetch_ecg, corruption_types_cifar10c, fetch_cifar10c, \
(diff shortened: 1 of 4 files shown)


jklaise:
This also seems to address #315.

ascillitoe:
@jklaise yep! Just need to add in some logic for test time. That being said we're only testing `test_fetch_nab` for now.

alibi_detect/utils/fetching.py
```diff
@@ -404,14 +406,36 @@ def fetch_state_dict(url: str, filepath: Union[str, os.PathLike],
...
+        resp.raise_for_status()
+        suffix = '.dill'
+    except RequestException:
+        try:
+            url_meta = os.path.join(url, 'meta.pickle')
+            resp = requests.get(url_meta, timeout=2)
+            resp.raise_for_status()
+            suffix = '.pickle'
+        except RequestException:
+            logger.exception('Timed out which searching for meta.dill or meta.pickle files at {}.'.format(url))
```
jklaise, line 422: which --> while

I was going to suggest using f-strings, but then I remembered (found?) this: https://blog.pilosus.org/posts/2020/01/24/python-f-strings-in-logging/ Basically it suggests using %-formatting with passing values as arguments to the `exception` method over f-strings or str.format.

alibi_detect/utils/fetching.py
```diff
@@ -404,14 +406,36 @@ def fetch_state_dict(url: str, filepath: Union[str, os.PathLike],
...
+            raise
+
+    # Load metadata and state_dict
+    meta = dill.load(BytesIO(resp.content))
+    try:
+        url_state = os.path.join(url, meta['name'] + suffix)
+        resp = requests.get(url_state)
+        resp.raise_for_status()
+    except RequestException:
+        logger.exception('Timed out which searching for corresponding state file at {}.'.format(url))
```
jklaise, line 432: Same as above

alibi_detect/tests/test_datasets.py
```diff
@@ -1,20 +1,128 @@
...
+    if return_X_y:
+        assert isinstance(data, tuple)
+        assert isinstance(data[0], pd.DataFrame) and isinstance(data[1], pd.DataFrame)
+    else:
+        assert isinstance(data, Bunch)
+        assert isinstance(data.data, pd.DataFrame) and isinstance(data.target, pd.DataFrame)
+
+
+# Genome dataset
+# TODO - Genome dataset is large compared to others - do we want to include in regular CI?
```
jklaise, line 128: Probably not, some notebooks are excluded for the same reason.

If we did we would need to use our own CI image with all the data pre-loaded, but then testing the fetching part is a moot point as it's already available locally.

ascillitoe: Fair point. The current tests only add on ~90 secs (locally) so maybe an OK compromise? We could maybe add slowly ones to non-nightly CI when we address #322?

ascillitoe: I'll remove TODO.

jklaise: I'd keep the TODO as it keeps a record of the discussion in the source.

Issue #348, opened 2021-09-27
Disrepancy between save target for `save_detector` and `fetch_detector`


jklaise:
Currently calling `save_detector` saves directly in the directory specified by `filepath`, however, calling `fetch_detector` saves it in a subdirectory composed of the passed `filepath` and `detector_name`.

saving:
https://github.com/SeldonIO/alibi-detect/blob/e582e6a4b58a6f26001993d7fec4abe1b183e859/alibi_detect/utils/saving.py#L100-L103

fetching:
https://github.com/SeldonIO/alibi-detect/blob/e582e6a4b58a6f26001993d7fec4abe1b183e859/alibi_detect/utils/fetching.py#L449-L452

ascillitoe:
@jklaise how do you feel about adding `detector_name` as an optional kwarg in `save_detector`, and only doing `Path(filepath).joinpath(detector_name)` if `detector_name` is given. Just thinking about minimising changes required in examples etc...

jklaise:
Hmm, maybe that's the least painful option. Basically want to fix what happens in examples now: train model -> `save_detector` -> flip boolean flag -> `fetch_detector` -> new folder created instead of overwriting existing one. This would involve changes to examples though.

ascillitoe:
Ah understood. So we actually need to modify the examples in any case.

PR #351, opened 2021-09-28
Added save_dir argument to save_detector

ascillitoe:
Fix for #348.


alibi_detect/utils/fetching.py
```diff
@@ -470,7 +470,7 @@ def fetch_detector(filepath: Union[str, os.PathLike],
     Initialised pre-trained detector.
     """
     # create path (if needed)
-    filepath = Path(filepath).joinpath(detector_name)
+    filepath = Path(filepath)
```
ascillitoe, line 473: @jklaise perhaps this is the most straightforward fix? There is no confusing logic for the user to contend with. 

We would have to modify all the notebooks though e.g. delete the `filepath = os.path.join(filepath, detector_name)` line in `cd_distillation_cifar10.ipynb`.

.pre-commit-config.yaml
```diff
@@ -4,6 +4,6 @@ repos:
     hooks:
       - id: flake8
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.740
+    rev: v0.812
```
ascillitoe, line 7: This has snuck in as needed this version for all type checks to pass. I have just enabled the pre-commit hooks locally, `mypy==0.812` was already installed locally (In other words, the code has been failing `mypy` checks for some time on `v0.740`, this isn't something new introduced by this PR). 

@jklaise if you are bothered, I can redo this as a separate commit/PR...

alibi_detect/utils/fetching.py
jklaise: This is probably what the user would expect to happen so I agree.

Wrt notebooks, should we modify the saved paths to be `./models/detector_name` so the current behaviour is unchanged (i.e. running multiple notebooks saves all detectors in their own subdirectory so they can be retrieved later).

.pre-commit-config.yaml
jklaise: Great, can you updated the lower bound in https://github.com/SeldonIO/alibi-detect/blob/5927d5b4bd8a4f7209cd42db344145f734074d6b/requirements/dev.txt#L4 please?

alibi_detect/utils/fetching.py
ascillitoe: That's a much nicer solution with the notebooks, I'll do that 👍🏻

ascillitoe: I've had a go at moving the `os.path.join(filepath,detector_name)` outside the `if load_pretrained:` block, so `filepath` should be consistent for fetch and save routes.

PR #352, opened 2021-09-28
Update and include od_vae_adult example

jklaise:
Resolves #364.

I've re-trained the detector and pushed the new version to our Google bucket.


jklaise:
Still an issue with the link at the bottom of https://seldon--352.org.readthedocs.build/projects/alibi-detect/en/352/od/methods/vae.html

PR #353, opened 2021-09-28
Update od_if_kddcup artefacts

jklaise:
Updated the saved artefact to use `scikit-learn==1.0.0`, minor update to the text of the notebook.


jklaise:
Looks like the `scikit-learn==1.0.0` artefact doesn't work with whatever the latest available `scikit-learn` version is for Python 3.6 which is annoying. This shouldn't prevent merging, however I'm conscious that this test will keep failing the build of all notebooks until we deprecate Python 3.6 (should be soon as it is end-of-life in Dec 2021).

arnaudvl (approved):

PR #359, opened 2021-10-07
Features/non parametric drift

ascillitoe:
## What is this PR

This **draft** PR implements four non-parametric drift detectors, intended for use as supervised drift detectors (suitable for detecting concept drift and/or any other types of malicious drift). `CVMDrift` and `CVMDriftOnline` apply the Cramer-von Mises statistical test to continuous data, whilst `FETDrift` and `FETDriftOnline` apply Fisher's exact test to binary data. Both detectors use what are essentially univariate tests, but utilize a multivariate correction to allow for multivariate data. 

## Experiments/Examples
Simple example notebooks available here: 
- Online: https://gist.github.com/ascillitoe/4a0a85007f8a025063e6ef754c631b81.
- Offline: https://gist.github.com/ascillitoe/ce91b9065a0a84df00fddad1b8b95d43.

Benchmarking script testing ART and ADD: 
https://gist.github.com/ascillitoe/0b051e3905bbf7ada9d8a1c421d7420b.

Results are below. Averaged over five initializations (+- is 1std), with ref data of length N=500, `n_bootstraps=50000`, varying number of dimensions d. 

| Detector                        | ERT | ART             | ADD          |
|:------------------------------ |:------:|:----------------:|:--------------:|
| `CVMDriftOnline`, d=1 | 150  | 160.7+-19.3 | 11.7 +- 0.4 |
| `CVMDriftOnline`, d=5 | 150  | 147.3+-13.9 | 3.3 +- 0.3   |
| `FETDriftOnline`, d=1  | 150  | 208.0+-30.3 | 19.4 +- 1.2 |
| `FETDriftOnline`, d=5  | 150  | 160.4+-47.7 | 8.8 +- 2.0  |

@ojcobb could do with your thoughts on whether above look ok. - **update**: After numerous fixes (starting at ab4e4f8), ART is now consistently ~170 (ERT=150) for all d and both alternative hypotheses. We expect ART to be slightly over ERT for the FET detector, so the performance now looks acceptable.

## Outstanding decisions

1. As with other online detectors, the test stats are currently accumulated in `self.test_stats` (until `reset()` called). For the `FETDriftOnline` detector this means storing an array of size `(t-t0,len(self.window_sizes),self.n_features)`. In other words, `n_features` times larger than for the other online detectors. Is this acceptable? If not, we could only store the test stats from the previous time step (needed for exponential smoothing).
2. In the docs "Algorithm Overview" currently "Images" is ticked for FET and CVM detectors. This is technically accurate, however using them for images is perhaps a stretch from their intended use case. Shall we leave as is, or remove to avoid confusion?
3. In "Drift->Detection Background->Online drift detection", the new online detectors are hinted at as following the same strategy as the existing detectors (ref to Cobb et al.), as they also involve threshold config at instantiation. However, there are differences here. Do we need to clarify here?
4. Configuring thresholds for `FETDriftOnline` is rather slow in the multivariate case, as need to do for each window and feature. We could have an option for a truly "non-parametric" approach, where we just simulate `n_window` times with the Bernoulli parameter fixed at 0.5. Docs would need to warn that this is faster but less accurate. @ojcobb what do you think?

## TODO

The code is annotated with a number of outstanding TODO blocks. In addition, the below more general TODO's remain:
- [x] The offline detectors are built on `cramervonmises_2samp` and `fisher_exact` from`scipy.stats`. `cramervonmises_2samp` was only added in `scipy==1.7.0`, but `scipy==1.6.0` dropped support for python 3.6. In this another reason to drop 3.6 support? - _**CVM test case skipped for python 3.6 tests. `UserWarning` added if `CVMDrift` instantiated with old scipy version.**_
- [x] Benchmarking. In particular wall time, memory requirements, wall time and memory usage for different reference data size `N`, number of streams `n_bootstraps`, and window sizes `window_size`. At present, wall times are acceptable, but memory requirements are quite high. **Decide on `device`.**
_@ojcobb note: Re high memory usage, need to check correct usage of binary vs float32 internally._ - _**binary vs float32 was not an issue, high memory usage was with binary array. This is now mitigated by processing streams in batches.**_
- [x] All the detectors still need checking in more detail. In particular, the `FETDriftOnline` detector needs further checks. The ADD values get very high when `p_h1` < `p_h0` in `examples/cd_online_supervised.ipynb`, but not when `p_h1` > `p_h0`. To start, it is worth checking the thresholds vs `t` etc.
_@ojcobb note: this might just be because of the one-sided nature of the custom online detector implementation. Could do with implementing option for two-sided alternative hypothesis for consistency with offline detector._ - _**`'greater'` and `'less'` single-sided tests now implemented. Two-sided test removed from offline.**_
- [x] Further checks to ensure the changes made to `BaseDriftOnline` do not break `MMDDriftOnline` or `LSDDDriftOnline` in any potential use case (all tests pass, and relevant example notebooks have been tested, but further checks would be prudent).
- [x] Methods docs pages.
- [x] Tests for all four detectors. - _**Very basic tests added.**_ 
- [x] Progress bar for `CVMDriftOnline`. This is challenging since `tqdm` cannot be used inside `numba` decorated methods running in `nopython` mode (i.e. the `_ids_to_stats` method). We could use a basic `print` statement called from only the first thread, but this would be ugly since `flush=True` etc can't be used in the `nopython` version of `print`. Using the `numba` ` objmode` context (https://numba.pydata.org/numba-doc/latest/user/withobjmode.html) might be an option, but will likely add a performance penalty. There is also [numba-progress](https://github.com/mortacious/numba-progress), but this is very new and currently doesn't have support for `guvectorize` functions. - _**Moved to "issues" in below comment.**_
- [x] Check behavior of `np.quantile` used in both online detectors. Does this handle tails OK or do we need something else? **- _simple utility function has been added in `utils/misc.py` to replicate behaviour of the tensorflow/pytorch specific `quantile` functions._**
- [x] The `numba` decorators seem to break the `sphinx` docs builds. I haven't found a fix yet, but moving to a newer version of `sphinx` i.e. `>=4.0` seems to fix this. Not ideal, but the easiest option would be simply to wait until we move to the new sphinx version next month... - **_**Moving to `sphinx v4.2.0` now which fixes this.**_**

## Additional issues to consider after initial implementation

(These will be added as issues once the PR is merged)

- [ ] For `FETDriftOnline` a mechanism to determine when the thresholds have stopped decreasing would be useful. 
- [ ] In `utils.misc.quantile` (and the tensorflow/pytorch specific versions) we could consider whether we want to stick with `type=7` as the default. Hyndman and Fan recommend `type=8` (see https://robjhyndman.com/hyndsight/sample-quantiles-20-years-later/).
- [x] Addition of progress bar for `CVMDriftOnline`. 
- [x] All online detectors could do with more detailed CI tests. i.e. check ART and ADD on synthetic data. Need to think more about this as this might introduce non-deterministic tests. 
- [x] Implement `alternative='two-sided'` option for `FET` detectors. We use `hypergeom.cdf` directly to perform the Fisher exact tests, since this can vectorise over multiple data streams (important for `FETDriftOnline`). The two-sided test is more challenging to do in a vectorised fashion due to the need to perform a binary search to find where to begin the halves (see [scipy implementation]. - **This was implemented as a vectorized numba implementation, but performance targeting ERT was poor. Removed in b3a993b80da5dc9d490996aacd6221260b9ae17a**. 
- [ ] Examples for supervised drift detection, and simple example in drift overview page.
- [ ] Refactor `get_thresholds` in all online detectors so that `self.t+1` is given as an argument in the conditional case, and `self.t` in the unconditional (see https://github.com/SeldonIO/alibi-detect/pull/359#discussion_r749252317).
- [ ] Open an issue to update all citations to the new `.bib` format.


alibi_detect/cd/cvm_online.py
```diff
@@ -0,0 +1,268 @@
...
+        t_max = 2 * self.max_ws - 1
+        stats = self._simulate_streams(t_max)
+        # At each t for each stream, find max stats. over window sizes
+        with warnings.catch_warnings():
+            warnings.filterwarnings(action='ignore', message='All-NaN slice encountered')
+            max_stats = np.nanmax(stats, -1)
+        # Now loop through each t and find threshold (at each t) that satisfies eqn. (2) in Ross et al.
+        thresholds = np.zeros((t_max, 1))
+        for t in range(t_max):
+            if t < np.min(self.window_sizes):
```
ojcobb, line 107: Should this be `if t < np.min(self.window_sizes)`?

alibi_detect/cd/cvm_online.py
```diff
@@ -0,0 +1,268 @@
...
+        batches = tqdm(enumerate(tqdm(idxs)), "Computing thresholds over %d batches" % n_batches) if self.verbose \
+            else enumerate(idxs)
+        for b, idx in batches:
+            xs = np.random.randn(len(idx), self.n + t_max)
+            ids = xs[:, None, :] >= xs[:, :, None]
+            stats[idx, :, :] = _ids_to_stats(ids[:, :self.n, :], ids[:, self.n:, :], np.asarray(self.window_sizes))
+
+        # Remove stats prior to windows being full
+        for k, ws in enumerate(self.window_sizes):
+            stats[:ws, k] = np.nan
```
ojcobb, line 140: Should this be `stats[:(ws-1), k] = np.nan`?

ojcobb: In fact doesn't the first axis correspond to the bootstrap id?

ojcobb (commented):
Looking good besides a few annoying off-by-one index errors. Other than that just a few clarifications in docstrings and capturing edge cases :) 

Not yet looked at tests and docs (will do after next round of changes).

alibi_detect/cd/mmd_online.py
```diff
@@ -103,7 +103,7 @@ def thresholds(self):
         return [self._detector.thresholds[min(s, self._detector.window_size-1)] for s in range(self.t)]
 
     def reset(self):
-        "Resets the detector but does not reconfigure thresholds."
+        """Resets the detector but does not reconfigure thresholds."""
```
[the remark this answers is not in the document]
ascillitoe: Not important at all. Officially """ should be used for docstrings so this gets rid of IDE warnings, but I agree its best to leave this out of the current PR to reduce the number of changes.

ascillitoe:
@ojcobb thanks for all your comments! The latest commit attempts to address most of them, except for the one regarding better handling of 1D data, and clarification of ERT handling.

alibi_detect/cd/base_online.py
```diff
@@ -162,3 +163,200 @@ def predict(self, x_t: Union[np.ndarray, Any],  return_test_stat: bool = True,
...
+
+    @abstractmethod
+    def _configure_ref(self):
+        pass
+
+    @abstractmethod
+    def _update_state(self, x_t: np.ndarray):
+        pass
+
+    def _check_x(self, x: Union[Any], x_ref: bool = False) -> np.ndarray:
```
ojcobb, line 270: Think `Union[Any]` is a typo?

alibi_detect/cd/cvm_online.py
```diff
@@ -0,0 +1,267 @@
...
+            n_bootstraps: int = 10000,
+            batch_size: int = 64,
+            n_features: Optional[int] = None,
+            verbose: bool = True,
+            input_shape: Optional[tuple] = None,
+            data_type: Optional[str] = None
+    ) -> None:
+        """
+        Online Cramer-von Mises (CVM) data drift detector using preconfigured thresholds, which tests for
+        any change in the distribution of continuous data.
```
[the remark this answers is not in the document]
ojcobb: Perhaps worth also noting here that detector is an adaptation of that proposed by Ross et al. (2012)

alibi_detect/cd/base_online.py
ascillitoe: Ooops!

alibi_detect/cd/cvm_online.py
```diff
@@ -0,0 +1,267 @@
...
+        # Only need to simulate streams for a single feature here.
+        t_max = 2 * self.max_ws - 1
+        stats = self._simulate_streams(t_max)
+        # At each t for each stream, find max stats. over window sizes
+        with warnings.catch_warnings():
+            warnings.filterwarnings(action='ignore', message='All-NaN slice encountered')
+            max_stats = np.nanmax(stats, -1)
+        # Now loop through each t and find threshold (at each t) that satisfies eqn. (2) in Ross et al.
+        thresholds = np.full((t_max, 1), np.nan)
+        for t in range(np.min(self.window_sizes), t_max):
```
ojcobb, line 108: Should this be `self.window_sizes - 1`?

alibi_detect/cd/cvm_online.py
```diff
@@ -0,0 +1,267 @@
...
+            cumsums[1:, i] = np.cumsum(ids_stream_all[b, :, i])
+
+        for k in range(n_windows):
+            ws = window_sizes[k]
+            win_cdf_ref = (cumsums[ws:, :n] - cumsums[:-ws, :n]) / ws
+            cdf_diffs_on_ref = np.empty_like(win_cdf_ref)
+            for j in range(win_cdf_ref.shape[0]):  # Need to loop through as can't broadcast in njit parallel
+                cdf_diffs_on_ref[j, :] = ref_cdf_all[:n] - win_cdf_ref[j, :]
+            stats[b, (ws-1):, k] = np.sum(cdf_diffs_on_ref * cdf_diffs_on_ref, axis=-1)
+            for t in range(ws, t_max):
```
ojcobb, line 262: Sadly I think there's still out-by-1 errors in this for loop. I think it should start from `ws-1` and we need to add 1 to `t`s indexing `cumsums`.

alibi_detect/cd/cvm_online.py
```diff
@@ -0,0 +1,267 @@
...
+                win_cdf_win = (cumsums[t, n + t - ws:n + t] - cumsums[t - ws, n + t - ws:n + t]) / ws
+                cdf_diffs_on_win = ref_cdf_all[n + t - ws:n + t] - win_cdf_win
+                stats[b, t, k] += np.sum(cdf_diffs_on_win * cdf_diffs_on_win)
```
ojcobb, line 265: Sadly I think there's still out-by-1 errors in this for loop. I think it should start from ws-1 and we need to add 1 to ts indexing cumsums.

alibi_detect/cd/fet.py
```diff
@@ -0,0 +1,285 @@
...
+            preprocess_fn: Optional[Callable] = None,
+            correction: str = 'bonferroni',
+            alternative: str = 'decrease',
+            n_features: Optional[int] = None,
+            input_shape: Optional[tuple] = None,
+            data_type: Optional[str] = None
+    ) -> None:
+        """
+        Fisher exact test (FET) data drift detector, which tests for a change in the mean of binary data.
+        For multivariate data, the Bonferroni or False Discovery Rate (FDR) correction is applied.
```
ojcobb, line 28: I know it's implied but lets be explicit that for multivariate its applied to each feature separately.

alibi_detect/cd/fet_online.py
```diff
@@ -0,0 +1,289 @@
...
+            alternative: str = 'decrease',
+            lam: float = 0.99,
+            n_features: Optional[int] = None,
+            verbose: bool = True,
+            input_shape: Optional[tuple] = None,
+            data_type: Optional[str] = None
+    ) -> None:
+        """
+        Online Fisher exact test (FET) data drift detector using preconfigured thresholds, which tests for a
+        change in the mean of binary data.
```
ojcobb, line 29: Lets say designed for univaiate case and mention multivariate case.

alibi_detect/cd/fet_online.py
```diff
@@ -0,0 +1,289 @@
...
+        thresholds = np.full((self.t_max, self.n_features), np.nan, dtype=np.float32)
+        for f in range(self.n_features):
+            # Compute stats for given feature (for each stream)
+            stats = self._simulate_streams(self.x_ref[:, f], pbar)
+            # At each t for each stream, find max stats. over window sizes
+            with warnings.catch_warnings():
+                warnings.filterwarnings(action='ignore', message='All-NaN slice encountered')
+                max_stats = np.nanmax(stats, -1)
+            # Find threshold (at each t) that satisfies eqn. (2) in Ross et al.
+            for t in range(np.min(self.window_sizes), self.t_max):
```
ojcobb, line 146: Should this be `np.min(self.window_sizes)-1`?

alibi_detect/cd/cvm_online.py
ascillitoe: My bad. Good spot!

alibi_detect/cd/fet_online.py
```diff
@@ -0,0 +1,301 @@
...
+            with warnings.catch_warnings():
+                warnings.filterwarnings(action='ignore', message='All-NaN slice encountered')
+                max_stats = np.nanmax(stats, -1)
+            # Find threshold (at each t) that satisfies eqn. (2) in Ross et al.
+            for t in range(np.min(self.window_sizes)-1, self.t_max):
+                # Compute (1-beta) quantile of max_stats at a given t, over all streams
+                threshold = np.float32(quantile(max_stats[:, t], 1 - beta, interpolate=False, type=6))
+                stats_below = max_stats[max_stats[:, t] < threshold]
+                # Check for stats equal to threshold
+                prob_of_equal = (max_stats[:, t] <= threshold).mean() - (max_stats[:, t] < threshold).mean()
```
ojcobb, line 163: Casting the thresholds to floats whilst keeping the stats as doubles is resulting in always having `prob_of_equal` equal to zero.

ascillitoe: Apologies, I introduced this issue when fixing the issue with stats not occasionally equalling thresholds at predict time (as we'd expect). I'm now init-ing `stats` as `np.float32` in `_simulate_streams`. I've run the debugger to confirm that `prob_of_equal>1` occasionally. I also ran the usual benchmark and got ART=158 (ERT=150).

alibi_detect/cd/fet.py
```diff
@@ -0,0 +1,246 @@
...
+        update_x_ref
+            Reference data can optionally be updated to the last n instances seen by the detector
+            or via reservoir sampling with size n. For the former, the parameter equals {'last': n} while
+            for reservoir sampling {'reservoir_sampling': n} is passed.
+        preprocess_fn
+            Function to preprocess the data before computing the data drift metrics.
+        correction
+            Correction type for multivariate data. Either 'bonferroni' or 'fdr' (False Discovery Rate).
+        alternative
+            Defines the alternative hypothesis. Options are 'less', 'greater' or 'two-sided'.
```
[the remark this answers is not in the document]
ojcobb: Sorry I was really unclear here. I meant to change description to be more like
`Options are 'less', 'greater' or 'two-sided' corresponding to an decrease, increase or change in the mean of the Bernoulli stream.` 
I think we should keep 'less', 'greater' or 'two-sided' to be consistent with scipy. 

On a separate note can we make 'greater' the default as I think detecting increase in the mean of a 0-1 loss stream will be the most common application.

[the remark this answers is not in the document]
ascillitoe: Upon reading your comment again that makes perfect sense, and agree with sticking with scipy notation. I'll change back!

doc/source/cd/methods/cvmdrift.ipynb
```diff
@@ -0,0 +1,137 @@
+{
```
ojcobb, line 1: <p>"emphirical" -&gt; "empirical"</p><p>Whilst the expression for omega^2 is of course correct I think integrals and measures etc are unnecessary in this case and some users would more easily understand the discrete sum version.</p><p> </p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/359/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

doc/source/cd/methods/fetdrift.ipynb
```diff
@@ -0,0 +1,157 @@
+{
```
ojcobb, line 1: <p>I think we should remove the true/false positive and one-hot examples which exhibit complete dependence rather than independence and therefore applying this detector not really a good idea.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/359/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

doc/source/cd/methods/onlinecvmdrift.ipynb
```diff
@@ -0,0 +1,151 @@
+{
```
ojcobb, line 1: <p>Lets mention at top that it's for continuous data.</p><p>"Therefore, it is recommended that users keep&nbsp;M reasonably small" - is a little ambiguous as to what we can handle. I think something more along the lines of "we recommend using fewer than 10,000 instances as reference".</p><p>Maybe note ERT still lower bounded in non-independent multivariate case.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/359/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

doc/source/cd/methods/onlinefetdrift.ipynb
```diff
@@ -0,0 +1,175 @@
+{
```
ojcobb, line 1: <p>"<code>n_bootstraps</code>  number" -&gt;  "<code>n_bootstraps</code>"</p><p>"the windows to be full" -&gt; "a window to be full" </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/359/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

ojcobb, line 1: <p>"starting from&nbsp;<code>t=min(windows_sizes)-1</code>" -&gt; "starting from&nbsp;<code>t=min(windows_sizes)</code>".   I think this is less confusing and the <code>-1</code>  can just be incorporated into the definition of runtime.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/359/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

doc/source/overview/algorithms.md
```diff
@@ -29,6 +29,8 @@ The following tables summarize the advised use cases for the current algorithms.
 |Detector|Tabular|Image|Time Series|Text|Categorical Features|Online|Feature Level|
 |:---|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
 |[Kolmogorov-Smirnov](../cd/methods/ksdrift.ipynb)|✔|✔| |✔|✔| |✔|
+|[Cramér-von Mises](../cd/methods/cvmdrift.ipynb)|✔|✔| |✔|✔|✔|✔|
```
ojcobb, line 32: CVM not applicable to categorical features (including text).
FET not applicable to images and text. Tbh I'm not sure about saying KS and CVM are either but I see that's the existing convention so maybe discuss with @avl.

README.md
```diff
@@ -110,18 +110,20 @@ The following tables show the advised use cases for each algorithm. The column *
...
-| Classifier               | ✔       | ✔     |  ✔          |  ✔    |  ✔                   |       |              |
-| Spot-the-diff            | ✔       | ✔     |  ✔          |  ✔    |  ✔                   |       |  ✔            |
-| Classifier Uncertainty   | ✔       | ✔     |  ✔          |  ✔    |  ✔                   |       |                |
-| Regressor Uncertainty    | ✔       | ✔     |  ✔          |  ✔    |  ✔                   |       |                |
+| Detector                         | Tabular | Image | Time Series | Text  | Categorical Features | Online | Feature Level |
+| :---                             |  :---:  | :---: |   :---:     | :---: |   :---:              | :---:  | :---:         |
+| Kolmogorov-Smirnov               | ✔       | ✔     |             | ✔     | ✔                    |        | ✔             |
+| Cramér-von Mises                 | ✔       | ✔     |             | ✔     | ✔                    | ✔      | ✔             |
+| Fisher's Exact Test              | ✔       | ✔     |             | ✔     | ✔                    | ✔      | ✔             |
```
ojcobb, line 117: See comment above.

doc/source/cd/methods/fetdrift.ipynb
ascillitoe: Ah yes oops, will remove that right away!

ojcobb:
Looking good - we're almost there! As well as the small comments I've made I think it could be worth ensuring the tests cover both the n_features = 1 and n_features > 1 cases and that in the former we can pass reference data and test data as vectors and floats respectively. It seems like this is something that could easily be accidentally broken in the future.

doc/source/overview/algorithms.md
ascillitoe: Yes I had the same thoughts regarding images and text. Good spot re CVM 👍🏻

arnaudvl: Yes it's tricky as there is a big difference to what is possible (just based on data types) and what is recommended. The latter also then strongly depends on the type of preprocessing. So it becomes a rather opinionated table. I would like to redo this table anyway later and separate things out such as online vs. offline / data modalities / insight at the feature and/or instance level.

ascillitoe:
@ojcobb thanks for the suggestion RE tests, that uncovered a slight issue with the existing `check_x` (now fixed). I've added tests for `n_features=1` and `3`, for the online cases. 

I've included the same capability for offline, but only test `n_features=2` for now, since the offline detectors can only currently accept 1D data of the form (N,1) and (1,1). I think we should leave this until a later PR as it will involve modifying `BaseUnivariateDrift` and all the detectors that inherit it.

ascillitoe:
@ojcobb, I added a citation to the online detector docs pages. Let me know if any opposition! 

@jklaise @arnaudvl slightly bad form, but I have snuck in a sphinx extension into this PR, `sphinxcontrib-bibtex`, which allows us to cite references from a central bib file. A new Bibliography page is then populated with nicely formated references. I've checked the html and latex docs builds and it looks nice to me, but if you think this needs more thought I can factor out to a later PR.

doc/source/overview/algorithms.md
ascillitoe: @arnaudvl related to this, we're now on the latest sphinx so we can redo this entire page with panels/tabs etc when we have time...

jklaise:
> @ojcobb, I added a citation to the online detector docs pages. Let me know if any opposition!
> 
> @jklaise @arnaudvl slightly bad form, but I have snuck in a sphinx extension into this PR, `sphinxcontrib-bibtex`, which allows us to cite references from a central bib file. A new Bibliography page is then populated with nicely formated references. I've checked the html and latex docs builds and it looks nice to me, but if you think this needs more thought I can factor out to a later PR.

Sounds good to me. We might need a follow-up PR to go over existing manual citations and consolidate into this new format.

ascillitoe:
@jklaise good idea! I've just a note to the "Additional issues" section on the opening comment, which I'll consolidate down into new issues post merge.

doc/source/cd/methods/cvmdrift.ipynb
```diff
@@ -0,0 +1,136 @@
...
+   "metadata": {},
+   "source": [
+    "# Cramér-von Mises\n",
+    "\n",
+    "## Overview\n",
+    "\n",
+    "The CVM drift detector is a non-parametric drift detector, which applies feature-wise two-sample [Cramér-von Mises](https://en.wikipedia.org/wiki/Cram%C3%A9r%E2%80%93von_Mises_criterion) (CVM) tests. For two empirical distributions $P(z)$ and $P_{ref}(z)$, the CVM test statistic is defined as\n",
+    "\n",
+    "$$\n",
+    "W = \\sum_{z\\in k} \\left| P(z) - P_{ref}(z) \\right|^2,\n",
```
ojcobb, line 21: Nitpick but can we use F instead of P? I think P would usually refer to pdfs rather than cdfs which is a little confusing here.

doc/source/cd/methods/onlinefetdrift.ipynb
```diff
@@ -0,0 +1,175 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Online Fisher's Exact Test\n",
+    "\n",
+    "## Overview\n",
+    "\n",
+    "The online [Fisher's Exact Test](https://en.wikipedia.org/wiki/Fisher%27s_exact_test) detector is a non-parametric method for online drift detection. Like the [offline Fisher's Exact Test](fetdrift.ipynb) detector, it applies an Fisher's Exact Test (FET) to each feature. It is intended for application to [Bernoulli](https://en.wikipedia.org/wiki/Bernoulli_distribution) streams, with binary data consisting of either `(True, False)` or `(0, 1)`. This detector is ideal for use in a supervised setting, monitoring drift in a model's instance level accuracy (i.e. correct prediction = 0, and incorrect prediction = 1). \n",
```
ojcobb, line 18: Can we make a brief comment somewhere saing adapted from Ross's paper.

ojcobb (approved):
Two tiny comments, otherwise LGTM!

doc/source/cd/methods/cvmdrift.ipynb
ascillitoe: Sure thing, will do!

doc/source/cd/methods/onlinefetdrift.ipynb
```diff
@@ -0,0 +1,175 @@
...
+    "\n",
+    "In a similar manner to that proposed by Ross et al. <cite data-cite=\"Ross2012b\"><!-- --></cite>, thresholds are configured by simulating `n_bootstraps` Bernoulli streams. The length of streams can be set with the `t_max` parameter. Since the thresholds are expected to converge after `t_max = 2*max(window_sizes) - 1` time steps, we only need to simulate trajactories and estimate thresholds up to this point, and `t_max` is set to this value by default. Following [[1]](#References), the test statistics are smoothed using an exponential moving average to remove their discreteness, allowing more precise quantiles to be targeted:\n",
```
arnaudvl, line 23: Typo: "trajectories"

PR #371, opened 2021-10-26
Refactor online detectors

ascillitoe:
This PR contains minor(ish?)  refactorings to the online drift detectors. A number of these are intended to be a stepping stone to implementing the non-parametric drift detectors in https://github.com/SeldonIO/alibi-detect/pull/359. 

### Summary of changes
1. `score` and `predict` are refactored to ensure state isn't being updated in multiple places. The state is now updated in `_update_state`, which is called by `score`. This means `score` can now be called without `predict` being called.
2. Preprocessing of `x_t` is moved to `_update_state`, meaning preprocessing is now done for `score` and `predict`. 
3. `score` is exposed as a public method of `MMDDriftOnline` and `LSDDDriftOnline`.
4. `x_t` is converted to `np.ndarray` if given as a `list`.
5. Minor rewording of some docstrings.

### DIFF
--- a/alibi_detect/cd/base_online.py
+++ b/alibi_detect/cd/base_online.py
@@ -90,10 +90,10 @@ class BaseDriftOnline(BaseDetector):
         pass
 
     @abstractmethod
-    def _update_state(self, x_t: Union[np.ndarray, list]):
+    def _update_state(self, x_t: Union[np.ndarray, Any]):
         pass
 
-    def _preprocess_xt(self, x_t: Union[np.ndarray, list]) -> np.ndarray:
+    def _preprocess_xt(self, x_t: Union[np.ndarray, Any]) -> np.ndarray:
         """
         Private method to preprocess a single test instance ready for _update_state.
 
--- a/alibi_detect/cd/base_online.py
+++ b/alibi_detect/cd/base_online.py
@@ -1,7 +1,7 @@
 from abc import abstractmethod
 import logging
 import numpy as np
-from typing import Callable, Dict,  Optional, Union
+from typing import Any, Callable, Dict,  Optional, Union
 from alibi_detect.base import BaseDetector, concept_drift_dict
 from alibi_detect.cd.utils import get_input_shape
 from alibi_detect.utils.frameworks import has_pytorch, has_tensorflow
@@ -125,7 +125,7 @@ class BaseDriftOnline(BaseDetector):
         "Resets the detector but does not reconfigure thresholds."
         self._initialise()
 
-    def predict(self, x_t: Union[np.ndarray, list],  return_test_stat: bool = True,
+    def predict(self, x_t: Union[np.ndarray, Any],  return_test_stat: bool = True,
                 ) -> Dict[Dict[str, str], Dict[str, Union[int, float]]]:
         """
         Predict whether the most recent window of data has drifted from the reference data.
--- a/alibi_detect/cd/base_online.py
+++ b/alibi_detect/cd/base_online.py
@@ -62,10 +62,19 @@ class BaseDriftOnline(BaseDetector):
         self.fpr = 1/ert
         self.window_size = window_size
 
+        # Preprocess reference data
         if isinstance(preprocess_fn, Callable):  # type: ignore
             self.x_ref = preprocess_fn(x_ref)
         else:
             self.x_ref = x_ref
+
+        # Convert reference data if necessary
+        if isinstance(self.x_ref, list):
+            self.x_ref = np.array(self.x_ref)
+        if self.x_ref.ndim == 1:
+            self.x_ref = self.x_ref[:, None]
+
+        # Other attributes
         self.preprocess_fn = preprocess_fn
         self.n = len(x_ref)  # type: ignore
         self.n_bootstraps = n_bootstraps  # nb of samples used to estimate thresholds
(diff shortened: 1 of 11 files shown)


alibi_detect/cd/base_online.py
```diff
@@ -99,7 +125,7 @@ def reset(self) -> None:
         "Resets the detector but does not reconfigure thresholds."
         self._initialise()
 
-    def predict(self, x_t: Union[np.ndarray, Any],  return_test_stat: bool = True,
+    def predict(self, x_t: Union[np.ndarray, list],  return_test_stat: bool = True,
```
arnaudvl, line 128: I don't think that changing form `Any` to `list` as type for `x_t` is correct? We're passing single instances to `predict`, which can be e.g. arrays, string values or custom objects. They are turned into batches (array or list) of size 1 by e.g. `x_t[None, :]` or `[x_t]`.

ascillitoe: Thanks, good point. I shall change this back now!

alibi_detect/cd/base_online.py
```diff
@@ -86,6 +89,29 @@ def _configure_thresholds(self):
     def _configure_ref_subset(self):
         pass
 
+    @abstractmethod
+    def _update_state(self, x_t: Union[np.ndarray, list]):
+        pass
+
+    def _preprocess_xt(self, x_t: Union[np.ndarray, list]) -> np.ndarray:
```
arnaudvl, line 96: @ascillitoe I believe the `list` instead of `Any` type also manifests in other places such as here.

ascillitoe: Sorry yes you're right. The previous behavior was for `predict` to take `Any`, and then predict applied the preprocessing such that `score` was safe to accept `np.ndarray` only. With the new `_preprocess_xt` and `_update_state` all this needs a rethink. I shall look again at this now.

ascillitoe: Ok I think I've updated all the relevant `list`'s to `Any`. This includes `predict` in `cd/mmd_online.py` and `cd/lsdd_online.py`, which were currently `list` in the existing code, but I think should also be `Any`?

alibi_detect/cd/pytorch/mmd_online.py
```diff
@@ -167,30 +167,32 @@ def _configure_thresholds(self):
 
         self.thresholds = thresholds
 
-    def score(self, x_t: np.ndarray) -> Union[float, None]:
+    def _update_state(self, x_t: Union[np.ndarray, Any]):
+        self.t += 1
+        x_t = super()._preprocess_xt(x_t)
```
arnaudvl, line 172: While I agree that this looks pretty clean, just wanted to make sure we are happy that `_update_state` also contains steps which are not related to state, such as preprocessing of `x_t`?

alibi_detect/cd/pytorch/lsdd_online.py
```diff
@@ -178,26 +178,29 @@ def _configure_ref_subset(self):
             h_init = self.c2s - self.k_xtc.mean(0)  # (Eqn 21)
             lsdd_init = h_init[None, :] @ self.H_lam_inv @ h_init[:, None]  # (Eqn 11)
 
-    def score(self, x_t: np.ndarray) -> Union[float, None]:
+    def _update_state(self, x_t: Union[np.ndarray, Any]):
+        self.t += 1
+        x_t = super()._preprocess_xt(x_t)
```
arnaudvl, line 183: Same comment as in the MMD equivalent: would it make sense to do preprocessing of `x_t` separately from `_update_state` and feed the preprocessed `x_t` to `_update_state`?

alibi_detect/cd/tensorflow/mmd_online.py
```diff
@@ -157,25 +157,27 @@ def _configure_thresholds(self):
 
         self.thresholds = thresholds
 
-    def score(self, x_t: np.ndarray) -> Union[float, None]:
+    def _update_state(self, x_t: Union[np.ndarray, Any]):
+        self.t += 1
+        x_t = super()._preprocess_xt(x_t)
```
arnaudvl, line 162: See comment PyTorch version.

alibi_detect/cd/tensorflow/lsdd_online.py
```diff
@@ -164,27 +164,29 @@ def _configure_ref_subset(self):
             h_init = self.c2s - tf.reduce_mean(self.k_xtc, axis=0)  # (Eqn 21)
             lsdd_init = h_init[None, :] @ self.H_lam_inv @ h_init[:, None]  # (Eqn 11)
 
-    def score(self, x_t: np.ndarray) -> Union[float, None]:
+    def _update_state(self, x_t: Union[np.ndarray, Any]):
+        self.t += 1
+        x_t = super()._preprocess_xt(x_t)
```
arnaudvl, line 169: See comment PyTorch version.

alibi_detect/cd/pytorch/lsdd_online.py
ascillitoe: Sure, we can certainly separate these. My thinking behind calling `_preprocess_xt` in `_update_state` was that essentially we are updating state with `x_t`, so it does make some sense to have it inside `_update_state` (as we're only preprocessing `x_t` in order to update the state). But I suppose thinking about more complex handling of state in the future (loading state etc), perhaps it is clearer for us to separate out instead.

ojcobb (approved):
LGTM, nice one! Thanks for catching some of my outdated type hints also.

arnaudvl (approved):

PR #381, opened 2021-11-09
Preprocess poc

ascillitoe:
This PR is a proof of concept for possible changes to the drift detector kwargs for reference data preprocessing. Changes are only made to the `KSDrift` detector. 

See the previous discussion here:
https://github.com/ascillitoe/alibi-detect/pull/2

### Key changes

The `preprocess_x_ref` kwarg is replaced by two kwargs:
1. `x_ref_preprocessed`: Signals whether the provided `x_ref` has already been preprocessed or not. If `True`, then `preprocess_fn` is only applied to the test data.
2. `preprocess_at_init`: This has the same meaning as the old `preprocess_x_ref` i.e. if `True`, `x_ref` is preprocessed at detector instantiation, otherwise preprocessing is performed when `predict` is called. 

Internally the two kwargs (and `preprocess_fn`) are used to set `self.preprocess_at_init` and `self.preprocess_at_pred`, which independently flag whether to perform preprocessing of `x_ref` at init and predict. 

The above kwargs are hopefully more clear to the user, as they draw a clearer distinction between whether or not to preprocess `x_ref`, and when to preprocess `x_ref`. `x_ref_preprocessed` also simplifies ongoing work on config based saving/loading (**Note:** `preprocess_at_init` and `preprocess_at_pred` are used internally, but we will save `x_ref_preprocessed` and `preprocess_at_init` back to config).

### Deprecation warning
If `preprocess_x_ref` is given, a `DeprecationWarning` is raised and internally `preprocess_x_ref` is set to `preprocess_at_init`.

### Tests
This has only been tested on the `cd_ks_cifar10` notebook.


alibi_detect/cd/base.py
```diff
@@ -746,7 +746,8 @@ def __init__(
             self,
             x_ref: Union[np.ndarray, list],
             p_val: float = .05,
-            preprocess_x_ref: bool = True,
+            x_ref_preprocessed: Optional[bool] = False,
+            preprocess_at_init: Optional[bool] = True,
```
jklaise, line 750: Technically `Optional[X]` is equivalent to `Union[X, None]` (https://docs.python.org/3/library/typing.html#typing.Optional), but here `None` is not valid and there is a default non-null value, so I would remove the `Optional` hints.

ascillitoe: oof I realized this (and corrected) for the new detectors PR but missed these. Will remove `Optional` now. Thanks!

alibi_detect/utils/warnings.py
```diff
@@ -0,0 +1,24 @@
+import functools
```
jklaise, line 1: Would be good to add a module docstring as well as docstrings to both functions.

alibi_detect/utils/warnings.py
```diff
@@ -0,0 +1,24 @@
...
+
+def rename_kwargs(func_name, kwargs, aliases):
+    for alias, new in aliases.items():
+        if alias in kwargs:
+            if new in kwargs:
+                raise TypeError('{} received both {} and {}'.format(
+                    func_name, alias, new))
+            warnings.warn('{} is deprecated; use {}'.format(alias, new),
+                          DeprecationWarning,
+                          3)
```
jklaise, line 23: Can you check that the string formatting is performed "early", i.e. there is no earlier output in the stacktrace that contains the non-evaluated string?

jklaise (changes requested):
LGTM, left a few minor comments.

alibi_detect/utils/warnings.py
jklaise: Couldn't resist testing:
```python
In [1]: def f():
   ...:     raise TypeError('{} received'.format(f.__name__))
   ...: 

In [2]: f()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-c43e34e6d405> in <module>
----> 1 f()

<ipython-input-1-acd52b9b6398> in f()
      1 def f():
----> 2     raise TypeError('{} received'.format(f.__name__))
      3 

TypeError: f received
```

So maybe not the worst output. Also it seems that defining the message first doesn't give much better results:

```python
In [3]: def f():
   ...:     msg = f'{f.__name__} received'
   ...:     raise TypeError(msg)
   ...: 
   ...: 

In [4]: f()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-c43e34e6d405> in <module>
----> 1 f()

<ipython-input-3-ce96ff3445ed> in f()
      1 def f():
      2     msg = f'{f.__name__} received'
----> 3     raise TypeError(msg)
      4 
      5 

TypeError: f received
```

ascillitoe: Good to know!

ascillitoe: P.s. I probably should have included an example earlier. This is the result of calling `KSDrift` with `preprocess_x_ref` as a kwarg:

```
<ipython-input-11-12ef60d618f6>:25: DeprecationWarning: preprocess_x_ref is deprecated; use preprocess_at_init
  cd = KSDrift(X_ref, p_val=p_val, preprocess_fn=preprocess_fn, preprocess_x_ref=True)
```

alibi_detect/utils/warnings.py
```diff
@@ -0,0 +1,24 @@
...
+    return deco
+
+
+def rename_kwargs(func_name, kwargs, aliases):
+    for alias, new in aliases.items():
+        if alias in kwargs:
+            if new in kwargs:
+                raise TypeError('{} received both {} and {}'.format(
+                    func_name, alias, new))
+            warnings.warn('{} is deprecated; use {}'.format(alias, new),
```
jklaise, line 21: I think I would also use backticks around `{}` fields to emphasize that these are names of variables, may need to check best practice though.

jklaise: Also, it seems that the error could be either `ValueError` or `TypeError`, not sure what's best... Semantically I think `ValueError` fits better.

ascillitoe: Good idea. Looking at the behaviour of `UnboundLocalError` it seems like this error uses `'{}'`, but then [here](https://www.python.org/dev/peps/pep-0572/#appendix-c-no-changes-to-scope-semantics) backticks are returned. I reckon we go with backticks as we use that to represent variables in docstrings in various places...

arnaudvl (approved):

PR #389, opened 2021-11-24
WIP: Config driven detectors

ascillitoe:
This is a PR to add config-driven save/load functionality via a `config.toml` file. Owing to the complexity of this PR, it will now be implemented as three sequential PR's:

1. https://github.com/SeldonIO/alibi-detect/pull/458: Preparatory changes to detectors.
2. https://github.com/SeldonIO/alibi-detect/pull/462: Pydantic validation of config's.
3. https://github.com/SeldonIO/alibi-detect/pull/469: Main save/load functionality, testing, docs.

## Main features
- Detectors (drift only) can be instantiated from a `config.toml` file by passing it to `load_detector`. 
- The `config.toml` begins as an "unresolved" file (i.e. with artefacts spec'd as a local filepath or function registry via [catalogue](https://github.com/explosion/catalogue)). The initial file is validated with pydantic, before being resolved (i.e. artefacts loaded). The fully resolved config is then validated. `alibi_detect.utils.loading.validate_config` will be made public so that validation of the config dict can be explicitly performed.
- A fully spec'd `config.toml` can be generated from a detector at runtime by passing it to `save_detector`.

## Example

Below is a sample *config.toml* file to spec the MMDDrift detector from the [IMDB example](https://docs.seldon.io/projects/alibi-detect/en/latest/examples/cd_text_imdb.html):

```toml
x_ref = "x_ref.npy"
name = "MMDDrift"

[preprocess_fn]
src = "@cd.tensorflow.preprocess.preprocess_drift"
batch_size = 32
max_len = 100
tokenizer.src = "tokenizer/"

[preprocess_fn.model]
type = "UAE"
src = "model/"

[preprocess_fn.embedding]
src = "embedding/"
type = "hidden_state"
layers = [-1, -2, -3, -4, -5, -6, -7, -8]
```

If the *config.toml* (and its referenced artefacts) are stored in `filepath`, running the following will instantiate this detector:

```python
from alibi_detect.utils.loading import load_detector
cd = load_detector(filepath)
```

For more details, see the new [docs page](https://seldon--389.org.readthedocs.build/projects/alibi-detect/en/389/overview/config_files.html).

## Outstanding tasks 

- [x] CI tests are needed. **Done but...** subject to decision on what (if anything) to add to https://github.com/SeldonIO/alibi-testing.
- [x] Documentation, including a comprehensive description of config specification.
- [x] The `KSDrift` `get_config` method has been updated to the latest config specification. This needs doing for all other detectors. 
- [x] Save/load of learned detectors.
- [x] General tidy-up: In particular decide on more maintainable approach to registering internal alibi-detect functions such as `preprocess_drift`, and decide on an approach for handling backward compatibility/deprecation of current `save_detector`/`load_detector`.
- [x] Clarify use of `custom_obj` in the `model` dict. It is likely this isn't currently usable. 
- [x] Address use of `alibi-testing`. The config tests are implemented using `pytest-cases`, but the cases and fixtures are relatively standalone at the moment so `alibi-testing` is not used (and so perhaps should be removed). Maybe best to incorporate `alibi-testing` in a later PR when we rework all Alibi Detect tests.
- [x] Need to explicitly document which detectors are currently supported.
- [x]  `.get_config` is added for some pytorch specific detectors, but the top-level save/load functionality is not supported (a suitable error is raised). Perhaps should comment out the pytorch `get_config`'s and raise an error in these too for now.

## Out-of-scope
Below items are out-of-scope and will be implemented in a future PR:
- Save/load for online detectors.
- Saving/loading of PyTorch and scikit-learn models (and detectors with these backends).
- Save/load of detector state (e.g. for learned detectors and online detectors).
- Outlier and adversarial detectors.


alibi_detect/cd/base.py
```diff
@@ -740,6 +873,46 @@ def predict(self, x: Union[np.ndarray, list], return_p_val: bool = True, return_
...
+        Get the detector's configuration dictionary.
+
+        Returns
+        -------
+        The detector's configuration dictionary.
+        """
+        cfg: Dict[str, Any] = {
+            'version': __version__,
+            'config_spec': __config_spec__,
+            'name': 'LSDDDrift'
```
jklaise, line 888: For this and some others, what is the reason we're not using `self.__class__.__name__` here? I note that we are using it for e.g. `BaseUnivariateDrift`

ascillitoe: This is done for the detectors with tensorflow/pytorch backends, as we want the name to be `LSDDDrift` instead of `LSDDDriftTF` etc. (We want the non-backend specific wrapper names as this is what we'll use to re-instantiate at load time)

ascillitoe:
image

ascillitoe:
image

ascillitoe:
@mauicv there are very likely still things to iron out, but I think this is probably at a stage where a first pass of it might be useful! 

Unorthodox I know, but I've uploaded some schematics above to give an overview of the general saving and loading workflow.

ascillitoe:
Additional TODO:

Test loading of detectors using a classifiers hidden outputs as a preprocess_fn i.e. 

```python
clf = fetch_tf_model('cifar10', 'resnet32')
preprocess_fn = partial(preprocess_drift, model=HiddenOutput(clf, layer=-1), batch_size=128)
detector = MMDDrift(X_ref, backend='tensorflow', p_val=.05,preprocess_fn=preprocess_fn)
```

doc/source/overview/config_files.md
```diff
@@ -0,0 +1,753 @@
...
+|`kernel.proj`             |           |            |✔       |✔         |
+|`kernel.kernel_a`         |           |✔           |✔       |✔         |
+|`kernel.kernel_b`         |           |✔           |✔       |✔         |
+|`initial_diffs`           |✔          |            |        |          |
+```
+
+```{note}
+When TensorFlow, PyTorch, or scikit-learn `model`'s are specified, the model type should be specified via the `backend` field in the *config.toml*.
+For example, if `model` is a TensorFlow model, set `backend='tensorflow'. This is the case even for detectors that don't take a `backend` kwarg.
+```
```
jklaise, line 174: This note currently doesn't render correctly: 
image

More importantly, I'm rather confused by the terminology, e.g. what is meant by `model` in this note as opposed to `model` in the table above?

doc/source/overview/config_files.md
```diff
@@ -0,0 +1,753 @@
...
+[preprocess_fn]
+src = "@cd.tensorflow.preprocess.preprocess_drift"
+batch_size = 32
+preprocess_batch_fn = "batch_fn.dill"
+```
+
+#### Models 
+
+A `model` dictionary may be used to specify a model within a `preprocess_fn`, or a model for detectors such as the
+[ClassifierDrift](../api/alibi_detect.cd.classifier.rst) detector. The possible fields are:
```
jklaise, line 258: Hmm.. does this mean that a top-level `model` field can be parsed either as part of `preprocess_fn.model` or as a `model` going straight into the detector? What if both are needed? I personally don't like this ambiguity...

doc/source/overview/config_files.md
```diff
@@ -0,0 +1,753 @@
...
+  - Filepath to directory storing the model (relative to the `config.toml` file, or absolute). 
+  -
+* - type
+  - The type of model to be loaded. Options:
+  <br/> - `"custom"`: A generic custom model. 
+  <br/> - `"HiddenOutput"`: An Alibi Detect [HiddenOutput](../api/alibi_detect.cd.tensorflow.rst) model. 
+  <br/> - `"UAE"`: An Alibi Detect [UAE](../api/alibi_detect.cd.tensorflow.rst) model. 
+  - `"custom"`
+* - custom_obj
+  - Dictionary of custom objects. Passed to the tensorflow
```
jklaise, line 279: What are possible examples/use cases of `custom_obj`?

doc/source/overview/config_files.md
```diff
@@ -0,0 +1,753 @@
...
+  - Projection to be applied to the inputs. Should be a TensorFlow model specified as a registered 
+  model, or as an artefact dictionary following the layout in {ref}`model-table`. 
+  - 
+* - `eps`
+  - The proportion (in [0,1]) of weight to assign to `kernel_b`. Specified as a float, or set to `"trainable"`. 
+  - `"trainable"`
+```
+
+As shown in the example below, when a `DeepKernel` is specified, `kernel_a` and `kernel_b` are themselves kernels, 
+and can be specified via the {ref}`kernel-table`.
```
jklaise, line 437: It's not clear to me from the schema that a `DeepKernel` is specified, it seems implicit due to the presence of `kernel_a` and `kernel_b`?

doc/source/overview/config_files.md
```diff
@@ -0,0 +1,753 @@
...
+* - Field
+  - Description
+  - Default (if optional)
+* - `kernel_a`
+  -  Kernel to apply to projected inputs. A string referencing a kernel serialized in a `.dill` file, a registered 
+  kernel object, or a kernel artefact dictionary (see {ref}`kernel-table`).
+  - `"@utils.tensorflow.kernels.GaussianRBF"` <br/>(with `trainable = True`)
+* - `kernel_b`
+  -  Kernel to apply to raw inputs. A string referencing a kernel serialized in a `.dill` file, a registered kernel 
+  object, or a kernel artefact dictionary (see {ref}`kernel-table`).
```
jklaise, line 425: Should these be more descriptive names than `kernel_a` and `kernel_b`?

doc/source/overview/config_files.md
ascillitoe: Yeh the same `ModelConfig` is used to specify a top-level `model` field to be passed in as an argument to a detector such as `ClassifierDrift`, but it can also be used to define a model to be passed into the `preprocess_fn` (but the latter would require `preprocess_fn.model` to be set, not `model`). For example, something like the below should be valid:

```toml
name = "ClassifierDrift"
x_ref = "x_ref.npy"

[model]
type = "custom"
src = "model"

[preprocess_fn]
src = "@cd.tensorflow.preprocess.preprocess_drift"

[preprocess_fn.model]
type = "HiddenOutput"
src = "preprocess_fn/model"
```
The same applies for the `DeepKernel.proj` field, which is expected to be a TF/torch `model`.

The slight ambiguity here comes from the fact that I've tried to keep field names in sync with detector and function arg/kwarg names wherever possible, so that we don't have to maintain a mapping between config fields and detector/function args/kwarg (and a user can go to the relevant api docs for more info on a field). e.g. all the top-level fields in the config (except for `version` and `config_spec`) should match the names of a spec'd detectors args/kwargs. Also a similar story when `preprocess_drift` used for the `preprocess_fn`...

alibi_detect/cd/chisquare.py
```diff
@@ -2,15 +2,18 @@
...
 
 
 class ChiSquareDrift(BaseUnivariateDrift):
+    @deprecated_alias(preprocess_x_ref='preprocess_at_init')
     def __init__(
             self,
             x_ref: Union[np.ndarray, list],
             p_val: float = .05,
             categories_per_feature: Optional[Dict[int, int]] = None,
-            preprocess_x_ref: bool = True,
```
ascillitoe, line 13: This change has been done to draw a distinction between **whether or not** to preprocess x_ref (`x_ref_preprocessed`), and **when** to preprocess x_ref (`preprocess_at_init`). Original PR: https://github.com/SeldonIO/alibi-detect/pull/381

This is necessary since `save_detector` saves the preprocessed `x_ref` (if preprocessing performed at init), and so further preprocessing of `x_ref` should not be done when the detector is loaded again with `load_detector`.

The existing save/load functionality gets around this issue by saving two sets of kwargs within the `state_dict`; a set of kwargs to be used to instantiate the detector, and a set of kwargs to override attributes with post-instantiation. This approach isn't desirable now the config is human readable, hence the changes.

alibi_detect/cd/classifier.py
```diff
@@ -43,6 +44,7 @@ def __init__(
             device: Optional[str] = None,
             dataset: Optional[Callable] = None,
             dataloader: Optional[Callable] = None,
+            input_shape: Optional[tuple] = None,
```
ascillitoe, line 47: `input_shape` added to allow saving of tensorflow models (it was missing from some detectors which aren't currently save-able).

alibi_detect/cd/fet.py
```diff
@@ -108,3 +115,21 @@ def feature_score(self, x_ref: np.ndarray, x: np.ndarray) -> Tuple[np.ndarray, n
...
+
+    def get_config(self) -> dict:
+        """
+        Get the detector's configuration dictionary.
+
+        Returns
+        -------
+        The detector's configuration dictionary.
+        """
+        cfg = super().get_config()
```
ascillitoe, line 127: As much of the config as possible is populated in the parent classes `get_config`. Only the kwargs which are specific to a given detector subclass as populated in the subclasses `get_config` e.g. `alternative` in this case.

alibi_detect/cd/base.py
```diff
@@ -268,6 +292,50 @@ def predict(self, x: Union[np.ndarray, list], return_p_val: bool = True,
...
+        kwargs = {
+            'p_val': self.p_val,
+            'x_ref_preprocessed': self.preprocess_at_init or self.x_ref_preprocessed,
+            'preprocess_at_init': self.preprocess_at_init,
+            'update_x_ref': self.update_x_ref,
+            'preprocess_fn': self.preprocess_fn,
+            'preds_type': self.preds_type,
+            'binarize_preds': self.binarize_preds,
+            'train_size': self.train_size,
+            'n_folds': self.skf.n_splits if self.skf is not None else None,
```
ascillitoe, line 329: Some kwargs are conditional on others, and kwarg's are not always added explicitly as class attributes. Therefore this sort of logic is required in `get_config`'s to infer kwarg's from attributes.

requirements/dev.txt
```diff
@@ -6,11 +6,13 @@ mypy>=0.812, <0.900
 pytest>=5.3.5, <7.0.0
 pytest-cov>=2.6.1, <4.0.0
 pytest-xdist>=1.28.0, <3.0.0 # for distributed testing, currently unused (see setup.cfg)
+pytest_cases>=3.6.8, <4.0.0
 pytest-timeout>=1.4.2, <3.0.0 # for notebook tests
 jupytext>=1.12.0, <2.0.0 # for notebook tests
 ipykernel>=5.1.0, <6.0.0 # for notebook tests
 nbconvert>=6.0.7, <7.0.0 # for notebook tests
 ipywidgets>=7.6.5, <8.0.0 # for notebook tests
+alibi-testing @ git+git://github.com/SeldonIO/alibi-testing@master#egg=alibi-testing # pre-trained models for testing
```
ascillitoe, line 15: This isn't currently used. Removal is pending decision on whether we leave the reworking of `alibi-detect` testing to use `alibi-testing` to a later PR.

requirements/docs.txt
```diff
@@ -5,6 +5,8 @@ sphinx-autodoc-typehints>=1.12.0, <=2.0.0
 sphinx-rtd-theme>=1.0.0, <=2.0.0
 sphinxcontrib-apidoc>=0.3.0, <0.4.0
 sphinxcontrib-bibtex>=2.1.0, <3.0.0
+sphinx-panels>=0.5.0, <1.0.0
```
ascillitoe, line 8: TODO: remove

scripts/gen_configs.py
```diff
@@ -0,0 +1,153 @@
+# TODO pytorch
```
ascillitoe, line 1: This file is part of an effort to automate the generation of configs for use in examples, testing etc. Will likely remove this and save somewhere else for now...

mauicv:
Did you consider a [recursive strategy](https://gist.github.com/mauicv/4fbf29cc57fdc0c0b1b033c033d4832b) for the config serialization and deserialization functions btw?  I think this is how [keras](https://github.com/keras-team/keras/blob/4aabedcb294be97b2321c47e428c16ba3e687d8f/keras/utils/generic_utils.py#L521) approaches it. It's maybe a bit risky though as it's already complex functionality and adding more abstraction perhaps isn't needed.

alibi_detect/cd/base.py
```diff
@@ -46,8 +50,13 @@ def __init__(
...
             p-value used for the significance of the test.
-        preprocess_x_ref
-            Whether to already preprocess and store the reference data.
+        x_ref_preprocessed
+            Whether the given reference data `x_ref` has been preprocessed yet. If `x_ref_preprocessed=True`, only
+            the test data `x` will be preprocessed at prediction time. If `x_ref_preprocessed=False`, the reference
+            data will also be preprocessed.
+        preprocess_at_init
+            Whether to preprocess the reference data when the detector is instantiated. Otherwise, the reference
+            data will be preprocessed at prediction time. Only applies if `x_ref_preprocessed=False`.
```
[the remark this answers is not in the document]
ascillitoe: The `preprocess_at_init` only refers to the reference data `x_ref`, not the test data `x` passed to `.predict()`. So in either case we would still need to pass in `preprocess_fn`, as even if `x_ref` is preprocessed at instantiation, we still need to preprocess `x` at predict time. 

There is an argument for do we want to do any preprocessing at all, or simply expect data to arrive preprocessed. However that is a discussion for another day I think! 

Probs important to note that the above doesn't significantly change the existing functionality. It's just that this behavior was encapsulated by `preprocess_x_ref` before, whereas now the logic is separated out into `preprocess_at_init` and `x_ref_preprocessed`. See this comment https://github.com/SeldonIO/alibi-detect/pull/389#discussion_r817960156.

alibi_detect/cd/base.py
```diff
@@ -86,13 +97,23 @@ def __init__(
...
+        self.preprocess_at_pred = not preprocess_at_init and not x_ref_preprocessed and preprocess_fn is not None
+        self.preprocess_at_init = preprocess_at_init and not x_ref_preprocessed and preprocess_fn is not None
+        self.x_ref_preprocessed = x_ref_preprocessed
+        # Check if preprocess_fn is valid
+        if (self.preprocess_at_init or self.preprocess_at_pred) \
+                and not isinstance(preprocess_fn, Callable):  # type: ignore[arg-type]
+            raise ValueError("`preprocess_fn` is not a valid Callable.")
+
+        # optionally preprocess reference data (now instead of at predict)
+        if self.preprocess_at_init:
```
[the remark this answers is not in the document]
ascillitoe: Yes it pretty much is. That's a nice idea, I'll have a look at doing that. 

I initially wanted to add a specific drift detector base class to avoid duplication, but decided this would be too big a change for the PR. A mix in would be a nice solution I think...

alibi_detect/cd/base.py
```diff
@@ -268,6 +292,50 @@ def predict(self, x: Union[np.ndarray, list], return_p_val: bool = True,
...
+            'config_spec': __config_spec__,
+            'name': 'ClassifierDrift'
+        }
+
+        # x_ref
+        cfg.update({'x_ref': self.x_ref})
+
+        # Preprocess field
+        if self.preprocess_fn is not None:
+            cfg.update({'preprocess_fn': self.preprocess_fn})
```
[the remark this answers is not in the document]
ascillitoe: Again, good idea

alibi_detect/utils/warnings.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def _rename_kwargs(func_name: str, kwargs: Dict[str, Any], aliases: Dict[str, str]):
+    """
+    Private function to rename deprecated kwarg to new name, and raise DeprecationWarning.
+    """
+    for alias, new in aliases.items():
+        if alias in kwargs:
+            if new in kwargs:
+                raise ValueError(f'{func_name} received both `{alias}` and `{new}`')
```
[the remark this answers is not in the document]
ascillitoe: How about the following?

```python
raise ValueError(f"{func_name} received both the deprecated kwarg `{alias}` and it's replacement `{new}`.")
```

scripts/gen_configs.py
```diff
@@ -0,0 +1,153 @@
...
+import os
```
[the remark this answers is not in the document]
ascillitoe: Oops sorry, its old, I'll delete.

doc/source/overview/config_files.md
jklaise: So I'm becoming of the opinion that we should stick as close as possible to the API to avoid confusion. However, this should then be explicitly documented and for every example in the docs there should be a corresponding link to the API docs for the component/function being spec'ed (e.g. when discussing the deep kernel kernel setting, link to the API doc).

alibi_detect/cd/chisquare.py
jklaise: Ah I see now this was merged into the current branch. Wonder if it's worth it splitting out and committing to master?

doc/source/overview/config_files.md
ascillitoe: Makes total sense, I shall add some more links/clarifications in the docs!

alibi_detect/cd/chisquare.py
jklaise: I would say it makes sense and would separate concerns and reduce the size of this PR a bit!

ascillitoe:
> Did you consider a [recursive strategy](https://gist.github.com/mauicv/4fbf29cc57fdc0c0b1b033c033d4832b) for the config serialization and deserialization functions btw? I think this is how [keras](https://github.com/keras-team/keras/blob/4aabedcb294be97b2321c47e428c16ba3e687d8f/keras/utils/generic_utils.py#L521) approaches it. It's maybe a bit risky though as it's already complex functionality and adding more abstraction perhaps isn't needed.

Hey @mauicv just adding a few comments based on our offline discussion regarding this. Main thought is that I generally agree with you here, a recursive strategy would significantly tidy up the save/load logic. However, I think the main challenges here aren't actually to do with how we traverse the nested dict's (i.e. recursion), but actually a number of related issues:

- The line you linked to `serialized_item = serialize_keras_object(item)` sort of hints at the main challenge for us here. The keras serialize/deserialize approach is able to be quite elegant since it is primarily serializing/deserializing keras objects. They can ensure all objects have `serialize`/`deserialize` methods, so that any point in that recursion these methods can be easily called. We already do something similar with the detector `get_config()` methods, however we also need to be able to serialize/deserialize a wide range of objects from different libraries (e.g. our own `preprocess_fn`'s, TensorFlow and pytorch models, HuggingFace tokenizers etc. I feel like we should make `preprocess_drift` a class, which would allow us to add save/load methods to it, but think we'd also have to add thin wrappers for pytorch models etc. Otherwise we'd still need lots of extra logic hidden amongst the recursion. 
- Lots of our artifacts also rely on each other (i.e. we need to chain `embedding`'s and `model`'s together). I think we'd need to use something like the [CustomObjectScope](https://github.com/keras-team/keras/blob/4aabedcb294be97b2321c47e428c16ba3e687d8f/keras/utils/generic_utils.py#L53) and [SharedObjectLoadingScope](https://github.com/keras-team/keras/blob/4aabedcb294be97b2321c47e428c16ba3e687d8f/keras/utils/generic_utils.py#L180). This would be quite a job to think about how this works for something like PyTorch models.

To summarise, IMO it might be worth looking at which of our artefacts can be easily reworked into classes with `get_config`/`from_config` methods, but I'm not sure it will be straightforward to fully adopt a more universal approach like Keras uses....

ascillitoe:
Something just occurred to me. The current logic in this PR is to load a `config.toml` for drift detectors, and use the "legacy" `.dill`  approach for other detectors. However, this introduces a fairly significant breaking change, in that all loading of all old drift detectors will not be supported. Officially this isn't a problem, since we don't recommend loading detectors generated from old versions anyway. However, in practice, a lot of alibi (and Deploy) examples rely on old remote artefacts. Unless we want to align projects to automatically generate artefacts with this PR, I think the safest thing to do is to add back in a legacy fall-back option and raise a DeprecationWarning. Does anyone have any objections to that?

cc @SachinVarghese @RafalSkolasinski @jklaise @arnaudvl @mauicv

alibi_detect/cd/base.py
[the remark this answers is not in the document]
ascillitoe: Done in https://github.com/SeldonIO/alibi-detect/pull/458

alibi_detect/cd/base.py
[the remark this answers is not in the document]
ascillitoe: Done in https://github.com/SeldonIO/alibi-detect/pull/458

alibi_detect/cd/base.py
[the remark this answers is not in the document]
ascillitoe: I've added a mixin for the duplication in the `get_config` methods, based on your comment below. 

However, for the duplication in this case, I think this is highlighting the fact that it might make sense for us to have a base drift detector class. Similar duplication was present before this PR, so personally, I'd be in favor of leaving this out of this PR.

alibi_detect/utils/warnings.py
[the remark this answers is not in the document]
ascillitoe: Done in https://github.com/SeldonIO/alibi-detect/pull/458

alibi_detect/cd/sklearn/classifier.py
```diff
@@ -232,3 +241,13 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray, n
...
         probs_sort = probs_oof[np.argsort(idx_oof)]
         return p_val, dist, probs_sort[:n_ref, 1], probs_sort[n_ref:, 1]
+
+    def get_config(self) -> dict:
+        """
+        Get the detector's configuration dictionary.
+
+        Returns
+        -------
+        The detector's configuration dictionary.
```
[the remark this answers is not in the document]
ascillitoe: Done in #458

alibi_detect/cd/base.py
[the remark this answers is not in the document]
ascillitoe: Discussed in #458

alibi_detect/cd/tensorflow/classifier.py
```diff
@@ -122,7 +138,8 @@ def __init__(
         self.loss_fn = BinaryCrossentropy(from_logits=(self.preds_type == 'logits'))
         self.dataset = partial(dataset, batch_size=batch_size, shuffle=True)
         self.predict_fn = partial(predict_batch, preprocess_fn=preprocess_batch_fn, batch_size=batch_size)
-        self.train_kwargs = {'optimizer': optimizer(learning_rate=learning_rate), 'epochs': epochs,
+        optimizer = optimizer(learning_rate=learning_rate) if isinstance(optimizer, type) else optimizer
```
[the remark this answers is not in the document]
ascillitoe: @mauicv any follow up needed on this one? (just marking things as resolved etc)

Issue #239, opened 2021-05-06
Simplify tensorflow model types


arnaudvl:
`tf.keras.Sequential` is an instance of `tf.keras.Model` so could be left out across types in alibi-detect.

ascillitoe:
Superseded by #668.

PR #393, opened 2021-11-26
changed typings for Sequential

N1m6:
Solve #239


jklaise (approved):
Thanks for your contribution!

PR #395, opened 2021-11-29
Updated error message in RegressorUncertaintyDrift class

tmisirpash:
The block in RegressorUncertaintyDrift which raises the backend ImportError should contain "RegressorUncertaintyDrift" in the error message.


arnaudvl:
Thanks for catching this!

Issue #290, opened 2021-07-13
SpectralResidual results in an odd plot with instance_scores spiking only at the beginning and at the end and flattening at everywhere intermediate.


rajdeep-biswas:
This is from a proprietary dataset with ~550k rows that I can not share here but here are the parameters -
```
od = SpectralResidual(
    threshold=1.,
    window_amp=20,
    window_local=20,
    n_est_points=10,
    n_grad_points=5
)
```

This results in the following (screenshot) -

![image](https://user-images.githubusercontent.com/32306614/125512854-29731a6d-6977-4755-8cd9-a357dbdec570.png)
```
'instance_score': array([0.52549029, 0.74446383, 0.65938995, ..., 0.54188361, 0.67161538,
         0.82524479])
```

Another example -

```
od = SpectralResidual(
    threshold=0.01,
    window_amp=5,
    window_local=5,
    n_est_points=2,
    n_grad_points=5
)
```

This results in -

![image](https://user-images.githubusercontent.com/32306614/125513262-2f793e56-c873-4991-8a11-cad3a5d43b58.png)
```
'instance_score': array([ 2.74739451, -0.1974772 , -0.65613109, ..., -0.04469103,
          0.27587623,  0.91586209])
```

tnpradeep:
Hi,

Even I have observed similar behavior, whatever configuration you provide, always see spikes in beginning and end.
![image](https://user-images.githubusercontent.com/13568762/126621826-8d5e35d5-6987-4b8e-9194-b046abe2dc43.png)

Please explain why ? and how to make it better ?

Thanks
Pradeep

RobertSamoilescu:
Spikes fixes in PR #396

RobertSamoilescu:
@rajdeep-biswas, @tnpradeep, thank you for reporting the issue.
The PR #396 that fixed the issue has been merged into master.

Issue #250, opened 2021-05-14
Set random seeds for all tests


ojcobb:

ascillitoe:
I have renamed this issue to include revisiting randomisation in general across the library. 

Currently, random seeds generally have to be set explicitly to enforce determinism in some detectors. Even with this, determinism isn't guaranteed in some detectors (e.g. see https://github.com/SeldonIO/alibi-detect/issues/390). Some detectors also have a `seed` kwarg, but this only applies to some random operations within the detector.

We should explore reworking this to offer users the option of more easily enforcing reproducibility. Perhaps by adding a `seed` kwarg to all detectors, which would set all relevant seeds internally (without setting any global seeds).

PR #396, opened 2021-11-30
fixed SR begining/ending spikes

RobertSamoilescu:
This PR addresses the spikes at the beginning and ending of the scores reported in #290.
Previous implementation was performing the convolution over the entire spectrum returned by the FFT, thus including the initial bias term. Besides the asymmetry introduced by the convolution operation, which resulted in complex numbers when applying the IFFT, the convolution operation was introducing a considerable bias since the numpy implementation pads the signal with 0 before convolving. The same bias is also valid in time domain. To address the bias, I introduced an option to chose the padding strategy: 
* `constant` - pads the signal with 0
* `replicate` - replicates the most extreme value
* `reflect` - reflects the sginal

After benchmarking the various padding strategy, I concluded that the `reflect` strategy works best, with the `window_amp` centered in the current value (i.e., `side=bilateral`) .

All the issues listed above were significant when the synthetic signal from the example notebook was lifted (i.e., add an offset of 100). 

In addition to the previous fixes, I modified the implementation to performe the averaging in time domain over the previous `window_local` data points (i.e., is a local average of the preceding `window_local` points for the current index) as suggested in the paper.


alibi_detect/od/sr.py
```diff
@@ -5,15 +5,31 @@
...
+PADDING_CONSTANT = 'constant'
+PADDING_REPLICATE = 'replicate'
+PADDING_REFLECT = 'reflect'
+PADDINGS = [PADDING_CONSTANT, PADDING_REFLECT, PADDING_REPLICATE]
+
+# padding sides
+SIDE_BILATERAL = 'bilateral'
+SIDE_LEFT = 'left'
+SIDE_RIGHT = 'right'
+SIDES = [SIDE_BILATERAL, SIDE_RIGHT, SIDE_LEFT]
```
jklaise, line 21: Might be slightly nicer to make these enums, inheriting from both `str` and `enum.Enum`, e.g. `class Padding(str, Enum)` so then can refer to e.g. `Padding.CONSTANT` and you could validate user passed strings against allowed values using one of the methods here: https://stackoverflow.com/questions/29503339/how-to-get-all-values-from-python-enum-class/29503414

ascillitoe:
I haven't looked through in detail yet, but two minor thoughts; given the relative complexity of the new kwargs, it might be worth updating `doc/source/od/methods/sr.ipynb` with more details, and also adding some basic tests for the different padding strategies and padding sides?

alibi_detect/od/sr.py
```diff
@@ -5,15 +5,31 @@
...
 
 class SpectralResidual(BaseDetector, ThresholdMixin):
 
     def __init__(self,
                  threshold: float = None,
                  window_amp: int = None,
                  window_local: int = None,
+                 padding_amp_method: str = PADDING_REFLECT,
+                 padding_local_method: str = PADDING_REFLECT,
+                 padding_amp_side: str = SIDE_BILATERAL,
```
jklaise, line 32: Since this is user facing, I think we should be more explicit, e.g. types as `Literal['constant', 'replicate', 'reflect'] = 'reflect'` which unfortunately introduces some name duplication (unless there's a way to define `Literal` types given a list of string constant from the padding enum?) but is likely worth it for clarity.

alibi_detect/od/sr.py
```diff
@@ -78,6 +123,84 @@ def infer_threshold(self,
         # update threshold
         self.threshold = np.percentile(iscore, threshold_perc)
 
+    @staticmethod
+    def pad_same(X: np.ndarray,
+                 W: np.ndarray,
+                 method: str = "replicate",
+                 side: str = 'bilateral'
```
jklaise, line 130: more explicit typing via `Literal` as above

alibi_detect/od/sr.py
```diff
@@ -1,24 +1,25 @@
+import enum
 import logging
 import numpy as np
-from typing import Dict
+from typing import Dict, Literal
```
jklaise, line 4: Small thinkg I forgot, `Literal` is not in the standard lib for Python 3.6 so for support would need to do something like this:
https://github.com/SeldonIO/alibi/blob/0ccf1b726b6a865b5a94d727847653ffc7e68a4f/alibi/explainers/ale.py#L11-L14

We will drop Python 3.6 support very soon though, though probably not for the next release so best to put this in. @ascillitoe

jklaise: Actually it is only present in standard lib for Python 3.8+ so we won't be able to get rid of the version check for a while as 3.7 will be supported for some time.

alibi_detect/od/tests/test_sr.py
```diff
@@ -7,7 +7,9 @@
 # create normal time series and one with perturbations
 t = np.linspace(0, 0.5, 1000)
 X = np.sin(40 * 2 * np.pi * t) + 0.5 * np.sin(90 * 2 * np.pi * t)
-idx_pert = np.random.randint(0, 1000, 10)
+
+np.random.seed(0)
```
ascillitoe, line 11: @RobertSamoilescu this is probably why the issue with `test_datasets` came up. `np.random.seed(0)` is set here instead of within `def test_sr()`, hence it's probably setting the seed globally for all tests, instead of just this specific test. This is a more general issue which I've flagged in https://github.com/SeldonIO/alibi-detect/issues/250.

ascillitoe:
@RobertSamoilescu I've opened this https://github.com/SeldonIO/alibi-detect/issues/398 to remind us to uncomment `data.meta['attack_type'] == attack` once the dataset is fixed.

alibi_detect/od/sr.py
jklaise: Also, consistent use of `'` over `"` please.

alibi_detect/od/sr.py
```diff
@@ -91,11 +218,36 @@ def saliency_map(self, X: np.ndarray) -> np.ndarray:
         -------
         Array with saliency map values.
         """
+        assert X.shape[0] > self.conv_amp.shape[0], "The length of the input signal should be greater " \
+                                                    "than the amplitude window"
```
jklaise, line 222: This sounds like we should raise an error instead if this is triggered by user error? Or is this something that should never really happen?

RobertSamoilescu: I will raise an error

jklaise (approved):
LGTM. I'll defer to @arnaudvl to have a look at the maths.

alibi_detect/od/sr.py
```diff
@@ -1,19 +1,37 @@
...
+    RIGHT = 'right'
+
+
 class SpectralResidual(BaseDetector, ThresholdMixin):
 
     def __init__(self,
                  threshold: float = None,
                  window_amp: int = None,
                  window_local: int = None,
+                 padding_amp_method: Literal['constant', 'replicate', 'reflect'] = 'reflect',
```
arnaudvl, line 32: Does it make sense to have different padding methods for `amp` and `local`?

RobertSamoilescu: It can definitely be reduced to a single padding method that would be applied in both time and frequency domain. 
It was designed with separate padding methods for benchmarking purposes.  The `'reflect'` padding performed best in my experiments for both domains.

arnaudvl (approved):

jklaise (approved):

ascillitoe (approved):

PR #402, opened 2021-12-06
nbsphinx_link removed

ascillitoe:
This PR removes `nbsphinx_link`, as this was forcing sphinx to rebuild the entire docs each time, instead of performing an incremental build on changed files. 

The example notebooks have been moved to `doc/source/examples`, so that sphinx can access them without `nbsphinx_link`. When a sphinx build is performed, symlinks are generated in `examples/` to point to the notebooks in `doc/source/examples`, so that we still have easily accessible examples closer to the root directory.


jklaise:
I think we need to update the script which picks up changed notebooks for testing. Basically, `pytest` can't handle forward slashes when specifying which tests to run.

testing/test_notebooks.py
```diff
@@ -14,8 +15,8 @@
 # changed notebooks. `pytest` does not allow `/` as part of the test name for the -k argument.
 # This also means that the approach is limited to all notebooks being in the `NOTEBOOK_DIR`
 # top-level path.
-NOTEBOOK_DIR = 'examples'
-ALL_NOTEBOOKS = {Path(x).name for x in glob.glob('examples/*.ipynb')}
+NOTEBOOK_DIR = 'doc/source/examples'
+ALL_NOTEBOOKS = {Path(x).name for x in glob.glob(os.path.join(NOTEBOOK_DIR, '*.ipynb'))}
```
jklaise, line 19: Hmm is this correct?

Also if we do need this probably easiest to stick with `pathlib`?

doc/source/conf.py
```diff
@@ -57,14 +57,23 @@
...
-# Bibliography
+# Create symlinks for example notebooks
+import glob
+nb_files = [os.path.basename(f) for f in glob.glob(os.path.join('examples','*.ipynb')) 
+        if not os.path.basename(f).startswith('temp_')]
+for nb_file in nb_files:
+    target = os.path.join('../../examples', nb_file)
+    if os.path.exists(target):
+        os.remove(target)
+    os.symlink(os.path.join('../doc/source/examples', nb_file), target)
```
jklaise, line 74: Can you explain how this is intended to work? Symlinks are created on every build docs run, but how is this related to the new `.ipynb` files under `examples` which point to the real files under `doc/source/examples`?

testing/test_notebooks.py
ascillitoe: I think so? If I add a `print(ALL_NOTEBOOKS)` and run `pytest -s testing/test_notebooks.py` I get:

```
{'od_aegmm_kddcup.ipynb', 'ad_ae_cifar10.ipynb', 'cd_distillation_cifar10.ipynb', 'cd_mmd_cifar10.ipynb', 'od_seq2seq_ecg.ipynb', 'alibi_detect_deploy.ipynb', 'cd_clf_cifar10.ipynb', 'od_sr_synth.ipynb', 'cd_ks_cifar10.ipynb', 'od_vae_cifar10.ipynb', 'cd_model_unc_cifar10_wine.ipynb', 'od_prophet_weather.ipynb', 'cd_text_imdb.ipynb', 'cd_spot_the_diff_mnist_wine.ipynb', 'od_llr_mnist.ipynb', 'od_mahalanobis_kddcup.ipynb', 'cd_online_wine.ipynb', 'simple_2d_drift.ipynb', 'cd_online_camelyon.ipynb', 'od_if_kddcup.ipynb', 'od_ae_cifar10.ipynb', 'cd_mol.ipynb', 'cd_chi2ks_adult.ipynb', 'od_vae_adult.ipynb', 'od_vae_kddcup.ipynb', 'od_seq2seq_synth.ipynb', 'od_llr_genome.ipynb', 'cd_text_amazon.ipynb'}

Will change to Pathlib in any case!
```

jklaise: Hmm I need to remind myself but the way it's done currently might be intentional due to the issue with `pytest` not supporting forward slashes on the test selection which we do on the `.yml` file.

doc/source/conf.py
ascillitoe: So the bare `.ipynb` files in `examples/` are actually just symlinks pointing to the real ones under `doc/source/examples/`. This enables one to run `jupyter-notebook examples/<>.ipynb` and open the notebook `doc/source/examples/<>.ipynb`.

See below:
image

testing/test_notebooks.py
jklaise: I think this line clarifies it: basically it should just be the notebook name (no directories) because of the `pytest` limitation: https://github.com/SeldonIO/alibi-detect/blob/4299d90fb9daada035abc4449dc5db0a8bd21f1a/.github/workflows/test_changed_notebooks.yml#L55-L56

doc/source/conf.py
jklaise: Ah so those are the same links that were generated running `conf.py` and commited to the repo?

testing/test_notebooks.py
ascillitoe: I'm only appending the directory name in the `glob` call though. `Path(x).name` is still what is stored in the dict i.e. the dict contains only notebook names. I think the behaviour is the same as before? e.g. before we had `glob.glob('examples/*.ipynb')}`?

doc/source/conf.py
ascillitoe: yep! I thought it might be nice to create them via the `conf.py` to reduce the change of us forgetting to add symlinks when we create new notebooks?

testing/test_notebooks.py
ascillitoe: p.s. I think the current error in the notebook tests is actually coming from what you mentioned above. The github workflow is removing `examples` from the path names, but not the new `doc/source`. I'll fix that now and we can check it works!

doc/source/conf.py
jklaise: Sounds good to me, would need to add an instruction to commit the symlinks for new notebooks also.

ascillitoe: @jklaise I've added instructions in the docs readme in the alibi version of this PR: https://github.com/SeldonIO/alibi/pull/554.

testing/test_notebooks.py
ascillitoe: The tests seem to run now, so I think this is resolved.

ascillitoe:
@jklaise I think this is ready for a final review, but we should probably delay merging until after the 0.8.0 release anyway...

jklaise:
@ascillitoe thanks, I'll have another look. I think we haven't removed the requirement from `requirements/dev.txt`.

jklaise:
Any idea why some of the notebooks are failing? They should be filtered against excluded notebooks so theoretically all should pass.

jklaise (approved):
LGTM, nice work on the prolog fix and the symlinks generation!

Just one typo in `simple_2d_drift.ipynb` which I mentioned alread - nice that it was picked up by the notebook tests since this notebook was not being tested previously.

ascillitoe:
Thanks @jklaise! Fixed and `simple_2d_drift.ipynb` is passing now.

PR #403, opened 2021-12-06
Dev/fix latest numpy types

jklaise:
This PR is based on the latest `numpy` which has type information. This flags a few issues with our types which have been resolved in this PR. It is a pre-requisite for upgrading to `tensorflow` 2.7.

Most of this PR is pretty straight-forward, however I've left `TODO: TBD` notes where I would welcome a discussion, this is mostly to do with implicit narrowing of types that either `mypy` can't handle (fixed easily with appropriate `type: ignore[code]`) or more seriously with our logic where we haven't considered that a type might not be narrowed down sufficiently (e.g. is the user `list` always an `np.ndarray` by the time it hits a function that only works on `np.ndarray`?).

There will be another PR enabling strict optional typing which will likely reveal a few more issues with current typing.

### DIFF
--- a/alibi_detect/ad/adversarialae.py
+++ b/alibi_detect/ad/adversarialae.py
@@ -1,16 +1,18 @@
 import logging
+from typing import Callable, Dict, List, Tuple, Union, cast
+
 import numpy as np
 import tensorflow as tf
+from alibi_detect.base import (BaseDetector, FitMixin, ThresholdMixin,
+                               adversarial_correction_dict,
+                               adversarial_prediction_dict)
+from alibi_detect.models.tensorflow.autoencoder import AE
+from alibi_detect.models.tensorflow.losses import loss_adv_ae
+from alibi_detect.models.tensorflow.trainer import trainer
+from alibi_detect.utils.tensorflow.prediction import predict_batch
 from tensorflow.keras.layers import Dense, Flatten
 from tensorflow.keras.losses import kld
 from tensorflow.keras.models import Model
-from typing import Callable, Dict, List, Tuple, Union
-from alibi_detect.models.tensorflow.autoencoder import AE
-from alibi_detect.models.tensorflow.trainer import trainer
-from alibi_detect.models.tensorflow.losses import loss_adv_ae
-from alibi_detect.utils.tensorflow.prediction import predict_batch
-from alibi_detect.base import (BaseDetector, FitMixin, ThresholdMixin,
-                               adversarial_prediction_dict, adversarial_correction_dict)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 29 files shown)


alibi_detect/cd/pytorch/lsdd.py
```diff
@@ -83,16 +83,20 @@ def __init__(
         else:
             self.device = torch.device('cpu')
 
+        # TODO: TBD: the several type:ignore's below are because x_ref is typed as an np.ndarray
```
ascillitoe, line 86: Do you think `--allow-redefinitions` would cause mypy to miss anything significant elsewhere? I wonder if it's best to activate it now as redefining variable types is pretty common for us? 

Otherwise, we could go with a convention of renaming variables instead, i.e. something like
```
x_ref_torch = cast(torch.Tensor, torch.as_tensor(self.x_ref).to(self.device)
```

Although the latter seems undesirable to me...

setup.cfg
```diff
@@ -20,6 +20,7 @@ exclude =
 [mypy]
 ignore_missing_imports = True
 strict_optional = False
+show_error_codes = True
```
ascillitoe, line 23: Good idea!

alibi_detect/od/llr.py
```diff
@@ -245,7 +245,8 @@ def logp(self, dist, X: np.ndarray, return_per_feature: bool = False, batch_size
         Log probabilities.
         """
         logp_fn = partial(dist.log_prob, return_per_feature=return_per_feature)
-        return predict_batch(X, logp_fn, batch_size=batch_size)
+        # TODO: TBD: can this be any of the other types from predict_batch? i.e. tf.Tensor or tuple
```
ascillitoe, line 248: I'm not sure about this one. I'd have guessed that `predict_batch` won't return `tf.Tensor` here since we don't explicitly set `dtype` (so `return_np=True`), but it looks like `tuple` might be possible? Hopefully @arnaudvl has more input!

ascillitoe (approved):
Generally LGTM! Only bit I wonder about is whether to allow type redefinitions now?

alibi_detect/ad/adversarialae.py
```diff
@@ -247,10 +249,12 @@ def score(self, X: np.ndarray, batch_size: int = int(1e10), return_predictions:
         # model predictions
         y = predict_batch(X, self.model, batch_size=batch_size)
         y_recon = predict_batch(X_recon, self.model, batch_size=batch_size)
+        y = cast(np.ndarray, y)  # help mypy out
```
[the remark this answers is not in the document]
jklaise: It doesn't work because `y` is already defined and `--allow-redefinitions` is not enabled.

alibi_detect/cd/pytorch/lsdd.py
jklaise: Renaming wouldn't need the `cast` at all I think.

I'm a bit wary of enabling `--allow-redefinitions` as it's unclear as you say what we could potentially miss. Even so, it is quite limited as redefinitions would be allowed only within the same scope.

alibi_detect/ad/adversarialae.py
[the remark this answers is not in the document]
ascillitoe: Ah that makes sense, thanks!

alibi_detect/cd/pytorch/lsdd.py
ascillitoe: Sorry yes `cast` is making more sense to me now!

I'd have thought that within the same scope would mostly be fine for us? But in any case, it's not clear to me what "within the same block and nesting depth" in the mypy docs means. Also, I take your point that it's probably worth being cautious about changing this.

jklaise: Well I don't know what happened but I just removed all these ignore's and locally the `mypy` check passes now...

jklaise: Nevermind, I was being silly as had downgraded back to `numpy 1.19`...

ascillitoe: Too good to be true!

jklaise:
There are a few more issues with the recently dropped `numpy 1.22` for which I will push another commit shortly.

PR #405, opened 2021-12-09
Supervised example on the penguin dataset

ascillitoe:
This adds a new notebook demonstrating supervised drift detection with the offline FETDrift and CVMDrift detectors, in preparation for the v0.8.0 release.

The 0.8.0 CHANGELOG is also also snuck into this PR.

An example using the online versions will be added in a future PR post-release.


ascillitoe:
> Since importing seaborn would be good to mention that this is needed and can be installed via !pip install seaborn (same as with the dataset below).

Mmmn yeh wasn't sure whether to mention `seaborn` as it is installed via `pip install alibi-detect[examples]`. Do we still want to mention example-wide dependencies in each example?

Will add a note about installing the dataset dependency regardless.

jklaise:
I see what you mean, but I think it's better to be explicit, otherwise the notebook wouldn't run and there's no obvious reason why if we don't comment on it.

PR #406, opened 2021-12-10
fix docs for learnedkernelldrift

kuromt:
This is minor fix.

learnedkernel does not support `n_folds` argument.
https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/cd/learned_kernel.py#L16


jklaise (approved):

jklaise:
Thanks!

Issue #180, opened 2021-02-18
Allow non-tensorflow classifiers in ClassifierDrift


arnaudvl:
Extend method for eg scikit-learn or xgboost classifiers.

arnaudvl:
Partially addressed by #198 .

arnaudvl:
Can be closed if we provide a simple xgboost example.

PR #414, opened 2021-12-29
Sklearn classifier drift extension

RobertSamoilescu:
This PR extends the classifier drift to support `sklearn` models (related to issue #180).

The `sklearn` support is included via the following module: https://github.com/RobertSamoilescu/alibi-detect/tree/classifier_sklearn/alibi_detect/cd/sklearn similar to `pytorch` and `tensorflow`.

It should cover the majority (if not all) `sklearn` models, regardless of the support of `predict_proba` method.

The binomial test should be supported by every model. The KS test can be performed via the score values provided by the classifier that support `decision_function` method and not `predict_proba`. 

Moreover, calibration can be added on top of each model regardless of the support for `decision_function` or `predict_proba`.


alibi_detect/cd/base.py
```diff
@@ -80,9 +80,6 @@ def __init__(
         if isinstance(train_size, float) and isinstance(n_folds, int):
             logger.warning('Both `n_folds` and `train_size` specified. By default `n_folds` is used.')
 
-        if preds_type not in ['probs', 'logits']:
-            raise ValueError("'preds_type' should be 'probs' or 'logits'")
-
```
RobertSamoilescu, line 85: Needs to be moved to each backend specifically:
*  pytorch and tensorflow support probs and logits
* sklearn supports probs and scores

alibi_detect/cd/tensorflow/classifier.py
```diff
@@ -20,7 +20,7 @@ def __init__(
             preprocess_x_ref: bool = True,
             update_x_ref: Optional[Dict[str, int]] = None,
             preprocess_fn: Optional[Callable] = None,
-            preds_type: str = 'preds',
```
RobertSamoilescu, line 23: typo

alibi_detect/cd/classifier.py
```diff
@@ -162,10 +188,12 @@ def predict(self, x: Union[np.ndarray, list],  return_p_val: bool = True,
...
-        'data' contains the drift prediction and optionally the p-value, performance of the classifier
-        relative to its expectation under the no-change null, the out-of-fold classifier model
-        prediction probabilities on the reference and test data, and the trained model.
+        Dictionary containing 'meta' and 'data' dictionaries
+
+         - 'meta' - has the model's metadata.
+
+         - 'data' - contains the drift prediction and optionally the p-value, performance of the classifier \
+        relative to its expectation under the no-change null, the out-of-fold classifier model \
+        prediction probabilities on the reference and test data, and the trained model. \
```
RobertSamoilescu, line 197: To display nicely.

alibi_detect/cd/pytorch/classifier.py
```diff
@@ -123,6 +123,10 @@ def __init__(
             seed=seed,
             data_type=data_type
         )
+
+        if preds_type not in ['probs', 'logits']:
+            raise ValueError("'preds_type' should be 'probs' or 'logits'")
+
```
RobertSamoilescu, line 129: see comment above where deletion was required.

alibi_detect/cd/base.py
jklaise: Should the docstring be updated also?

alibi_detect/cd/classifier.py
```diff
@@ -1,6 +1,10 @@
 import numpy as np
 from typing import Callable, Dict, Optional, Union
-from alibi_detect.utils.frameworks import has_pytorch, has_tensorflow
+from alibi_detect.utils.frameworks import has_pytorch, has_tensorflow, has_sklearn
+
+if has_sklearn:
+    from sklearn.base import ClassifierMixin
+    from alibi_detect.cd.sklearn.classifier import ClassifierDriftSklearn
```
jklaise, line 7: Note that since `scikit-learn` is a core dependency of `detect` this will always be true. That being said it's probably ok to leave this for consistency with `tensorflow` which is also currently a core dependency (but might not be in the future).

alibi_detect/cd/classifier.py
```diff
@@ -16,7 +20,7 @@ class ClassifierDrift:
     def __init__(
             self,
             x_ref: Union[np.ndarray, list],
-            model: Callable,
+            model: Union[ClassifierMixin, Callable],
```
jklaise, line 23: Not really related to this PR, but we should type `model` with the `tensorflow` and `torch` types not general `Callable`. This would require `if TYPE_CHECKING: import tensorflow, torch` at the top of the module.

alibi_detect/cd/classifier.py
```diff
@@ -85,37 +93,48 @@ def __init__(
...
         device
             Device type used. The default None tries to use the GPU and falls back on CPU if needed.
             Can be specified by passing either 'cuda', 'gpu' or 'cpu'. Only relevant for 'pytorch' backend.
         dataset
-            Dataset object used during training.
+            Dataset object used during training. Only relevant for 'tensorflow' and 'pytorch' backends.
         dataloader
             Dataloader object used during training. Only relevant for 'pytorch' backend.
+        use_calibration
+            Whether to use calibration. Whether to use calibration. Calibration can be used on top of any model.
```
jklaise, line 121: Repeated sentence.

alibi_detect/cd/classifier.py
```diff
@@ -85,37 +93,48 @@ def __init__(
...
 
         backend = backend.lower()
-        if backend == 'tensorflow' and not has_tensorflow or backend == 'pytorch' and not has_pytorch:
+        if (backend == 'tensorflow' and not has_tensorflow) or (backend == 'pytorch' and not has_pytorch) or \
+                (backend == 'sklearn' and not has_sklearn):
             raise ImportError(f'{backend} not installed. Cannot initialize and run the '
                               f'ClassifierDrift detector with {backend} backend.')
-        elif backend not in ['tensorflow', 'pytorch']:
+        elif backend not in ['tensorflow', 'pytorch', 'sklearn']:
             raise NotImplementedError(f'{backend} not implemented. Use tensorflow or pytorch instead.')
```
jklaise, line 138: Add `sklearn` to the message.

doc/source/cd/methods/classifierdrift.ipynb
```diff
@@ -45,9 +45,9 @@
     "\n",
     "* `preprocess_fn`: Function to preprocess the data before computing the data drift metrics.\n",
     "\n",
-    "* `preds_type`: Whether the model outputs 'probs' (probabilities) or 'logits'.\n",
+    "* `preds_type`: Whether the model outputs 'probs' (for 'tensorflow', 'pytorch', 'sklearn' models), 'logits' (for 'pytorch', 'tensorflow' models), 'scores' (for 'sklearn' models if `decision_function` is supported).\n",
```
jklaise, line 48: Please add `probabilities` back in the description to be explicit about what `probs` is.

jklaise:
Is the empty `sklearn.preprocess.py` module required? Does it need functionality similar to the `torch/tensorflow` ones or not?

alibi_detect/utils/saving.py
```diff
@@ -89,7 +89,7 @@ def save_detector(detector: Data, filepath: Union[str, os.PathLike]) -> None:
     filepath
         Save directory.
     """
-    if 'backend' in list(detector.meta.keys()) and detector.meta['backend'] == 'pytorch':
+    if 'backend' in list(detector.meta.keys()) and detector.meta['backend'] in ['pytorch', 'sklearn']:
         raise NotImplementedError('Detectors with PyTorch backend are not yet supported.')
```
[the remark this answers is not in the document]
ascillitoe: It would be easy to do, as sklearn models are usually dill friendly. But as it currently stands the config PR will (temporarily) strip any saving support out again. I could also add save/load support for `sklearn` into the config PR, but it would increase the size of the PR...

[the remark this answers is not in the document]
jklaise: Ok let's do it later then.

alibi_detect/cd/sklearn/classifier.py
```diff
@@ -0,0 +1,215 @@
...
+                logger.warning('`retrain_from_scratch=False` sets automatically the parameter `warm_start=True` '
+                               'for the given classifier. Please consult the documentation to ensure that the '
+                               '`warm_start=True` is applicable in the current context (i.e., for tree-based '
+                               'models such as RandomForest, setting `warm_start=True` is not applicable since the '
+                               'fit function expects the same dataset and an update/increase in the number of '
+                               'estimators - previous fitted estimators will be kept frozen while the new ones '
+                               'will be fitted).')
+            else:
+                logger.warning('Current classifier does not support `warm_start`. The model will be retrained '
+                               'from scratch every iteration.')
```
jklaise, line 119: Slightly niche question, but is it the case that if a classifier doesn't support `warm_start` subsequent calls to `fit` are idempotent (i.e. can we expect any classifier state or weights to be always reset to the same initial values)? The answer is probably yes at it appears the API is unified in this way (e.g. [SGDClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html) supports `warm_start` which just takes the existing weights and keeps iterating if `True`).

alibi_detect/cd/sklearn/tests/test_classifier_sklearn.py
```diff
@@ -0,0 +1,119 @@
+import pytest
```
jklaise, line 1: I think it would be great to test `_clone_model` method explicitly and separately with different kinds of estimators and user kwargs as there is quite a lot of logic in there. Would be useful to add in expected errors with `pytest.raises` and also test the outputs of the "custom" `predict_proba` methods.

doc/source/cd/methods/classifierdrift.ipynb
```diff
@@ -30,12 +30,12 @@
     "\n",
```
jklaise, line 30: <p>Aditional --&gt; Additional</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/414/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

doc/source/examples/cd_clf_adult.ipynb
```diff
@@ -0,0 +1,434 @@
+{
```
jklaise, line 1: <p>Can we add a paragraph describing what this notebook is showcasing (i.e. classifier based drift detection using sklearn model and additional calibration). Also worth briefly mentioning how the drifted datasets are constructed later on.</p><p> </p><p>Also add a sentence about <code>alibi</code> needed for the <code>adult</code> dataset.</p><p> </p><p>Finally, this notebook requires a level 1 #Heading and all other headings should be level 2 ## or lower, otherwise the rendering on the docs is off: https://seldon--414.org.readthedocs.build/projects/alibi-detect/en/414/examples/cd_clf_adult.html</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/414/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

jklaise, line 1: <p>Nitpick: I think the first warning is something that we could clean up elsewhere too, i.e. <code>train_size</code>  has a default value but <code>n_folds</code>  doesn't, however, it doesn't feel natural that a warning should be raised because a user provides <code>n_folds</code>  argument just because <code>train_size</code>  has a default.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/414/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

jklaise, line 1: <p>I feel like a paragraph explaining the usefulness of calibration (either here or in the method description notebook, with a link to it from here) would be nice.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/414/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

alibi_detect/cd/base.py
RobertSamoilescu: I don't think so ... why would that be the case?

alibi_detect/cd/classifier.py
RobertSamoilescu: I added for consistency with `tf` as you mentioned.

alibi_detect/cd/base.py
jklaise: The base class can take `probs`, `logits` and `scores` now right?

RobertSamoilescu: Yes, but I moved that check into each backend. So for `tensorflow` & `pytorch` would be `['probs', 'logits']` and for `sklearn` would be `['probs', 'scores']`. For each backend the docs are updated accordingly. The base class now is just storing the `preds_type`.

jklaise: OK but the base class docstring is probably slightly confusing still:
```
        preds_type
            Whether the model outputs probabilities or logits
```

jklaise (approved):
Looks good, I converted from draft so tests are run, there might be a few typing or linter things outstanding.

alibi_detect/cd/sklearn/classifier.py
```diff
@@ -0,0 +1,234 @@
...
+        if preds_type not in ['probs', 'scores']:
+            raise ValueError("'preds_type' should be 'probs' or 'scores'")
+
+        self.meta.update({'backend': 'sklearn'})
+        self.original_model = model
+        self.use_calibration = use_calibration
+        self.calibration_kwargs = dict() if calibration_kwargs is None else calibration_kwargs
+        self.model = self._clone_model()  # type: ClassifierMixin
+
+    def _has_predict_proba(self, model) -> bool:
```
arnaudvl, line 104: Should this function be somewhere in `sklearn.utils` instead? @jklaise

doc/source/examples/cd_clf_adult.ipynb
```diff
@@ -0,0 +1,456 @@
...
+    "\n",
+    "Under the hood, drift detectors leverage a function (also known as a test-statistic) that is expected to take a large value if drift has occurred and a low value if not. The power of the detector is partly determined by how well the function satisfies this property. However, specifying such a function in advance can be very difficult. \n",
+    "\n",
+    "## Detecting drift with a learned classifier\n",
+    "\n",
+    "The classifier-based drift detector simply tries to correctly distinguish instances from the reference data vs. the test set. The classifier is trained to output the probability that a given instance belongs to the test set. If the probabilities it assigns to unseen tests instances are significantly higher (as determined by a Kolmogorov-Smirnoff test) than those it assigns to unseen reference instances then the test set must differ from the reference set and drift is flagged. To leverage all the available reference and test data, stratified cross-validation can be applied and the out-of-fold predictions are used for the significance test. Note that a new classifier is trained for each test set or even each fold within the test set.\n",
```
arnaudvl, line 14: Smirnoff -> Smirnov

doc/source/cd/methods/classifierdrift.ipynb
arnaudvl: Also Smirnoff -> Smirnov

alibi_detect/cd/sklearn/classifier.py
jklaise: If it becomes reused for other things then that would make sense. I was going to suggest factoring this out as a stand-alone function to begin with, but it's a bit inconvenient as it relies on a dataset (ideally, this would be a single-argument function with `model`, but @RobertSamoilescu tried using `inspect` which didn't seem to give the desired results on the latest `sklearn`, so the prediction call is still in place). I would say it's fine right now as it's the only place we use it and better to have it defined in the same module.

arnaudvl: Ok sounds good!

arnaudvl (approved):

Issue #342, opened 2021-09-22
Implement bagged classifier drift detector without data splitting


ojcobb:
If, for the classifier drift detector, we use a bagged model for differentiating reference and test instances we can use the out-of-bag predictions in the permutation test and avoid data splitting. Perhaps sufficiently motivates an implementation of the classifier drift detector where the default classifier is a random forest and no data splitting is performed.

Probably makes sense to implement alongside #180.

ojcobb:
This would obviously be more useful for tabular data rather than text and images.

arnaudvl:
#426

PR #426, opened 2022-01-14
Sklearn Classifier Drift - RF oob

RobertSamoilescu:
This PR implements the `RandomForestClassifier` out-of-bag extension mentioned in #342. 
The extension uses the entire dataset and avoids training `n_folds` classifiers resulting in a significant speed up.


alibi_detect/cd/sklearn/classifier.py
```diff
@@ -232,3 +266,19 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray, n
         p_val, dist = self.test_probs(y_oof, probs_oof, n_ref, n_cur)
         probs_sort = probs_oof[np.argsort(idx_oof)]
         return p_val, dist, probs_sort[:n_ref, 1], probs_sort[n_ref:, 1]
+
+    def _score_rf(self, x: np.ndarray) -> Tuple[float, float, np.ndarray, np.ndarray]:
+        x_ref, x = self.preprocess(x)
+        x, y, _ = self.get_splits(x_ref, x, return_splits=False)
+        self.model.fit(x, y)
+        # it is possible that some inputs do not have OOB scores. This is probably means that too few trees were
```
ascillitoe, line 274: If we are not going to raise an error, I think we should at least raise a warning?

alibi_detect/cd/sklearn/classifier.py
```diff
@@ -134,6 +139,30 @@ def _clone_model(self):
...
+                raise ValueError('OOB supported only for RandomForestClassifier. '
+                                 f'Received a model of type {model.__class__.__name__}')
+
+            if self.use_calibration:
+                self.use_calibration = False
+                logger.warning(
+                    'Calibration cannot be user when `use_oob=True`. Setting `use_calibration=False`.'
+                )
+
+            model.oob_score = True
```
ascillitoe, line 154: This is a nice approach! I was going to implement it with an error raised in this instance, this is much better.

alibi_detect/cd/sklearn/classifier.py
RobertSamoilescu: there is a warning raised by `sklearn` saying that some values are `NaN`.

ascillitoe: Ah yes sorry, the `"Some inputs do not have OOB scores. This probably means too few trees were used to compute any reliable OOB estimates"`. I think that's actually probably descriptive enough for us to just leave it at that?

doc/source/examples/cd_clf_adult.ipynb
```diff
@@ -258,7 +258,8 @@
      "output_type": "stream",
```
jklaise, line 258: <p>I'm curious where does the 63% figure come from? Probably a "citation needed" moment here :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/426/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

ascillitoe: From [Bostrom](https://people.dsv.su.se/~henke/papers/bostrom08b.pdf) quotes this figure: 

_"Since on average about 0.632 of the training examples are used to grow
a single tree in the forest, it means that each example will
be out-of-bag for a little more than a third of the trees in the
forest."_

jklaise: Hmm, is this something that's configurable though? In any case, let's include the citation.

alibi_detect/cd/sklearn/classifier.py
```diff
@@ -134,6 +139,30 @@ def _clone_model(self):
...
+        # oob checks
+        if self.use_oob:
+            if not isinstance(model, RandomForestClassifier):
+                raise ValueError('OOB supported only for RandomForestClassifier. '
+                                 f'Received a model of type {model.__class__.__name__}')
+
+            if self.use_calibration:
+                self.use_calibration = False
+                logger.warning(
+                    'Calibration cannot be user when `use_oob=True`. Setting `use_calibration=False`.'
```
jklaise, line 151: user --> used

doc/source/examples/cd_clf_adult.ipynb
ascillitoe: Don't hold me to this, but I believe it's set by the `max_samples` in scikit-learn. Agree it should probably be clarified and cited here.

alibi_detect/cd/sklearn/classifier.py
```diff
@@ -134,6 +139,30 @@ def _clone_model(self):
...
+                )
+
+            model.oob_score = True
+            model.bootstrap = True
+            logger.warning(
+                '`use_oob=True` sets automatically the parameter `boostrap=True` and `oob_score=True`. '
+                '`train_size` and `n_folds` are ignored when `use_oob=True`.'
```
jklaise, line 158: Would be a tiny bit more explicit, i.e. "parameters" --> "classifier parameters" as it wasn't clear without reading the code that these are not our parameters.

alibi_detect/cd/sklearn/classifier.py
```diff
@@ -134,6 +139,30 @@ def _clone_model(self):
...
+            )
+        else:
+            if isinstance(model, RandomForestClassifier):
+                model.oob_score = False
+                logger.warning('`use_oob=False` sets automatically the parameter `oob_score=False`.')
```
jklaise, line 163: Same comment as above

alibi_detect/cd/sklearn/tests/test_classifier_sklearn.py
```diff
@@ -270,3 +270,90 @@ def test_predict_proba2(model, pred_types, use_calibration, binarize_preds):
...
+    # this should pass because `use_oob=True` and the model is RF
+    cd = ClassifierDriftSklearn(x_ref=np.random.randn(100, 5),
+                                model=model,
+                                preds_type=preds_type,
+                                use_calibration=use_calibration,
+                                use_oob=use_oob)
+
+    assert cd.model.oob_score
+    assert cd.model.bootstrap
+    assert not cd.use_calibration   # shuld be set to `False` when `use_oob=True`
```
jklaise, line 347: shuld --> should

alibi_detect/cd/sklearn/tests/test_classifier_sklearn.py
```diff
@@ -270,3 +270,90 @@ def test_predict_proba2(model, pred_types, use_calibration, binarize_preds):
...
+
+
+@pytest.mark.parametrize('model', [RandomForestClassifier(oob_score=True),
+                                   RandomForestClassifier(oob_score=False)])
+@pytest.mark.parametrize('preds_type', ['probs'])
+@pytest.mark.parametrize('use_oob', [False])
+def test_clone_rf3(model, preds_type, use_oob):
```
jklaise, line 354: please clarify the aim of this test (a comment like in the above tests would do it)

jklaise (changes requested):
Overall LGTM, just small comments.

doc/source/examples/cd_clf_adult.ipynb
RobertSamoilescu: This is a consequence of the bootstrapping. Each tree is trained on a dataset obtained by sampling with replacement from the original dataset. Let us consider that the original dataset has `N ` instances. When sampling with replacement, a training instance has a probability of being missed equal to `1-1/N` each time we draw a sample from the original dataset. Since we sample `N` times, the probability of an instance to be missed (i.e., never selected) from the bootstrap dataset is equal to `(1 - 1/N)^N` which converges to `1/e` as `N` goes to infinity. Thus, the expected number of missed samples is equal to `N/e`, which implies that the expected number of unique samples is `N - N/e = N(1 - 1/e)`. Now, `1 - 1/e` is approximately 0.63.

jklaise: I think reference to the paper is enough.

jklaise (approved):

alibi_detect/cd/sklearn/classifier.py
arnaudvl: Agreed

alibi_detect/cd/base.py
```diff
@@ -154,6 +159,9 @@ def get_splits(self, x_ref: Union[np.ndarray, list], x: Union[np.ndarray, list])
         else:  # add 2 lists
             x = x_ref + x
 
+        if not return_splits:
```
arnaudvl, line 162: If we're not going to return splits, can't we just return `x, y`?

alibi_detect/cd/sklearn/classifier.py
```diff
@@ -232,3 +264,19 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray, n
         p_val, dist = self.test_probs(y_oof, probs_oof, n_ref, n_cur)
         probs_sort = probs_oof[np.argsort(idx_oof)]
         return p_val, dist, probs_sort[:n_ref, 1], probs_sort[n_ref:, 1]
+
+    def _score_rf(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray, np.ndarray]:
+        x_ref, x = self.preprocess(x)
+        x, y, _ = self.get_splits(x_ref, x, return_splits=False)
+        self.model.fit(x, y)
+        # it is possible that some inputs do not have OOB scores. This is probably means that too few trees were
+        # used to compute any reliable estimates. In this case shall we raise an error or keep going by
```
arnaudvl, line 273: Can probably only keep the first line of this comment since sklearn would throw the appropriate warning and the other 2 lines of the comment are more of a question for reviewers.

alibi_detect/cd/sklearn/classifier.py
```diff
@@ -232,3 +264,19 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray, n
...
+        # selecting only not NaN? Now we select the not NaN.
+        index_oob = np.where(np.all(~np.isnan(self.model.oob_decision_function_), axis=1))[0]
```
arnaudvl, line 275: More of a general comment, not to be addressed in this PR: some things reoccur often across many functions in the library, such as getting indices of arrays. Sometimes these are called `idx_*` or `ids_*`, `id_*` or here `index_*`. We should make this consistent across the library. @ascillitoe @jklaise

alibi_detect/cd/sklearn/classifier.py
```diff
@@ -232,3 +264,19 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray, n
...
+        probs_oob = self.model.oob_decision_function_[index_oob]
+        y_oob = y[index_oob]
+        # comparison due to ordering in get_split (i.e, x = [x_ref, x])
+        n_ref = np.sum(index_oob < len(x_ref)).item()
+        n_cur = np.sum(index_oob >= len(x_ref)).item()
```
arnaudvl, line 280: Change name to e.g. `n_test`?

RobertSamoilescu: The naming is consistent with the other backend implementations. For consistency I will keep it as it is.

arnaudvl (approved):

Issue #428, opened 2022-01-14
Consolidate notebook example dependencies


ascillitoe:
The numerous `!pip install`'s in example notebooks mean the test environment is not necessarily what we think it is when running notebook tests. 

We expect the installed dependencies to be those listed by the `python -m pip freeze` in `.github/workflows/test_all_notebooks.yml`. However, as an example, the `!pip install alibi` in `doc/source/examples/cd_chi2ks_adult.ipynb` means other dependencies will be installed (and `numpy`, `tensorflow` versions could even be changed if `alibi stable` isn't in sync with `alibi-detect latest`, which it probably won't be). Even worse, the installed dependencies are currently changing between notebooks. 

We should factor out all the dependencies installed in example notebooks into either the `extras_require = {"examples": ...}` or a `requirements/examples.txt` or similar. These additional dependencies can then be installed in `test_all_notebooks.yml` and `test_changed_notebooks.yml`.

PR #431, opened 2022-01-17
Final prep for v0.8.1 release

ascillitoe:
TODO:

- [x] Deprecate Python 3.6.
- [x] Add bounds for `numba` version.
- [x] https://github.com/SeldonIO/alibi-detect/issues/428
- [x] CHANGELOG.md


ascillitoe:
@mauicv taking this off draft to run tests, it's not actually ready yet.

ascillitoe:
Wrt to consolidating the examples dependencies, a few considerations:

- I've only included dependencies from notebooks that are included in tests. For example, `wilds` isn't included, since the `cd_online_camelyon.ipynb` and `cd_text_amazon.ipynb` notebooks are excluded from the tests.
- I've got rid of the `"examples"` in `extras_require`, and moved it to `testing/requirements.txt`. Thinking behind this is: 1) we don't actually expose the `pip install .[examples]` capability anywhere in the `alibi-detect` docs, and currently all necessary `!pip install` calls are documented in the notebooks anyway. 2) Adding dependencies to `setup.py` makes them more formal dependencies in my view (even if they're in `extras_require`), and brings them under dependabot etc. I'm not sure this is desirable for packages like `TimeSynth` which aren't in PyPi. 
- I've gone with `testing/requirements.txt` instead of `requirements/examples.txt`, since this list of dependencies is private/internal i.e. its only intended for testing usage, and doesn't include all dependencies used in examples (such as `wilds`).

@jklaise @mauicv The above are all relatively opinionated decisions, so happy to change if you disagree!

mauicv (approved):
All looks good to me!

testing/requirements.txt
```diff
@@ -0,0 +1,11 @@
+# Dependencies installed (with !pip install) in example notebooks are listed below.
+# Note: These dependencies are installed during the `test_all_notebooks` and 
+# `test_changed_notebooks` github workflows. The dependencies of notebooks excluded 
+# from testing (see `EXCLUDE_NOTEBOOKS` in `testing/test_notebooks.py`) are not 
+# included here.
+alibi==0.6.2
```
jklaise, line 6: Any reason to pin this?

jklaise: Yea I think latest stable (via `alibi<1.0.0`) is good, also might catch some bugs between libraries.

ascillitoe: Ok sounds good!

jklaise (approved):

PR #438, opened 2022-01-25
Fix issue with pytorch MMDDrift usage in imdb example

ascillitoe:
This PR clarifies the use of the MMDDrift detector (with pytorch backend) in the IMDB example. An error is raised because the reference data is passed to the detector as a NumPy ndarray of `str`'s. For non-numeric data (e.g. text data), data should be passed as a list. 

The notebook is fixed by changing the `X_ref`, `X_h0` and `X_emb` variables to be `List[str]` instead of NumPy `_str` ndarrays. Since the `model` passed to `preprocess_fn` in the "Train embeddings from scratch" section expects a `ndarray`, a `preprocess_batch_fn` is added here to convert the lists to `ndarray`'s internally. Since this is a nice opportunity to make the `preprocess_batch_fn` vs `preprocess_fn` distinction clear, a note has been added explaining this.

### DIFF
--- a/doc/source/examples/cd_online_camelyon.ipynb
+++ b/doc/source/examples/cd_online_camelyon.ipynb
@@ -18,7 +18,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This notebook requires the `wilds`, `torch` and `torchivision` packages which can be installed via `pip`:"
+    "This notebook requires the `wilds`, `torch` and `torchvision` packages which can be installed via `pip`:"
    ]
   },
   {
--- a/doc/source/examples/cd_online_camelyon.ipynb
+++ b/doc/source/examples/cd_online_camelyon.ipynb
@@ -11,13 +11,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This notebook demonstrates a typical workflow for applying online drift detectors to streams of image data. For those unfamiliar with how the online drift detectors operate in `alibi_detect` we recommend first checking out the more introductory example [Online Drift Detection on the Wine Quality Dataset](https://docs.seldon.io/projects/alibi-detect/en/latest/examples/cd_online_wine.html) where online drift detection is performed for the wine quality dataset.\n",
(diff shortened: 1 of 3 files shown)


doc/source/examples/cd_online_camelyon.ipynb
```diff
@@ -11,13 +11,7 @@
...
    "metadata": {},
    "source": [
-    "This notebook demonstrates a typical workflow for applying online drift detectors to streams of image data. For those unfamiliar with how the online drift detectors operate in `alibi_detect` we recommend first checking out the more introductory example [Online Drift Detection on the Wine Quality Dataset](https://docs.seldon.io/projects/alibi-detect/en/latest/examples/cd_online_wine.html) where online drift detection is performed for the wine quality dataset.\n",
-    "\n",
-    "Install the `wilds` library to fetch the dataset used in the example:\n",
-    "\n",
-    "```bash\n",
-    "pip install wilds\n",
-    "```"
+    "This notebook demonstrates a typical workflow for applying online drift detectors to streams of image data. For those unfamiliar with how the online drift detectors operate in `alibi_detect` we recommend first checking out the more introductory example [Online Drift Detection on the Wine Quality Dataset](https://docs.seldon.io/projects/alibi-detect/en/latest/examples/cd_online_wine.html) where online drift detection is performed for the wine quality dataset."
```
arnaudvl, line 14: Removed the wilds package install instructions?

ascillitoe: Only because its already in the explicit pip install below, see here for rendered example:

https://github.com/SeldonIO/alibi-detect/blob/52f542301c9d945f768728fc627ba497e882b785/doc/source/examples/cd_online_camelyon.ipynb

ascillitoe: p.s. just noticed typo in `torchvision`, will fix!

arnaudvl (approved):

ascillitoe:
Thanks @arnaudvl!

PR #442, opened 2022-02-02
Updated readme and docs with conda install instructions

ascillitoe:
- Updated "Getting Started" section of docs with new tabs style install instructions:

image

- README.md has been updated by @sugatoray with conda install instructions, and automatic python version badges etc using PyPI classifiers.


doc/source/overview/getting_started.md
```diff
@@ -2,11 +2,112 @@
...
+```bash
+pip install alibi-detect[prophet]
+```
+````
+
+`````
+``````
+
+
+``````{dropdown} Install via conda-forge
```
jklaise, line 64: Is it possible to install via `conda` but with `prophet`? If not, then probably not worth mixing up `pip` and `conda` and just sticking with the `pip` recommended route for that functionality.

jklaise (approved):
LGTM, nice dropdowns and panels!

doc/source/overview/getting_started.md
ascillitoe: I've not included the "prophet" tab within the "conda" drop-down for now, as haven't properly tested how well `prophet` works with conda, and also agreed that it would be confusing to be mentioning pip installs in the conda section.

image

We can deffo explore installing prophet with conda in the future though (or even as part of this PR if you think its important).

jklaise: Happy to leave it as is for now. I think this will come back as part of the dependency management work as we'll need to figure out how to do the options via `conda`.

ascillitoe: Sounds good!

mauicv (approved):
Looks good, sphinx-design seems great!

PR #454, opened 2022-03-02
WIP: Context aware drift detector

ascillitoe:
This PR implements the context-aware drift detector `ContextAwareDrift` based on @ojcobb and @arnaudvl's recent paper. 

**Note**: This PR implements the code only. The docs and examples will form part of a subsequent PR into the `context_drift` branch. 

## Example(s)

An example notebook demonstrating a basic use case is available here:
https://gist.github.com/ascillitoe/689f748f16962d5a0ebd49c8ea09a6c6

## Outstanding tasks
- [x] Tensorflow backend.
- [x] Tests.
- [x] Docs (methods and examples) - **Delay until subsequent PR into `context_drift`**.
- [ ] Improve docstrings (I've left these very sparse for now, and will update based on docs, in order to ensure notation and terms match).
- [ ] More detailed performance checks. See https://github.com/SeldonIO/alibi-detect/pull/454#issuecomment-1060599156 for discussion. @arnaudvl if we could run one of your heavier examples with both `pytorch` and `tensorflow` that should give us an idea of real-world performance of the TF backend.
- [x] Check difference in `GaussianRBF` kernels.
- [x] Think about whether to include `update_ref` option. If so need to include for `c_ref` too.

### DIFF
--- a/alibi_detect/cd/_domain_clf.py
+++ b/alibi_detect/cd/_domain_clf.py
@@ -11,7 +11,7 @@ class _DomainClf(ABC):
     currently hardcoded into the detector. Therefore, for now, these classes (and the domain_clf submodule) are
     kept private. This is subject to change in the future.
 
-    The classifiers should be fit on conditioning variables `c_all` and their domain `bools` as input (`0` for ref, `1`
+    The classifiers should be fit on conditioning variables `x` and their domain `y` (`0` for ref, `1`
     test). They should predict propensity scores (probability of being test instances) as output.
     Classifiers should possess a calibrate method to calibrate the propensity scores.
     """
@@ -20,21 +20,21 @@ class _DomainClf(ABC):
         raise NotImplementedError()
 
     @abstractmethod
-    def fit(self, c_all: np.ndarray, bools: np.ndarray):
+    def fit(self, x: np.ndarray, y: np.ndarray):
         raise NotImplementedError()
 
     @abstractmethod
-    def calibrate(self, c_all: np.ndarray, bools: np.ndarray):
+    def calibrate(self, x: np.ndarray, y: np.ndarray):
         raise NotImplementedError()
 
     @abstractmethod
-    def predict(self, c_all: np.ndarray) -> np.ndarray:
+    def predict(self, x: np.ndarray) -> np.ndarray:
         raise NotImplementedError()
 
 
 class _SVCDomainClf(_DomainClf):
     def __init__(self,
-                 c_kernel: Callable,
+                 kernel: Callable,
                  cal_method: str = 'sigmoid',
                  clf_kwargs: dict = None):
         """
@@ -44,7 +44,7 @@ class _SVCDomainClf(_DomainClf):
 
         Parameters
         ----------
-        c_kernel
+        kernel
             Kernel used to pre-compute the kernel matrix from data matrices.
         cal_method
             The method to be used to calibrate the detector. This should be a method accepted by the scikit-learn
@@ -52,52 +52,52 @@ class _SVCDomainClf(_DomainClf):
         clf_kwargs
             A dictionary of keyword arguments to be passed to the :py:class:`~sklearn.svm.SVC` classifier.
         """
-        self.c_kernel = c_kernel
+        self.kernel = kernel
         self.cal_method = cal_method
         clf_kwargs = clf_kwargs or {}
-        self.clf = SVC(kernel=self.c_kernel, **clf_kwargs)
+        self.clf = SVC(kernel=self.kernel, **clf_kwargs)
 
-    def fit(self, c_all: np.ndarray, bools: np.ndarray):
+    def fit(self, x: np.ndarray, y: np.ndarray):
         """
         Method to fit the classifier.
 
         Parameters
         ----------
-        c_all
+        x
             Array containing conditioning variables for each instance.
-        bools
+        y
             Boolean array marking the domain each instance belongs to (`0` for reference, `1` for test).
         """
         clf = self.clf
-        clf.fit(c_all, bools)
+        clf.fit(x, y)
         self.clf = clf
 
-    def calibrate(self, c_all: np.ndarray, bools: np.ndarray):
+    def calibrate(self, x: np.ndarray, y: np.ndarray):
         """
         Method to calibrate the classifier's predicted probabilities.
 
         Parameters
         ----------
-        c_all
+        x
             Array containing conditioning variables for each instance.
-        bools
+        y
             Boolean array marking the domain each instance belongs to (`0` for reference, `1` for test).
         """
         clf = CalibratedClassifierCV(self.clf, method=self.cal_method, cv='prefit')
-        clf.fit(c_all, bools)
+        clf.fit(x, y)
         self.clf = clf
 
-    def predict(self,  c_all: np.ndarray) -> np.ndarray:
+    def predict(self,  x: np.ndarray) -> np.ndarray:
         """
         The classifier's predict method.
 
         Parameters
         ----------
-        c_all
+        x
             Array containing conditioning variables for each instance.
 
         Returns
         -------
         Propensity scores (the probability of being test instances).
         """
-        return self.clf.predict_proba(c_all)[:, 1]
+        return self.clf.predict_proba(x)[:, 1]
--- a/alibi_detect/cd/_domain_clf.py
+++ b/alibi_detect/cd/_domain_clf.py
@@ -5,9 +5,11 @@ from sklearn.svm import SVC
 from sklearn.calibration import CalibratedClassifierCV
 
 
-class DomainClf(ABC):
+class _DomainClf(ABC):
     """
-    Base class for domain classifiers used in :py:class:`~alibi_detect.cd.ContextAwareDrift`.
+    Base class for domain classifiers used in :py:class:`~alibi_detect.cd.ContextAwareDrift`.The `SVCDomainClf` is
+    currently hardcoded into the detector. Therefore, for now, these classes (and the domain_clf submodule) are
+    kept private. This is subject to change in the future.
 
     The classifiers should be fit on conditioning variables `c_all` and their domain `bools` as input (`0` for ref, `1`
     test). They should predict propensity scores (probability of being test instances) as output.
@@ -30,7 +32,7 @@ class DomainClf(ABC):
         raise NotImplementedError()
 
 
-class SVCDomainClf(DomainClf):
+class _SVCDomainClf(_DomainClf):
     def __init__(self,
                  c_kernel: Callable,
                  cal_method: str = 'sigmoid',
--- a/alibi_detect/cd/pytorch/context_aware.py
+++ b/alibi_detect/cd/pytorch/context_aware.py
@@ -11,6 +11,8 @@ logger = logging.getLogger(__name__)
 
 
 class ContextMMDDriftTorch(BaseContextMMDDrift):
+    lams: Optional[Tuple[torch.Tensor, torch.Tensor]] = None
+
     def __init__(
             self,
             x_ref: Union[np.ndarray, list],
@@ -23,7 +25,6 @@ class ContextMMDDriftTorch(BaseContextMMDDrift):
             c_kernel: Callable = GaussianRBF,
             n_permutations: int = 1000,
             prop_c_held: float = 0.25,
-            lams: Union[int, Tuple[float, float]] = 20,
             n_folds: int = 5,
             batch_size: Optional[int] = 256,
             device: Optional[str] = None,
@@ -58,9 +59,6 @@ class ContextMMDDriftTorch(BaseContextMMDDrift):
             Number of permutations used in the permutation test.
         prop_c_held
             Proportion of contexts held out to condition on.
-        lams
-            Ref and test regularisation parameters. Either a tuple containing the two parameters as floats, or an
-            int defining the list of parameters to search over via `[2**(-i) for i in range(lams)]`.
         n_folds
             Number of cross-validation folds used when tuning the regularisation parameters.
         batch_size
@@ -86,7 +84,6 @@ class ContextMMDDriftTorch(BaseContextMMDDrift):
             c_kernel=c_kernel,
             n_permutations=n_permutations,
             prop_c_held=prop_c_held,
-            lams=lams,
             n_folds=n_folds,
             batch_size=batch_size,
             input_shape=input_shape,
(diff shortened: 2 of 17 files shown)


ascillitoe:
### Definition of domain classifier

@ojcobb @arnaudvl this point might require a bit more discussion. At the moment I've set the domain classifier kwarg as:
```
domain_clf: Union[Type[DomainClf], DomainClf] = SVCDomainClf
```

where `SVCDomainClf` is a subclass of `DomainClf`. Both of these are public, currently accessible under `alibi_detect.cd.domain_clf`. 

Internally we then do:
https://github.com/SeldonIO/alibi-detect/blob/d9edc6bbeea7538d308316c443fa763b3be1885f/alibi_detect/cd/base.py#L1025-L1032

This approach allows for 3 options:

1. Don't set `domain_clf`. It then defaults to `SVCDomainClf`, and this **class** is instantiated internally with `self.clf = domain_clf(self.c_kernel)`. 
2. Set `domain_clf` as an **object** (an instantiated subclass of `DomainClf`). This **object** is already ready to go so `self.clf=domain_clf`. An example use case would be to set `domain_clf` as a `SVCDomainClf` with non-default args/kwargs, but it could also be an instantiated custom subclass of `DomainClf`. 
3. Set `domain_clf` as a generic **subclass** of `DomainClf`. This will then be instantiated with no args/kwargs i.e. `self.clf = domain_clf()`.

The idea behind expecting (and telling) the user to make their custom `domain_clf` a subclass of `DomainClf` is that we can then have more control over what the `domain_clf` looks like. Admittedly at the moment it doesn't actually add much, since having `__call__` as an @abstractmethod isn't much different from just typing as `domain_clf: Callable`. However, this approach would work nicely if we add runtime typechecking later, plus I'd hope a user would glance at the `BaseClf` docs and signature when writing their subclass. 

Would be good to get your thoughts on this. In particular, @arnaudvl I could do with understanding what flexibility/capability you need for this wrt to the examples you've put together. 

p.s. If we go with the above, I would need to tighten the checks shown in the code snippet as currently passing in an object which isn't of `DomainClf` type will result in an ugly error...

arnaudvl:
@ascillitoe nice work! A few initial thoughts and things to check:
- [ ] Accounting of the diagonal for the built-in GaussianRBF kernel vs. the context-aware drift approach is different. Check `+n` term [here](https://github.com/ascillitoe/alibi-detect/blob/context_drift/alibi_detect/utils/pytorch/kernels.py#L52) vs. research code. From our experiments this impacts the detector's performance.
- [ ] Settle on default heuristic as well as instances and/or contexts used when the kernel is configured at the first pass [here](https://github.com/ascillitoe/alibi-detect/blob/context_drift/alibi_detect/cd/pytorch/context_aware.py#L147). The decision here could also affect the above bullet point.
- [ ] Test efficiency einsum operations [from here onwards](https://github.com/ascillitoe/alibi-detect/blob/context_drift/alibi_detect/cd/pytorch/context_aware.py#L188). While nice to look at, it isn't (wasn't?) always the most efficient way as shown e.g. [here](https://github.com/pytorch/pytorch/issues/32591), [here](https://github.com/pytorch/pytorch/issues/10661) or [here](https://discuss.pytorch.org/t/einsum-running-100x-slower-than-expected-when-controlling-for-number-of-ops/89339). Ideally would be benchmarked vs. the "manual" equivalent.
- [ ] Maybe rename [`cond_prop`](https://github.com/ascillitoe/alibi-detect/blob/context_drift/alibi_detect/cd/base.py#L954) to `prop_c_held` or similar? Reason is that we don't really use conditioning in the naming of the variables and this variable tells which proportion of the context `c` is held out of the test data. 
- [ ] We use `bools == 0` and `bools == 1` a lot [from here onwards](https://github.com/ascillitoe/alibi-detect/blob/context_drift/alibi_detect/cd/pytorch/context_aware.py#L167). Can we instead just do `inds_0, inds_1 = bools == 0, bools == 1` and use `inds_0, inds_1` from then on instead of redoing the `==` each time?
- [ ] As already highlighted and to be checked with @ojcobb , 20 seems pretty arbitrary [here](https://github.com/ascillitoe/alibi-detect/blob/context_drift/alibi_detect/cd/pytorch/context_aware.py#L174). Can of course set this as a default kwarg. Similar with `n_folds` on the next line.
- [ ] Given that the implementation is an MMD-based instantiation of the more general framework, does it make sense to rename the implemented method from `ContextAwareDrift` to `ContextAwareMMDDrift`? Not exactly the most catchy name I must admit...
- [ ] On the `domain_clf` topic, I think I would be in favour of a more traditional "fit -> predict" implementation instead of bundling everything in `__call__`. This feels more natural in my opinion as it doesn't differ from e.g. your usual scikit-learn classifier, where now the contexts are your data and the domain your labels. The default fit step would then also encompass the optional calibration. Users also then don't have to figure out what they need to do under an unfamiliar API.

ascillitoe:
> * Test efficiency einsum operations [from here onwards](https://github.com/ascillitoe/alibi-detect/blob/context_drift/alibi_detect/cd/pytorch/context_aware.py#L188). While nice to look at, it isn't (wasn't?) always the most efficient way as shown e.g. [here](https://github.com/pytorch/pytorch/issues/32591), [here](https://github.com/pytorch/pytorch/issues/10661) or [here](https://discuss.pytorch.org/t/einsum-running-100x-slower-than-expected-when-controlling-for-number-of-ops/89339). Ideally would be benchmarked vs. the "manual" equivalent.

I've run some simple tests for [PyTorch](https://colab.research.google.com/drive/1e1ShsF6bZtZwAPgXpqWfxOYLM4Q6LC7h?usp=sharing) and [Tensorflow](https://colab.research.google.com/drive/14lvcpF8F-ysOCX6oaMe8c3ZRge_gGDZ5?usp=sharing) (follow links for notebooks). Conclusions appears to be that `einsum` is approx. equivalent to explicit operations with PyTorch, and sometimes even faster than the explicit op. with TensorFlow.

ascillitoe:
> * As already highlighted and to be checked with @ojcobb , 20 seems pretty arbitrary [here](https://github.com/ascillitoe/alibi-detect/blob/context_drift/alibi_detect/cd/pytorch/context_aware.py#L174). Can of course set this as a default kwarg. Similar with `n_folds` on the next line.

I've added a `n_folds` kwarg, and have incorporated functionality to specify the number of lambda's to try within the `lams` kwarg: 

```python
lams: Union[int, Tuple[float, float]] = 20
```

So either `lams` is a tuple of `float`'s representing the lamda's to use, or it is an `int` specifying the number of lamda's to try. We could also have an option for `Tuple[List[float], List[float]]`, where the list's define the lambda's to try (instead of looping through `[2**(-i) for i in range(lams)]`), however I've left this out for now to reduce complexity.

ascillitoe:
> * Given that the implementation is an MMD-based instantiation of the more general framework, does it make sense to rename the implemented method from `ContextAwareDrift` to `ContextAwareMMDDrift`? Not exactly the most catchy name I must admit...

Renamed to `ContextMMDDrift` for now as a little less wordy. Could also do `MMDDriftContext` to line up more with `MMDDriftOnline`. Thoughts?

ascillitoe:
> [ ] On the `domain_clf` topic, I think I would be in favour of a more traditional "fit -> predict" implementation instead of bundling everything in `__call__`. This feels more natural in my opinion as it doesn't differ from e.g. your usual scikit-learn classifier, where now the contexts are your data and the domain your labels. The default fit step would then also encompass the optional calibration. Users also then don't have to figure out what they need to do under an unfamiliar API.

Based on offline discussion, have removed the `domain_clf` kwarg and hardcoded for now.

ascillitoe:
> @ascillitoe nice work! A few initial thoughts and things to check:

@arnaudvl I've ticked some of these and responded in the above comments, please feel free to untick anything you feel needs further discussion!

arnaudvl:
> > * As already highlighted and to be checked with @ojcobb , 20 seems pretty arbitrary [here](https://github.com/ascillitoe/alibi-detect/blob/context_drift/alibi_detect/cd/pytorch/context_aware.py#L174). Can of course set this as a default kwarg. Similar with `n_folds` on the next line.
> 
> I've added a `n_folds` kwarg, and have incorporated functionality to specify the number of lambda's to try within the `lams` kwarg:
> 
> ```python
> lams: Union[int, Tuple[float, float]] = 20
> ```
> 
> So either `lams` is a tuple of `float`'s representing the lamda's to use, or it is an `int` specifying the number of lamda's to try. We could also have an option for `Tuple[List[float], List[float]]`, where the list's define the lambda's to try (instead of looping through `[2**(-i) for i in range(lams)]`), however I've left this out for now to reduce complexity.

This could lead to confusion as `lambdas` now contains 2 separate functionalities? Probably in favour of splitting this out.

ascillitoe:
> > > * As already highlighted and to be checked with @ojcobb , 20 seems pretty arbitrary [here](https://github.com/ascillitoe/alibi-detect/blob/context_drift/alibi_detect/cd/pytorch/context_aware.py#L174). Can of course set this as a default kwarg. Similar with `n_folds` on the next line.
> > 
> > 
> > I've added a `n_folds` kwarg, and have incorporated functionality to specify the number of lambda's to try within the `lams` kwarg:
> > ```python
> > lams: Union[int, Tuple[float, float]] = 20
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So either `lams` is a tuple of `float`'s representing the lamda's to use, or it is an `int` specifying the number of lamda's to try. We could also have an option for `Tuple[List[float], List[float]]`, where the list's define the lambda's to try (instead of looping through `[2**(-i) for i in range(lams)]`), however I've left this out for now to reduce complexity.
> 
> This could lead to confusion as `lambdas` now contains 2 separate functionalities? Probably in favour of splitting this out.

'lams' kwarg now hardcoded again. 'n_folds' is retained.

ascillitoe:
**Note**: tests are not running for this branch as we only run CI for PR's into `master`. @jklaise @arnaudvl we probably need to follow up on the discussion about github branches etc. If we decide to continue using more branches we might need to change this.

alibi_detect/cd/pytorch/context_aware.py
```diff
@@ -0,0 +1,264 @@
...
+        Tensor of instances with dimension [Ny, features].
+    dist
+        Tensor with dimensions [Nx, Ny], containing the pairwise distances between `x` and `y`.
+
+    Returns
+    -------
+    The computed bandwidth, `sigma`.
+    """
+    n = min(x.shape[0], y.shape[0])
+    n_median = (np.prod(dist.shape) - n) // 2
```
ojcobb, line 262: Whilst I doubt it would make a tangible difference, now that I look at this again I think that really it should just be `n_median = np.prod(dist.shape) // 2`

alibi_detect/cd/pytorch/context_aware.py
```diff
@@ -0,0 +1,264 @@
...
+
+        This private method function uses cross validation to select the regularisation parameter that
+        minimises squared error on the out-of-fold instances. The error is a distance in the RKHS and is
+        therefore an MMD-like quantity itself.
+        """
+        n = len(L)
+        fold_size = n // n_folds
+        K, L = K.type(torch.float64), L.type(torch.float64)
+        losses = torch.zeros_like(lams, dtype=torch.float).to(K.device)
+        for fold in range(n_folds):
```
ojcobb, line 226: A change I made to my private repo since sharing the prototype is to shuffle K and L prior to running the cross-val procedure. Would be important in cases where the user doesn't shuffle the data themselves and the ordering is non-random. I recommend adding
```
perm = torch.randperm(n)
K, L = K[perm][:,perm], L[perm][:,perm]
```

alibi_detect/cd/tensorflow/context_aware.py
```diff
@@ -0,0 +1,279 @@
...
+        Tensor of instances with dimension [Ny, features].
+    dist
+        Tensor with dimensions [Nx, Ny], containing the pairwise distances between `x` and `y`.
+
+    Returns
+    -------
+    The computed bandwidth, `sigma`.
+    """
+    n = min(x.shape[0], y.shape[0])
+    n_median = (tf.math.reduce_prod(dist.shape) - n) // 2
```
ojcobb, line 277: See comment on pytorch analogue.

alibi_detect/cd/tensorflow/context_aware.py
```diff
@@ -0,0 +1,279 @@
...
+
+        This private method function uses cross validation to select the regularisation parameter that
+        minimises squared error on the out-of-fold instances. The error is a distance in the RKHS and is
+        therefore an MMD-like quantity itself.
+        """
+        n = len(L)
+        fold_size = n // n_folds
+        K, L = tf.cast(K, tf.float64), tf.cast(K, tf.float64)
+        losses = tf.zeros_like(lams, dtype=tf.float64)
+        for fold in range(n_folds):
```
ojcobb, line 217: See comment on pytorch analogue.

alibi_detect/cd/pytorch/context_aware.py
ascillitoe: Ah oops, I shuffled for the held out split but not for this. Will update!

ojcobb:
Great work @ascillitoe - found very few issues with the code! I will have many more comments on docstrings etc in the follow-up pr but I've left those for now.

My only point of contention at the moment is that we've tried to create an abstraction around the domain classifier (i.e. `DomainClf`) but coupled its use very highly to the specifics of `SVCDomainClf`. Calling `.fit()` and `.predict()` on all of the data would not be a good idea in general. However I think this is probably something that we can return to once we have a better idea of the flexibility we wish to expose on this front.

ascillitoe:
> Great work @ascillitoe - found very few issues with the code! I will have many more comments on docstrings etc in the follow-up pr but I've left those for now.
> 
> My only point of contention at the moment is that we've tried to create an abstraction around the domain classifier (i.e. `DomainClf`) but coupled its use very highly to the specifics of `SVCDomainClf`. Calling `.fit()`, `.calibrate()` and `.predict()` on all of the data would not be a good idea in general. However I think this is probably something that we can return to once we have a better idea of the flexibility we wish to expose on this front.

Thanks @ojcobb! I'll address those issues you highlighted shortly. I agree about `DomainClf` in its current state. I only really left it in as thinking about when we expose a domain classifier kwarg later. Perhaps for now the easiest thing to do is to simply make the entire `domain_clf` submodule private? In the future when it is exposed, I think we'll need to document its use very carefully and suggest cross-validation etc for more models with more DoF's?

alibi_detect/cd/_domain_clf.py
```diff
@@ -0,0 +1,103 @@
+from abc import ABC, abstractmethod
+from typing import Callable
+import numpy as np
+from sklearn.svm import SVC
+from sklearn.calibration import CalibratedClassifierCV
+
+
+class _DomainClf(ABC):
```
arnaudvl, line 8: Does it make sense throughout this file to make the naming more conventional, e.g. `x, y, kernel` instead of `c_all, bools, c_kernel`?

ascillitoe: Fair point, that might be more intuitive for those familiar with the scikit-learn style API

arnaudvl (approved):

PR #458, opened 2022-03-14
Config driven detectors - part 1

ascillitoe:
This is the first of a series of PR's for the config-driven detector functionality. The original PR (#389) has been split into a number of smaller PR's to aid the review process. The PR's will be as follows:

1. Modifications to detector kwargs, and addition of `.get_config` methods to support subsequent PR's.
2. The main save/load functionality in `saving.py` and `loading.py`, pydantic validation, and documentation.
3. Testing.

These will be merged into the `config_driven_detectors` branch, before a final merge into `master`. 

**Note**: The full original PR #389 has been split up to aid the review process. This initial PR is **not** fully functioning, in the sense that the `alibi_detect.utils.saving` submodule still expects the old `preprocess_x_ref` kwarg and attribute. I don't think it is worth reworking `saving` with these updated kwarg's if we are soon to significantly rework the submodule anyway... 

### Summary of PR

This PR includes two main changes:
#### 1. Modification of detector kwarg's
- The `preprocess_x_ref` kwarg for all offline detectors is separated into two kwargs, `x_ref_preprocessed` and `preprocess_at_init`, in order to draw a distinction between whether or not to preprocess x_ref (x_ref_preprocessed), and when to preprocess x_ref (preprocess_at_init). See https://github.com/SeldonIO/alibi-detect/pull/389#discussion_r817960156 and https://github.com/SeldonIO/alibi-detect/pull/381.
- Addition of `input_shape` kwarg. This was missing from some detectors, and has been added as is used when saving preprocessing models (see https://github.com/SeldonIO/alibi-detect/pull/389#discussion_r817961504).

#### 2. Addition of `get_config` methods
This method has been added to each detector. It takes a detector's attributes and returns a config dictionary. This config dictionary will form the basis of the later saving work, with artifacts in the config being saved, before the config dict is saved as a toml file.

The `get_config` methods are added following the usual hierarchical approach wrt to the base classes and backend specific subclasses. For example the generalizable bits of the config are defined in a mxin in `alibi_detect/base.py`, the backend agnostic bits are defined in `alibi_detect/cd/base.py`, and the backend specific bits are then defined in the backend subclasses e.g. `MMDDriftTF`.

Complexity arises here because as some kwarg's are conditional on others, and the inital kwarg's are not always added as attributes, therefore can be complicated to infer the original values of the kwarg's at `get_config` time (see https://github.com/SeldonIO/alibi-detect/pull/389#discussion_r817969708).


alibi_detect/base.py
```diff
@@ -94,6 +93,27 @@ def infer_threshold(self, X: np.ndarray) -> None:
         pass
 
 
+class DriftConfigMixin:
```
jklaise, line 96: Short docstring about what this class is for would be useful (same stuff as in the PR comments really).

alibi_detect/utils/warnings.py
```diff
@@ -0,0 +1,32 @@
...
+def _rename_kwargs(func_name: str, kwargs: Dict[str, Any], aliases: Dict[str, str]):
+    """
+    Private function to rename deprecated kwarg to new name, and raise DeprecationWarning.
+    """
+    for alias, new in aliases.items():
+        if alias in kwargs:
+            if new in kwargs:
+                raise ValueError(f"{func_name} received both the deprecated kwarg `{alias}` "
+                                 f"and it's replacement `{new}`.")
+            warnings.warn(f'`{alias}` is deprecated; use `{new}`.', DeprecationWarning, 3)
```
jklaise, line 31: Question - why stacklevel 3?

Also, I don't think we fully arrived at a conclusion whether using `UserWarning` over `DeprecationWarning` would be better. `UserWarning` is always shown and won't be turned off by Python optimizations, whereas it's not the case for `DeprecationWarning` which is more intended for developers. That being said, the users of `alibi` are developers so the point is perhaps not important.

alibi_detect/cd/mmd.py
```diff
@@ -92,6 +100,7 @@ def __init__(
         else:
             self._detector = MMDDriftTorch(*args, **kwargs)  # type: ignore
         self.meta = self._detector.meta
+        self._detector.backend = backend
```
jklaise, line 103: I assume this info is needed by the backend class so it can call it's own specific `get_config` method and save that info?

alibi_detect/cd/pytorch/lsdd.py
```diff
@@ -165,3 +174,24 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+        Get the detector's configuration dictionary.
+
+        Returns
+        -------
+        The detector's configuration dictionary.
+        """
+        cfg = super().get_config()
+
+        # backend
+        cfg.update({'backend': 'pytorch'})
```
jklaise, line 189: I thought the `backend` kwarg was set as an attribute of the backend class by the agnostic class, hence wouldn't need to be hard-coded here? (same comment applies to `mmd` etc. below).

jklaise (commented):
LGTM, happy to merge if the questions are answered.

alibi_detect/cd/cvm.py
```diff
@@ -88,3 +95,15 @@ def feature_score(self, x_ref: np.ndarray, x: np.ndarray) -> Tuple[np.ndarray, n
...
+
+    def get_config(self) -> dict:
+        """
+        Get the detector's configuration dictionary.
+
+        Returns
+        -------
+        The detector's configuration dictionary.
+        """
+        cfg = super().get_config()
```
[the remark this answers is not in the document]
ascillitoe: The `get_config` is done at a number of levels. The mixin class does the top-level drift detector generic stuff i.e. adds `version` and `x_ref`, then most of the attributes are added in the more specific subclasses i.e. `BaseUnivariateDrift.get_config()`. Then any final detector-specific bits are added in the detector-specific subclass method. There is nothing specific for `CVMDrift`, but as an example, `FETDrift` adds in the `alternative` field:

https://github.com/SeldonIO/alibi-detect/blob/9eb5ace11967fcc7af02afdbca60f9553096a223/alibi_detect/cd/fet.py#L119

This can't be added in `BaseUnivariateDrift` since not all univariate drift detectors have an `alternative` kwarg. 

P.s. The use of this of this mixin is probably a good argument for having a drift detector base class (in between `BaseDetector` and `BaseUnivariateDrift` etc). However, I didn't want to change the core of the code too much in this PR...

alibi_detect/cd/pytorch/lsdd.py
ascillitoe: This wasn't actually the original intention (the `self.backend` was intended for use at save time, in the later PR). 

However, this is actually a nice idea, since we can then move the `cfg.update({'backend': self.backend})` up to the base class.

mauicv (commented):
Looks good to me!

alibi_detect/cd/base.py
```diff
@@ -317,13 +381,23 @@ def __init__(
         if p_val is None:
             logger.warning('No p-value set for the drift threshold. Need to set it to detect data drift.')
 
-        # optionally already preprocess reference data
-        self.p_val = p_val
-        if preprocess_x_ref and isinstance(preprocess_fn, Callable):  # type: ignore[arg-type]
+        # x_ref preprocessing logic
```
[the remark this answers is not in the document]
ascillitoe: I agree with you that it would certainly be nice to factor out. But, I didn't want to do as part of this PR as it's a change to the core of the detectors and not really related to the config functionality (it only looks as if I've added stuff above as have had to rework the logic a bit due to the kwarg changes).

Also think above is motivation for adding a drift detector base class i.e. `BaseDrift(BaseDetector)` which `BaseUnivariateDrift` would inherit from, rather than adding lots of Mixin's...

alibi_detect/cd/mmd.py
ascillitoe: I've got rid of this, as the `backend` is actually already stored in `self.meta['backend']`. I'm then accessing this in the base classes to reduce duplication a little.

PR #462, opened 2022-03-16
Config driven detectors - part 2

ascillitoe:
This is the second of a series of PR's for the config-driven detector functionality. The original PR (https://github.com/SeldonIO/alibi-detect/pull/389) has been split into a number of smaller PR's to aid the review process.

### Summary of PR
This PR introduces validation of the detector config files using [pydantic](https://pydantic-docs.helpmanual.io/). Pydantic models are defined for each detector in `alibi_detect.utils.schemas`. These are then used in `alibi_detect.utils.loading.validate_config`; a public function to validate a detector's config dictionary.

#### Use cases
The `validate_config` function will be used in a number of ways. Consider the `load_detector` workflow` (functionality to be introduced in part III):

image

`validate_config` will be used:
- On the *unresolved* config (where artefacts are still strings), e.g. `validate_config(cfg)`. 
- After `resolve_config` has been called (so artefacts are now runtime objects such as ndarray's) e.g. `validate_config(cfg, resolved=True)`.
- On raw config dictionaries (when an advanced user has parsed a config.toml themselves with `read_config`, and perhaps has manipulated the config dict). For this use case, the `validate_config` has been made public. 

#### Validate config return
A small (but important!) detail -> The `validate_config` returns a `dict`, which is the output of the pydantic model's `.dict()` method. This is the same config dict that was passed in, but with missing fields populated by their default values. This is a somewhat philosophical design choice (I think), in that it means we internally we always deal with fully populated config dicts. We can set additional metadata etc in `schemas.py` at a later date, and can set defaults for artefact dictionaries etc (see below). A downside is that we will have to ensure the kwarg defaults set in `schemas.py` are kept in sync with the detector kwargs themself (this could also be a positive if we wanted divergence in some cases).

#### Pydantic models for artefacts
Artefacts are specified in a config with dictionaries, which may then have further artefact dictionaries within them (see preprocess_fn`, `model` etc in the `config.toml` example in schematic above). These artefact dictionaries are also validated e.g. see `PreprocessConfig` and `PreprocessConfigResolved` in `schemas.py`. 

### Example
Example notebook: 
https://gist.github.com/ascillitoe/8ea123776151368a7a6a91688a1809cb


alibi_detect/base.py
```diff
@@ -102,16 +102,23 @@ class DriftConfigMixin:
     preprocess_fn: Optional[Callable] = None
 
     def drift_config(self):
+        name = self.__class__.__name__
+        # strip off any backend suffix
+        backends = ['TF', 'Torch', 'Sklearn']
+        for backend in backends:
+            if name.endswith(backend):
+                name = name[:-len(backend)]
```
jklaise, line 110: Doesn't feel right to have to do thi, but at the same time I have no other ideas :)

alibi_detect/cd/tensorflow/mmd.py
```diff
@@ -139,7 +139,7 @@ def get_config(self) -> dict:
         # kernel logic
         if isinstance(self.kernel, GaussianRBF):
             # If default kernel, we don't need to spec
-            sigma = self.kernel.sigma.numpy().tolist() if not self.infer_sigma else None
+            sigma = self.kernel.sigma.numpy() if not self.infer_sigma else None
```
jklaise, line 142: What's the reason for the change `list -> ndarray`?

setup.py
```diff
@@ -46,6 +46,7 @@ def readme():
         "tqdm>=4.28.1, <5.0.0",
         "requests>=2.21.0, <3.0.0",
         "numba>=0.50.0, !=0.54.0, <0.56.0",  # Avoid 0.54 due to: https://github.com/SeldonIO/alibi/issues/466
+        "pydantic>=1.8.0, <1.9.0",
```
jklaise, line 49: Should we only upper bound at `2.0.0` or do you think it's likely to have breaking changes for minor versions?

Btw `1.9.0` is out since Dec 21 so probably should bump it either way. https://github.com/samuelcolvin/pydantic/releases

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
+# mypy: ignore-errors
```
jklaise, line 1: Casting a broad net here?

alibi_detect/utils/loading.py
```diff
@@ -0,0 +1,50 @@
+# type: ignore
```
jklaise, line 1: Does this do anything?

alibi_detect/base.py
ascillitoe: Eurgh I know, I hate it. I wanted to do something clever with checking `self.__class__.__base__` or similar i.e. determine if `self` is a backend-specific detector by checking if its parent class is a detector itself (instead of a base class in `cd/base.py`. The problem is that `MMDDriftTF` doesn't actually inherit `MMDDrift`, since in `MMDDrift` we do `self._detector = MMDDrift()`. So when `MMDDriftTF.get_config()` is called by `MMDDrift.get_config()`, it thinks its parent class is `BaseMMDDrift`. Don't know if that makes sense?

alibi_detect/cd/tensorflow/mmd.py
ascillitoe: I previously had the `tolist()` as when the `get_config` output is used in the later `save_detector` stuff coming in part III, `sigma` needs to be a list so it can be saved to TOML. However, since its typed as ndarray in the pydantic model, I realised this meant the `get_config` output would actually raise a `ValidationError` when passed to `validate_config`. Think its important the `get_config` outputs always return valid configs, so will offload this conversion to list to `saving.py` in part III instead. Think this is in more inline with the general strategy anyway, of `get_config` returning a "resolved" config, and then `saving.py` being responsible for turning it into an "unresolved" config.

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
...
+
+
+class DetectorConfig(CustomBaseModel):
+    """
+    Base detector config schema. Only fields universal across all detectors should be defined in here.
+    """
+    name: str
+    version: str = __version__
+    config_spec: str = __config_spec__
+    backend: Literal['tensorflow', 'pytorch', 'sklearn'] = 'tensorflow'
```
jklaise, line 46: I might be out of the loop, but how come we default to `tensorflow`? There are also detectors that have no backend right?

jklaise: Ah just read the comment below, though I have to say I don't fully understand it. I think we need to be quite careful with the semantics and what `backend` really means, especially if we distinguish between built-in backends for specific drift functionality or user/built-in `preprocess_fn` functionality. Not sure if we want to tackle this right now though.

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
...
+    layers: Optional[List[int]] = None  # TODO - add check conditional on embedding type (see docstring in above)
+    src: str
+
+
+class TokenizerConfig(CustomBaseModel):
+    """
+    Unresolved schema for text tokenizers.
+    """
+    src: str
+    kwargs: Optional[dict] = {}
```
jklaise, line 76: Had to check `pydantic` docs but it's ok as they deepcopy mutable defaults, so we're good here. Just calling it out for others as it is not obvious if you're used to regular Python behaviour.

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
...
+    custom_obj: Optional[dict] = None
+
+
+class EmbeddingConfig(CustomBaseModel):
+    """
+    Unresolved schema for text embeddings.
+    """
+    type: Literal['pooler_output', 'last_hidden_state',
+                  'hidden_state', 'hidden_state_cls']  # See alibi_detect.models.tensorflow.embedding
+    layers: Optional[List[int]] = None  # TODO - add check conditional on embedding type (see docstring in above)
```
jklaise, line 67: Not sure what the TODO/docstring refers to here.

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
...
+
+
+class ModelConfig(CustomBaseModel):
+    """
+    Unresolved schema for (ML) models. Note that the model "backend" e.g. 'tensorflow', 'pytorch', 'sklearn', is set
+    by backend in DetectorConfig.
+    """
+    type: Literal['custom', 'HiddenOutput', 'UAE'] = 'custom'
+    src: str
+    custom_obj: Optional[dict] = None
```
jklaise, line 58: Did we come to some conclusion about `custom` (this is user model I guess?) and especially `custom_obj` from the previous PR comment?

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
...
+
+    # Below kwargs are only passed if function == @preprocess_drift
+    model: Union[str, ModelConfig, None] = None  # TODO - make required field when src is preprocess_drift
+    embedding: Union[str, EmbeddingConfig, None] = None
+    tokenizer: Union[str, TokenizerConfig, None] = None
+    device: Literal['cpu', 'cuda', 'gpu', None] = None
+    preprocess_batch_fn: Optional[str] = None
+    max_len: Optional[int] = None
+    batch_size: Optional[int] = int(1e10)
+    dtype: Optional[str] = None
```
jklaise, line 93: Should the `Union`s with `None` be converted to `Optional`? Or is there some semantic meaning in `None` for those cases?

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
...
+
+
+class KSDriftConfig(DriftDetectorConfig):
+    """
+    Unresolved schema for KSDrift detector.
+    """
+    preprocess_at_init: bool = True
+    update_x_ref: Optional[Dict[str, int]] = None
+    correction: str = 'bonferroni'
+    alternative: str = 'two-sided'
```
jklaise, line 181: Should be `Literal`s?

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
...
+    n_features: Optional[int] = None
+
+
+class ChiSquareDriftConfig(DriftDetectorConfig):
+    """
+    Unresolved schema for ChiSquareDrift detector.
+    """
+    preprocess_at_init: bool = True
+    update_x_ref: Optional[Dict[str, int]] = None
+    correction: str = 'bonferroni'
```
jklaise, line 191: `Literal`?

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
...
+    """
+    Unresolved schema for MMDDrift detector.
+    """
+    preprocess_at_init: bool = True
+    update_x_ref: Optional[Dict[str, int]] = None
+    kernel: Union[str, KernelConfig, None] = None
+    sigma: Optional[List[float]] = None
+    configure_kernel_from_x_ref: bool = True
+    n_permutations: int = 100
+    device: Literal['cpu', 'cuda', None] = None
```
jklaise, line 239: `Optional[Literal['cpu', 'cuda']]` ? Same question for below.

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
...
+    """
+    pass
+
+
+class CVMDriftConfigResolved(DriftDetectorConfigResolved, CVMDriftConfig):
+    pass
+
+
+class FETDriftConfigResolved(DriftDetectorConfigResolved, FETDriftConfig):
+    pass
```
jklaise, line 353: Are these to be done or are they good to go because everything is defined in the base models?

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
...
+
+
+class MMDDriftConfigResolved(DriftDetectorConfigResolved, MMDDriftConfig):
+    """
+    Resolved schema for MMDDrift detector.
+    """
+    kernel: Union[Callable, KernelConfigResolved, None]
+    sigma: Optional[np.ndarray]
```
jklaise, line 361: I'm assuming there is no default value because the default should come from one of the parent models?

alibi_detect/utils/loading.py
```diff
@@ -0,0 +1,50 @@
...
+from alibi_detect.version import __version__
+from alibi_detect.utils.schemas import DETECTOR_CONFIGS, DETECTOR_CONFIGS_RESOLVED, __config_spec__
+import warnings
+
+
+def validate_config(cfg: dict, resolved: bool = False) -> dict:
```
jklaise, line 7: Do we need to unit test this?

alibi_detect/utils/loading.py
```diff
@@ -0,0 +1,50 @@
...
+        raise ValueError('`name` missing from config.toml.')
+
+    # Validate detector specific config
+    if detector_name in DETECTOR_CONFIGS.keys():
+        if resolved:
+            cfg = DETECTOR_CONFIGS_RESOLVED[detector_name](**cfg).dict()  # type: ignore[attr-defined]
+        else:
+            cfg = DETECTOR_CONFIGS[detector_name](**cfg).dict()  # type: ignore[attr-defined]
+    else:
+        raise ValueError('Loading the specified detector from a config.toml is not yet supported.')
```
jklaise, line 36: I guess this error would also be triggered if `name` is invalid/misspelled.

jklaise (commented):
Looks good overall, just wanted some clarification on the conventions for the `pydantic` models, specifically wrt to `Optional` values and `None` defaults.

I also saw a few `TODOs` wrt to conditional checking of required parameters given some other parameters. Is this fairly easy to add (I'm assuming `pydantic` validators)?

alibi_detect/utils/schemas.py
ascillitoe: Yes a very broad net! All the mypy errors in this file are actually `assignment` errors, so we could be more specific. 

But I've struggled to actually fix them. Could do with input wrt to this. The errors are all similar to the following:
```
alibi_detect/utils/schemas.py:102: error: Incompatible types in assignment (expression has type "Callable[..., Any]", base class "PreprocessConfig" defined the type as "str")  [assignment]
```

I think its occuring because I originally set `src: str` in `PreprocessConfig`, but then override it to `src: Callable` in `PreprocessConfigResolved`. I wasn't able to find an obvious simple fix to this, except to not use the approach of having the `Resolved` config's inherit from the original configs. Not doing this would be fine, but just a fair bit more repetition...

alibi_detect/utils/loading.py
ascillitoe: Not right now, I'll remove. As things stand it will be required again in part III, but I will have another think about whether things can be rewritten there so it is not required.

alibi_detect/utils/schemas.py
```diff
@@ -0,0 +1,431 @@
...
+from alibi_detect.utils._types import Literal
+from alibi_detect.version import __version__, __config_spec__
+from alibi_detect.cd.tensorflow import HiddenOutput, UAE
+from alibi_detect.models.tensorflow import TransformerEmbedding
+from transformers import PreTrainedTokenizerBase
+from alibi_detect.utils.frameworks import has_pytorch, has_tensorflow  # , has_sklearn
+
+
+SUPPORTED_MODELS = Union['UAE', 'HiddenOutput', 'tf.keras.Model']
+SupportedModels = [UAE, HiddenOutput]
```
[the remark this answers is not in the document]
ascillitoe: @jklaise @mauicv Would appreciate input on this one.

`SUPPORTED_MODELS` is used for typing, wheras `SupportedModels` is a list of classes, to be used later on in `isinstance` checks i.e. https://github.com/SeldonIO/alibi-detect/blob/223b4cea07f4d767bce50f6f8ccdf065584a25c1/alibi_detect/utils/saving.py#L758. 

Ideally we'd only have one of them. However, I wasn't able to construct the typing `Union` from a list (or tuple) of classes. The reverse should be possible (i.e. use `Union` contents in `isinstance`, see first answer https://stackoverflow.com/questions/45957615/check-a-variable-against-union-type-at-runtime-in-python-3-6), but I wasn't able to get this working in our case since we use forward references (to handle TF or torch not being installed).

alibi_detect/base.py
mauicv: Can we also just overwrite it in the child class? So in the `get_config`we replace the `name` keys value with the correct class name which we know at that point in the code. It's still not a great way of solving it though. I think it's only better in the sense that the general behaviour is captured here and then deviations from that behaviour occur closer to the place they happen. If that makes sense?

alibi_detect/utils/schemas.py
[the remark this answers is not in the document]
jklaise: Ah I see. As a minimum I would document the different uses of the two constants here.

alibi_detect/utils/schemas.py
ascillitoe: Yep I had to check the same!

alibi_detect/utils/schemas.py
ascillitoe: No idea what docstring that is referring to. WIll remove that bit of comment. The TODO about conditional checks is related to some of the discussion below, so will defer to then.

ascillitoe: Ah sorry spoke to soon, the TODO is referring to the docstring for [alibi_detect.models.tensorflow.embedding](https://docs.seldon.io/projects/alibi-detect/en/latest/api/alibi_detect.models.tensorflow.embedding.html). The expected type for the `layers` kwarg is conditional on the embedding `type`. Not 100% sure if we can even do conditional things like this in pydantic models, but added a TODO as something to look into.

alibi_detect/utils/schemas.py
ascillitoe: No reason. I was having trouble with `Optional[Union[...]]` previously, but just tried again and it works fine now. Will amend!

alibi_detect/utils/schemas.py
ascillitoe: The latter.

alibi_detect/utils/loading.py
ascillitoe: Have changed the error to this:
```
raise ValueError(f'Loading the {detector_name} detector from a config.toml is not yet supported.')
```

so at least typo's will be a little more obvious to the user...

alibi_detect/utils/schemas.py
jklaise: Let's open a new issue to update these to `Literal`s.

ascillitoe: https://github.com/SeldonIO/alibi-detect/issues/463

alibi_detect/utils/schemas.py
ascillitoe: Replaced this with `# type: ignore[assignment]`. Not sure if there's a better approach...

ascillitoe:
@jklaise I've resolved a few of you comments, feel free to unresolve if you disagree. Am still looking into a few of the outstanding issues to see if anything can be improved.

mauicv (commented):
LGTM

alibi_detect/utils/loading.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # check version
+    version = cfg.pop('version', None)
+    if version is not None and version != __version__:
```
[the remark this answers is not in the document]
ascillitoe: Good idea IMO.

alibi_detect/utils/schemas.py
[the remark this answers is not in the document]
mauicv: I was going to suggest:

```py
reduce(lambda a, b: Union[a, b], SupportedModels)
```

but your way looks like the right way to do it!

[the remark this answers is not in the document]
ascillitoe: Ooh very pythonic. Shall keep this one in mind in case the current solution doesn't work in part III. Thanks!

alibi_detect/utils/loading.py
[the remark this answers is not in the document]
ascillitoe: @mauicv have some thoughts on this now, will get back to you tmrow!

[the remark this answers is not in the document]
ascillitoe: @jklaise @mauicv 
I'm now rethinking if we even need `config_spec`. We can't update `config_spec` without updating `version`, so the only valid use case to have both is if we wish to update `version` but not `config_spec`. This would make sense if we wished to gurantee config functionality across `version`'s, but not `config_spec`'s. However, I'm not sure we do want to do that, as save/loading across `version`'s is just as likely to cause problems. 

Another way to put it is, since the config is so tightly coupled with our API, do we actually need both `config_spec` and `version` number?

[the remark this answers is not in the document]
jklaise: Good question. I thought initially it would make sense as we wouldn't change `config_spec` every version and it would be nice to be able to have compatibility between "adjacent" versions. But would we be shooting ourselves in the foot by doing it? Are there examples where config wouldn't change but the saving/loading could be broken due to some other stuff changing?

A related rather gnarly issue I just thought of is artefacts of dependencies. I.e. we are compatible with many `tensorflow` or `torch` versions, but our config doesn't know which version the artefacts were saved with, so it's possible that the library version is the same but due to a change in dependency version some artefacts are broken. I guess this should align more with the "version your environment" philosophy though, not saying necessarily we should track the versions of the dependency artefacts as you can take it arbitrarily far (e.g. saving exact Python version too).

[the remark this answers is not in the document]
mauicv: Maybe it makes sense to only have the `config_spec` and have `version` only relate to the `api`. If a version update to the `api` also corresponds to breaking changes to the save load functionality then we update the `config_spec`? We're only really looking to track breaking changes right? So we bump the `config_spec` when we want to notify users they have to update there config files to work with later versions of the `api`? Although I guess those might just be considered Major changes? hmmm 🤔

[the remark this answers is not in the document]
ascillitoe: Based on offline discussion, we will retain `config_spec`, but only major-minor version numbers. Documentation of the intended use has been added to `alibi_detect/version.py`.

alibi_detect/utils/schemas.py
ascillitoe: Have made a note to discuss again in part III.

PR #474, opened 2022-04-01
Init DeepKernel RBFs inside init()

ojcobb:
For some reason initialising `tf.keras.Model` objects causes all available GPU memory to become occupied. This was happening upon import of `alibi_detect` due to the fact that `GaussianRBF` objects were being initialised in the definition of the `DeepKernel` class. This patch fix now sets the defaults to `None` and only initialises the `GaussianRBF` objects upon instantiation of a `DeepKernel` object.

### DIFF
--- a/alibi_detect/utils/tensorflow/kernels.py
+++ b/alibi_detect/utils/tensorflow/kernels.py
@@ -108,6 +108,7 @@ class DeepKernel(tf.keras.Model):
         The kernel to apply to the projected inputs. Defaults to a Gaussian RBF with trainable bandwidth.
     kernel_b
         The kernel to apply to the raw inputs. Defaults to a Gaussian RBF with trainable bandwidth.
+        Set to None in order to use only the deep component (i.e. eps=0).
     eps
         The proportion (in [0,1]) of weight to assign to the kernel applied to raw inputs. This can be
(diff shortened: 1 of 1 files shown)


alibi_detect/utils/tensorflow/kernels.py
```diff
@@ -117,11 +117,15 @@ class DeepKernel(tf.keras.Model):
...
-        kernel_b: Optional[tf.keras.Model] = GaussianRBF(trainable=True),
+        kernel_a: Optional[tf.keras.Model] = None,
+        kernel_b: Optional[tf.keras.Model] = None,
         eps: Union[float, str] = 'trainable'
     ) -> None:
         super().__init__()
+        if kernel_a is None:
+            kernel_a = GaussianRBF(trainable=True)
+        if kernel_b is None:
+            kernel_b = GaussianRBF(trainable=True)
```
jklaise, line 128: Would this new logic contradict the existing docstring:
```
    kernel_b
        The kernel to apply to the raw inputs. Defaults to a Gaussian RBF with trainable bandwidth.
        Set to None in order to use only the deep component (i.e. eps=0).
```

jklaise (approved):

PR #475, opened 2022-04-01
Add support for linear-time mmd estimator.

### LINKED ISSUE #288: Integrate linear time MMD detector

Srceh:
This PR implements the linear-time estimator in (Lemma14 in [paper](https://www.jmlr.org/papers/volume13/gretton12a/gretton12a.pdf)), as asked in #288.


Srceh:
> Do we think users are ever going to have equal reference and test batch sizes in practice? I'd guess almost always the reference set is going to be much larger. I wonder if we'd be better off using the B-stat estimator by default for the linear case rather than Gretton's estimator for equal sample sizes. This additionally has the advantage of a tunable parameter that allows for interpolation between a linear and quadratic time estimator.
> 
> @arnaudvl @Srceh
> 
> Edit: It's actually not quite this simple. However I think we should put some thought into how best to address the n!=m case.

Agree, maybe we can leave the current PR as it is for the linear-time one, and do a separate one for the additional B-stat implementation.

arnaudvl:
@Srceh @ojcobb Thinking if it wouldn't be cleaner to have a separate `LinearMMDDrift` and `BMMDDrift` (for lack of a better name) instead of grouping everything in the existing MMD implementation. It would be a bit easier to debug as well and can just share the MMD base class.

Srceh:
Now the linear-time estimator also uses Gaussian under null for the test threshold, so no permutation is required. It should be the fastest at the cost of lower test power and some unused samples.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -68,16 +68,25 @@ def __init__(
     def sigma(self) -> torch.Tensor:
         return self.log_sigma.exp()
 
-    def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch.Tensor],
-                infer_sigma: bool = False) -> torch.Tensor:
+    def forward(self, x: Union[np.ndarray, torch.Tensor],
+                y: Union[np.ndarray, torch.Tensor],
+                infer_sigma: bool = False,
+                diag: bool = False) -> torch.Tensor:
```
ojcobb, line 74: Given that they refer to the same thing, perhaps we could keep consistency between this kwarg name and the naming convention adopted for the squared distance functions? So perhaps `pairwise: bool = True`?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -68,16 +68,25 @@ def __init__(
...
+        if not diag:
+            dist = distance.squared_pairwise_distance(x.flatten(1), y.flatten(1))  # [Nx, Ny]
+        else:
+            dist = distance.squared_distance(x.flatten(1), y.flatten(1))  # [N, 1]
 
         if infer_sigma or self.init_required:
             if self.trainable and infer_sigma:
                 raise ValueError("Gradients cannot be computed w.r.t. an inferred sigma value")
-            sigma = self.init_sigma_fn(x, y, dist)
+            if not diag:
```
ojcobb, line 85: Is this a good default behaviour to have? Could we end up with O(n^2) costs in places where the linear time estimator is being used specifically because such a cost would be infeasible?

alibi_detect/utils/tensorflow/kernels.py
```diff
@@ -69,15 +69,24 @@ def __init__(
     def sigma(self) -> tf.Tensor:
         return tf.math.exp(self.log_sigma)
 
-    def call(self, x: tf.Tensor, y: tf.Tensor, infer_sigma: bool = False) -> tf.Tensor:
+    def call(self, x: tf.Tensor, y: tf.Tensor,
```
ojcobb, line 72: See comments on pytorch version

alibi_detect/utils/pytorch/distance.py
```diff
@@ -93,7 +115,43 @@ def batch_compute_kernel_matrix(
     return k_mat
 
 
-def mmd2_from_kernel_matrix(kernel_mat: torch.Tensor, m: int, permute: bool = False,
+def linear_mmd2(x: torch.Tensor,
```
ojcobb, line 118: Nitpick but probs worth keeping indentation within function definitions consistent with all of the other functions.

alibi_detect/utils/tensorflow/distance.py
```diff
@@ -83,7 +109,52 @@ def batch_compute_kernel_matrix(
...
+        Kernel function.
+    permute
+        Whether to permute the row indices. Used for permutation tests.
+    """
+    n = np.shape(x)[0]
+    m = np.shape(y)[0]
+    if n != m:
+        raise RuntimeError("Linear-time estimator requires equal size samples")
+    if not permute:
+        k_xx = kernel(x=x[0::2, :], y=x[1::2, :], diag=True)
```
ojcobb, line 136: Would this not fail for an odd value of n (and m)? `[0::2, :]`-based subsets would contain extra elements to `[1::2, :]`-based subsets.

alibi_detect/cd/mmd.py
```diff
@@ -18,6 +18,7 @@ def __init__(
             x_ref: Union[np.ndarray, list],
             backend: str = 'tensorflow',
             p_val: float = .05,
+            estimator: str = 'quad',
```
ojcobb, line 21: Would `estimator_complexity` be more descriptive? (Or at least make clear in the docstring)

alibi_detect/utils/pytorch/distance.py
```diff
@@ -93,7 +115,43 @@ def batch_compute_kernel_matrix(
...
+    if n != m:
+        raise RuntimeError("Linear-time estimator requires equal size samples")
+    k_xx = kernel(x=x[0::2, :], y=x[1::2, :], diag=True)
+    k_yy = kernel(x=y[0::2, :], y=y[1::2, :], diag=True)
+    k_xy = kernel(x=x[0::2, :], y=y[1::2, :], diag=True)
+    k_yz = kernel(x=y[0::2, :], y=x[1::2, :], diag=True)
+
+    h = k_xx + k_yy - k_xy - k_yz
+    mmd2 = h.sum() / (n / 2.)
+    var_mmd2 = ((h * h).sum() / (n / 2.)) - (mmd2 ** 2)
```
ojcobb, line 148: Should this be sample variance rather than population variance? (i.e. with 1 subtracted from (n/2.))

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,143 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+        n_hat = int(np.floor(min(n, m) / 2) * 2)
+        x_ref = torch.from_numpy(x_ref[:n_hat, :]).to(self.device)  # type: ignore[assignment]
+        x = torch.from_numpy(x[:n_hat, :]).to(self.device)  # type: ignore[assignment]
+        mmd2, var_mmd2 = linear_mmd2(x_ref, x, self.kernel)  # type: ignore[arg-type]
+        if self.device.type == 'cuda':
+            mmd2 = mmd2.cpu()
+        mmd2 = mmd2.numpy().item()
+        var_mmd2 = var_mmd2.numpy().item()
+        std_mmd2 = np.sqrt(var_mmd2)
+        p_val = 1 - stats.norm.cdf(mmd2 * np.sqrt(n_hat), loc=0., scale=std_mmd2*np.sqrt(2))
```
ojcobb, line 259: Nitpick but should this be a t-test?

alibi_detect/utils/pytorch/distance.py
```diff
@@ -93,7 +115,43 @@ def batch_compute_kernel_matrix(
...
+    m = np.shape(y)[0]
+    if n != m:
+        raise RuntimeError("Linear-time estimator requires equal size samples")
+    k_xx = kernel(x=x[0::2, :], y=x[1::2, :], diag=True)
+    k_yy = kernel(x=y[0::2, :], y=y[1::2, :], diag=True)
+    k_xy = kernel(x=x[0::2, :], y=y[1::2, :], diag=True)
+    k_yz = kernel(x=y[0::2, :], y=x[1::2, :], diag=True)
+
+    h = k_xx + k_yy - k_xy - k_yz
+    mmd2 = h.sum() / (n / 2.)
```
ojcobb, line 147: Is there a reason we don't just use `h.mean()` and `h.var()`?

alibi_detect/utils/tensorflow/distance.py
```diff
@@ -83,7 +109,52 @@ def batch_compute_kernel_matrix(
...
+        k_yy = kernel(x=y[0::2, :], y=y[1::2, :], diag=True)
+        k_xy = kernel(x=x[0::2, :], y=y[1::2, :], diag=True)
+        k_yz = kernel(x=y[0::2, :], y=x[1::2, :], diag=True)
+    else:
+        idx = np.random.permutation(m + n)
+        xy = tf.gather(tf.concat([x, y], axis=0), idx)
+        x_hat, y_hat = xy[:n, :], xy[n:, :]
+        k_xx = kernel(x_hat[0::2, :], x_hat[1::2, :], diag=True)
```
ojcobb, line 144: Seems like unnecessary duplication

alibi_detect/utils/tensorflow/distance.py
```diff
@@ -83,7 +109,52 @@ def batch_compute_kernel_matrix(
     return k_mat
 
 
-def mmd2_from_kernel_matrix(kernel_mat: tf.Tensor, m: int, permute: bool = False,
+def linear_mmd2(x: tf.Tensor,
+                y: tf.Tensor,
+                kernel: Callable,
+                permute: bool = False) -> Tuple[tf.Tensor, tf.Tensor]:
```
ojcobb, line 115: Is there a reason we offer permute option for tensorflow and not torch?

ojcobb, line 115: Is there a reason we offer permute option for tensorflow and not torch?

alibi_detect/utils/tensorflow/distance.py
```diff
@@ -83,7 +109,52 @@ def batch_compute_kernel_matrix(
...
+        k_xy = kernel(x=x[0::2, :], y=y[1::2, :], diag=True)
+        k_yz = kernel(x=y[0::2, :], y=x[1::2, :], diag=True)
+    else:
+        idx = np.random.permutation(m + n)
+        xy = tf.gather(tf.concat([x, y], axis=0), idx)
+        x_hat, y_hat = xy[:n, :], xy[n:, :]
+        k_xx = kernel(x_hat[0::2, :], x_hat[1::2, :], diag=True)
+        k_yy = kernel(y_hat[0::2, :], y_hat[1::2, :], diag=True)
+        k_xy = kernel(x_hat[0::2, :], y_hat[1::2, :], diag=True)
+        k_yz = kernel(y_hat[0::2, :], x_hat[1::2, :], diag=True)
```
ojcobb, line 147: Seems like unnecessary duplication

alibi_detect/utils/pytorch/distance.py
ojcobb, line 148: Nitpick but should this be sample variance rather than population variance? (i.e. with 1 subtracted from (n/2.))

alibi_detect/utils/pytorch/kernels.py
Srceh: Indeed, now uses `pairwise` as suggested.

alibi_detect/utils/pytorch/kernels.py
Srceh: Now directly use the median of the non-pairwise distance.

alibi_detect/utils/pytorch/distance.py
Srceh: Fixed, should be consistent all across now.

alibi_detect/cd/mmd.py
Srceh: Added extra description in the docstring.

alibi_detect/utils/pytorch/distance.py
Srceh: Now uses `h.mean()`, and `torch.var(, unbiased=True)` in the torch version. TF version uses `tf.reduce_mean` and manual correction.

alibi_detect/utils/tensorflow/distance.py
Srceh: Legacy issue, now removed for the tensorflow version.

alibi_detect/utils/tensorflow/distance.py
Srceh: Removed.

alibi_detect/cd/pytorch/mmd.py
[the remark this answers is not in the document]
Srceh: Nice spot, now fixed with t-test for both versions.

alibi_detect/utils/pytorch/distance.py
Srceh: Yes it should be, now both fixed with the unbiased estimation.

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+
+        Parameters
+        ----------
+        x
+            Batch of instances.
+
+        Returns
+        -------
+        p-value obtained from the permutation test, the MMD^2 between the reference and test set
+        and the MMD^2 values from the permutation test.
```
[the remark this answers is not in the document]
Srceh: Fixed, thanks for spotting.

alibi_detect/cd/base.py
```diff
@@ -570,12 +570,16 @@ def predict(self, x: Union[np.ndarray, list], return_p_val: bool = True, return_
...
-        p_val, dist, dist_permutations = self.score(x)
-        drift_pred = int(p_val < self.p_val)
+        p_val, dist, tmp_v = self.score(x)
+        if len(np.shape(tmp_v)) > 0:
+            dist_permutations = tmp_v
+            # compute distance threshold
+            idx_threshold = int(self.p_val * len(dist_permutations))
+            distance_threshold = np.sort(dist_permutations)[::-1][idx_threshold]
+        else:
+            distance_threshold = tmp_v
```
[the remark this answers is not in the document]
Srceh: Yes indeed, guess the best thing to do here is to follow your draft PR's template to modify the linear-time detector.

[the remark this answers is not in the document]
ascillitoe: I'd say if @arnaudvl and @ojcobb agree with the change in #489 , we should merge that, and then rebase this PR onto the latest `master`.

ascillitoe:
@Srceh final thing, were you planning on adding tests and updating the docs?

Srceh:
> @Srceh final thing, were you planning on adding tests and updating the docs?

Good question, haven't worked on anything specific, but happy to follow any suggestion.

ascillitoe:
> Good question, haven't worked on anything specific, but happy to follow any suggestion.

We'd usually aim to ensure any new functionality added to the library is documented and tested. Since these new detectors come under `MMDDrift`, I think its just a very simple case of making sure the new `estimator` kwarg is documented in `doc/source/methods/mmddrift.ipynb`  🙂 

Could also think about tweaking an existing example to show off (and explain) the linear time MMD if you think if its sufficiently exciting! (this is probs one for @arnaudvl and @ojcobb to weigh in on). I think we mentioned looking at adding some discussion of complexity to the docs at some point, so this could be something to leave until then...

Regarding tests, IMO it might be worth testing the estimators inside `alibi_detect/cd/tensorflow/tests/test_mmd_tf.py` and `alibi_detect/cd/pytorch/tests/test_mmd_pt.py`.

Srceh:
> > Good question, haven't worked on anything specific, but happy to follow any suggestion.
> 
> We'd usually aim to ensure any new functionality added to the library is documented and tested. Since these new detectors come under `MMDDrift`, I think its just a very simple case of making sure the new `estimator` kwarg is documented in `doc/source/methods/mmddrift.ipynb` 🙂
> 
> Could also think about tweaking an existing example to show off (and explain) the linear time MMD if you think if its sufficiently exciting! (this is probs one for @arnaudvl and @ojcobb to weigh in on). I think we mentioned looking at adding some discussion of complexity to the docs at some point, so this could be something to leave until then...
> 
> Regarding tests, IMO it might be worth testing the estimators inside `alibi_detect/cd/tensorflow/tests/test_mmd_tf.py` and `alibi_detect/cd/pytorch/tests/test_mmd_pt.py`.

Great thanks! will follow up on these.

arnaudvl:
> > > Good question, haven't worked on anything specific, but happy to follow any suggestion.
> > 
> > 
> > We'd usually aim to ensure any new functionality added to the library is documented and tested. Since these new detectors come under `MMDDrift`, I think its just a very simple case of making sure the new `estimator` kwarg is documented in `doc/source/methods/mmddrift.ipynb` slightly_smiling_face
> > Could also think about tweaking an existing example to show off (and explain) the linear time MMD if you think if its sufficiently exciting! (this is probs one for @arnaudvl and @ojcobb to weigh in on). I think we mentioned looking at adding some discussion of complexity to the docs at some point, so this could be something to leave until then...
> > Regarding tests, IMO it might be worth testing the estimators inside `alibi_detect/cd/tensorflow/tests/test_mmd_tf.py` and `alibi_detect/cd/pytorch/tests/test_mmd_pt.py`.
> 
> Great thanks! will follow up on these.

- Yes, tests should always be included for new methods. Can hopefully copy a lot over from existing (e.g. MMD) test and test the linear-time specific changes.
- Same for docs. The method doc can again be quite similar to the MMD one. Similar to online version of MMD.
- For the example, I think it is sufficient to just update an existing example with e.g. 100 runs of the usual MMD detector vs. the linear time one, and show timing differences between those runs (print in cell) but also show that power is likely to go down.

arnaudvl:
> > * The method doc can again be quite similar to the MMD one. Similar to online version of MMD.
> 
> Not actually sure we need a new method docs page, since the new detectors are dispatched via the existing `MMDDrift`? Just need to document the `estimator` kwarg in the existing `MMDDrift` doc I think? Could perhaps also add a sentence/paragraph detailing algorithm differences and considerations in choosing one or the other?

Ok yes makes sense.

alibi_detect/cd/mmd.py
```diff
@@ -4,10 +4,10 @@
 from alibi_detect.utils.frameworks import has_pytorch, has_tensorflow
 
 if has_pytorch:
-    from alibi_detect.cd.pytorch.mmd import MMDDriftTorch
+    from alibi_detect.cd.pytorch.mmd import MMDDriftTorch, LinearTimeDriftTorch
```
arnaudvl, line 7: Change to `LinearTimeMMDDriftTorch` for consistency with TensorFlow.

alibi_detect/cd/mmd.py
```diff
@@ -40,6 +41,11 @@ def __init__(
             Backend used for the MMD implementation.
         p_val
             p-value used for the significance of the permutation test.
+        estimator
+            Estimator used for the MMD^2 computation {'quad', 'linear'}. 'Quad' is the default and
+            uses the quadratic u-statistics on each square kernel matrix. 'Linear' uses the linear
+            time estimator as in Gretton et al. (2014), and the threshold is computed using the Gaussian
```
arnaudvl, line 47: Would be helpful to add a link to the paper + reference to the specific section in the paper.

alibi_detect/cd/mmd.py
```diff
@@ -88,9 +94,21 @@ def __init__(
 
         if backend == 'tensorflow' and has_tensorflow:
             kwargs.pop('device', None)
-            self._detector = MMDDriftTF(*args, **kwargs)  # type: ignore
+            if estimator == 'quad':
+                self._detector = MMDDriftTF(*args, **kwargs)  # type: ignore
+            elif estimator == 'linear':
+                kwargs.pop('n_permutations', None)
```
arnaudvl, line 100: Best to clarify in the docstrings that `n_permutations` is not used for the *linear* estimator.

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+            [mmd2_from_kernel_matrix(kernel_mat, n, permute=True, zero_diag=False)
+             for _ in range(self.n_permutations)]
+            )
         if self.device.type == 'cuda':
             mmd2, mmd2_permuted = mmd2.cpu(), mmd2_permuted.cpu()
         p_val = (mmd2 <= mmd2_permuted).float().mean()
         return p_val.numpy().item(), mmd2.numpy().item(), mmd2_permuted.numpy()
+
+
+class LinearTimeDriftTorch(BaseMMDDrift):
```
arnaudvl, line 139: As mentioned before, should probably be `LinearTimeMMDDriftTorch` (what a name...).

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+            kernel: Callable = GaussianRBF,
+            sigma: Optional[np.ndarray] = None,
+            configure_kernel_from_x_ref: bool = True,
+            n_permutations: int = 100,
+            device: Optional[str] = None,
+            input_shape: Optional[tuple] = None,
+            data_type: Optional[str] = None
+    ) -> None:
+        """
+        Maximum Mean Discrepancy (MMD) data drift detector using a permutation test, with linear-time estimator.
```
arnaudvl, line 156: It doesn't use a permutation test though?

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+            self,
+            x_ref: Union[np.ndarray, list],
+            p_val: float = .05,
+            preprocess_x_ref: bool = True,
+            update_x_ref: Optional[Dict[str, int]] = None,
+            preprocess_fn: Optional[Callable] = None,
+            kernel: Callable = GaussianRBF,
+            sigma: Optional[np.ndarray] = None,
+            configure_kernel_from_x_ref: bool = True,
+            n_permutations: int = 100,
```
arnaudvl, line 150: Should not contain the `n_permutations` kwarg.

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+        """
+        super().__init__(
+            x_ref=x_ref,
+            p_val=p_val,
+            preprocess_x_ref=preprocess_x_ref,
+            update_x_ref=update_x_ref,
+            preprocess_fn=preprocess_fn,
+            sigma=sigma,
+            configure_kernel_from_x_ref=configure_kernel_from_x_ref,
+            n_permutations=n_permutations,
```
arnaudvl, line 197: Does not need to pass `n_permutations` here.

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+        preprocess_fn
+            Function to preprocess the data before computing the data drift metrics.
+        kernel
+            Kernel used for the MMD computation, defaults to Gaussian RBF kernel.
+        sigma
+            Optionally set the GaussianRBF kernel bandwidth. Can also pass multiple bandwidth values as an array.
+            The kernel evaluation is then averaged over those bandwidths.
+        configure_kernel_from_x_ref
+            Whether to already configure the kernel bandwidth from the reference data.
+        n_permutations
```
arnaudvl, line 179: Again remove

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+        """ Compute and return full kernel matrix between arrays x and y. """
+        k_xy = self.kernel(x, y, self.infer_sigma)
+        k_xx = self.k_xx if self.k_xx is not None and self.update_x_ref is None else self.kernel(x, x)
+        k_yy = self.kernel(y, y)
+        kernel_mat = torch.cat([torch.cat([k_xx, k_xy], 1), torch.cat([k_xy.T, k_yy], 1)], 0)
+        return kernel_mat
+
+    def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
+        """
+        Compute the p-value resulting from a permutation test using the maximum mean discrepancy
```
arnaudvl, line 234: Remove reference to permutation test

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+
+        Returns
+        -------
+        p-value obtained from the null hypothesis, the MMD^2 between the reference and test set
+        and the MMD^2 threshold for the given significance level.
+        """
+        x_ref, x = self.preprocess(x)
+        n = x.shape[0]
+        m = x_ref.shape[0]
+        n_hat = int(np.floor(min(n, m) / 2) * 2)
```
arnaudvl, line 250: This behaviour needs to be well documented.

alibi_detect/cd/pytorch/mmd.py
ascillitoe: The names just get better and better 😅

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+        x_ref = torch.from_numpy(x_ref[:n_hat, :]).to(self.device)  # type: ignore[assignment]
```
arnaudvl, line 251: I believe there is a case to be made that there is an explicit check such that `n == m` and if not, an error is raised. The reason is that silently some unexpected behaviour can occur by only selecting the first `n_hat` reference/test instances. If say the reference data is ordered and contains samples from classes 1,2 and 3, then only choosing `:n_hat` could ignore all samples from class 3 and not form an i.i.d. sample anymore. So my preference would be explicit behaviour around this (raising errors) or if we allow this (which I am not in favour of now) then randomly sample `n_hat` instances from `x_ref` and `x`. Good to have some opinions @jklaise @ascillitoe @ojcobb

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+
+        # compute kernel matrix for the reference data
+        if self.infer_sigma or isinstance(sigma, torch.Tensor):
+            x = torch.from_numpy(self.x_ref).to(self.device)
+            self.k_xx = self.kernel(x, x, infer_sigma=self.infer_sigma)
+            self.infer_sigma = False
+        else:
+            self.k_xx, self.infer_sigma = None, True
+
+    def kernel_matrix(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
```
arnaudvl, line 224: Method is not used I believe?

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+
+        # initialize kernel
+        sigma = torch.from_numpy(sigma).to(self.device) if isinstance(sigma,  # type: ignore[assignment]
+                                                                      np.ndarray) else None
+        self.kernel = kernel(sigma) if kernel == GaussianRBF else kernel
+
+        # compute kernel matrix for the reference data
+        if self.infer_sigma or isinstance(sigma, torch.Tensor):
+            x = torch.from_numpy(self.x_ref).to(self.device)
+            self.k_xx = self.kernel(x, x, infer_sigma=self.infer_sigma)
```
arnaudvl, line 219: `self.k_xx` is not used?

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -118,17 +119,144 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+        m = x_ref.shape[0]
+        n_hat = int(np.floor(min(n, m) / 2) * 2)
+        x_ref = torch.from_numpy(x_ref[:n_hat, :]).to(self.device)  # type: ignore[assignment]
+        x = torch.from_numpy(x[:n_hat, :]).to(self.device)  # type: ignore[assignment]
+        mmd2, var_mmd2 = linear_mmd2(x_ref, x, self.kernel)  # type: ignore[arg-type]
+        if self.device.type == 'cuda':
+            mmd2 = mmd2.cpu()
+        mmd2 = mmd2.numpy().item()
+        var_mmd2 = var_mmd2.numpy().item()
+        std_mmd2 = np.sqrt(var_mmd2)
```
arnaudvl, line 258: Can directly use `torch.std(...)` in `linear_mmd2`? This would remove the few additional lines of code here.

alibi_detect/utils/pytorch/distance.py
```diff
@@ -30,6 +30,28 @@ def squared_pairwise_distance(x: torch.Tensor, y: torch.Tensor, a_min: float = 1
     return dist.clamp_min_(a_min)
 
 
+def squared_distance(x: torch.Tensor, y: torch.Tensor, a_min: float = 1e-30) -> torch.Tensor:
```
arnaudvl, line 33: Can we just apply a reduction to the `squared_pairwise_distance` instead of using an extra function?

alibi_detect/utils/pytorch/distance.py
```diff
@@ -93,8 +115,48 @@ def batch_compute_kernel_matrix(
...
+        Kernel function.
+    Returns
+    -------
+    MMD^2 between the samples.
+    """
+    n = np.shape(x)[0]
+    m = np.shape(y)[0]
+    if n != m:
+        raise RuntimeError("Linear-time estimator requires equal size samples")
+    k_xx = kernel(x=x[0::2, :], y=x[1::2, :], pairwise=False)
```
arnaudvl, line 143: We should be able to do this at init time (so `self.k_xx` becomes useful again).

alibi_detect/utils/pytorch/distance.py
```diff
@@ -93,8 +115,48 @@ def batch_compute_kernel_matrix(
...
+    k_yy = kernel(x=y[0::2, :], y=y[1::2, :], pairwise=False)
+    k_xy = kernel(x=x[0::2, :], y=y[1::2, :], pairwise=False)
+    k_yz = kernel(x=y[0::2, :], y=x[1::2, :], pairwise=False)
```
arnaudvl, line 146: Is `k_yz` the paper notation? B/c it might be easier to follow by just calling it `k_yx`.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -68,16 +68,24 @@ def __init__(
     def sigma(self) -> torch.Tensor:
         return self.log_sigma.exp()
 
-    def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch.Tensor],
-                infer_sigma: bool = False) -> torch.Tensor:
+    def forward(self, x: Union[np.ndarray, torch.Tensor],
```
arnaudvl, line 71: Nitpicking big time here, but let's try to keep same type of indentation as e.g. in the `DeepKernel` below.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -68,16 +68,24 @@ def __init__(
...
+                y: Union[np.ndarray, torch.Tensor],
+                infer_sigma: bool = False,
+                pairwise: bool = True) -> torch.Tensor:
 
         x, y = torch.as_tensor(x), torch.as_tensor(y)
-        dist = distance.squared_pairwise_distance(x.flatten(1), y.flatten(1))  # [Nx, Ny]
+        if pairwise:
```
arnaudvl, line 77: Check my comment in `distance.py` which might make this  `if else` redundant and reduce it to a kwarg of the distance function.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -68,16 +68,24 @@ def __init__(
...
+            dist = distance.squared_distance(x.flatten(1), y.flatten(1))  # [N, 1]
 
         if infer_sigma or self.init_required:
             if self.trainable and infer_sigma:
                 raise ValueError("Gradients cannot be computed w.r.t. an inferred sigma value")
-            sigma = self.init_sigma_fn(x, y, dist)
+            if pairwise:
+                sigma = self.init_sigma_fn(x, y, dist)
+            else:
+                sigma = (.5 * dist.flatten().sort().values[dist.shape[0] // 2 - 1].unsqueeze(dim=-1)) ** .5
```
arnaudvl, line 88: Again I think we can avoid the hard-coding of this behaviour and fall back on `self.init_sigma_fn` but with the desired linear detector behaviour.

alibi_detect/cd/pytorch/mmd.py
ascillitoe: This seems like quite an issue atm. Agree the safest option would be to explicitly check for `n == m`. Otherwise, we could check, and randomly subsample if `n != m`, with a warning raised to inform the user we are doing this.

ascillitoe (changes requested):
"Requesting changes" to ensure we do not merge until #489 has been merged and `predict` updated in this PR.

alibi_detect/cd/pytorch/mmd.py
Srceh: @arnaudvl The base class requires this method for initialisation, I wonder what would be the preferable solution here? the minimal thing could be to simply leave a pseudo method.

alibi_detect/utils/tensorflow/distance.py
```diff
@@ -20,14 +26,44 @@ def squared_pairwise_distance(x: tf.Tensor, y: tf.Tensor, a_min: float = 1e-30,
         Lower bound to clip distance values.
     a_max
         Upper bound to clip distance values.
-
+    pairwise
```
ascillitoe, line 29: Isn't it a bit unclear to have a function named `squared_pairwise_distance` that optionally computes non-pairwise distances? Perhaps `squared_pairwise_distance` should be renamed `squared_distance`? Or the `pairwise=False` functionality separated out into a separate distance function?

Srceh: I was thinking the same.  It was previously a separate function and @arnaudvl was suggesting making the repeated parts minimal. Guess changing the function name across all related methods would be preferable.

alibi_detect/cd/mmd.py
Srceh: Fixed.

alibi_detect/cd/mmd.py
Srceh: Fixed.

alibi_detect/cd/mmd.py
Srceh: Fixed.

alibi_detect/cd/pytorch/mmd.py
Srceh: Fixed.

alibi_detect/cd/pytorch/mmd.py
Srceh: No it doesn't, removed now.

alibi_detect/cd/pytorch/mmd.py
Srceh: Fixed.

alibi_detect/cd/pytorch/mmd.py
Srceh: Fixed.

alibi_detect/cd/pytorch/mmd.py
Srceh: Fixed.

alibi_detect/cd/pytorch/mmd.py
Srceh: Fixed.

alibi_detect/cd/pytorch/mmd.py
Srceh: Yes, now described in the docstring.

alibi_detect/cd/pytorch/mmd.py
Srceh: Currently implemented as raise error for `n!=m`. Guess the subsampling should be implemented on a stand-alone part, so that it can be used with other detectors?

alibi_detect/cd/pytorch/mmd.py
Srceh: Fixed, not `self.k_xx` is reused for later calculation.

alibi_detect/cd/pytorch/mmd.py
Srceh: The new version uses `np.sqrt(np.clip(var_mmd2, 1e-8, 1e-8))` for numeric stability.

alibi_detect/utils/pytorch/distance.py
Srceh: Now implemented as a single function.

alibi_detect/utils/pytorch/distance.py
Srceh: Fixed, now the kernel matrix is reused for prediction.

alibi_detect/utils/pytorch/distance.py
```diff
@@ -93,8 +115,48 @@ def batch_compute_kernel_matrix(
...
+        Batch of instances of shape [Ny, features].
+    kernel
+        Kernel function.
+    Returns
+    -------
+    MMD^2 between the samples.
+    """
+    n = np.shape(x)[0]
+    m = np.shape(y)[0]
+    if n != m:
```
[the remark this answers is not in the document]
Srceh: Fixed.

alibi_detect/utils/pytorch/distance.py
Srceh: Typo, thanks for noticing, fixed.

alibi_detect/utils/pytorch/kernels.py
Srceh: Fixed, now as part of the `squared_pairwise_distance` function argument.

alibi_detect/utils/pytorch/kernels.py
Srceh: Slightly tricky as the default `init_sigma_fn` is used by other detectors. Might be easier to keep the additional line here?

Srceh:
> Left a number of comments related to the PyTorch implementation. Let's work through those first and then we can apply the desired changes to TensorFlow as well.

TF version is also fixed for the above ones replied with "fixed".

ascillitoe:
@Srceh I've just merged the `score`/`predict` refactoring (#489). This will have introduced a few conflicts you need to resolve. It should simplify your life wrt to the implementation in this PR though!

Srceh:
> @Srceh I've just merged the `score`/`predict` refactoring (#489). This will have introduced a few conflicts you need to resolve. It should simplify your life wrt to the implementation in this PR though!

Nice! will start working on that!

ascillitoe:
@Srceh I have now merged in the v0.10.0 related changes from `master`. This primarily involved changes to the kwargs related to preprocessing, tweaking some tests, and adding your `estimator` kwarg to the `MMDDrift` pydantic models (see `saving/schemas.py`).

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -123,21 +124,162 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, float]:
...
         if self.device.type == 'cuda':
             mmd2, mmd2_permuted = mmd2.cpu(), mmd2_permuted.cpu()
         p_val = (mmd2 <= mmd2_permuted).float().mean()
         # compute distance threshold
         idx_threshold = int(self.p_val * len(mmd2_permuted))
         distance_threshold = torch.sort(mmd2_permuted, descending=True).values[idx_threshold]
         return p_val.numpy().item(), mmd2.numpy().item(), distance_threshold.numpy()
+
+
+class LinearTimeMMDDriftTorch(BaseMMDDrift):
```
ascillitoe, line 149: Since these new subclasses don't make use of `self.n_permutations` (set in `BaseMMDDrift`), shall we set this to `None`? I had a moment of confusion when updating the tests since `self.n_permuations == 100` when `estimator == 'linear'`.

alibi_detect/cd/mmd.py
```diff
@@ -100,9 +108,21 @@ def __init__(
...
+                self._detector = LinearTimeMMDDriftTF(*args, **kwargs)  # type: ignore
```
ascillitoe, line 115: Since the logic to set `self._detector` is located here, we should add additional tests to `alibi_detect/cd/tests/test_mmd.py` to check that the correct subclass is selected conditional on `backend` and `estimator`.

alibi_detect/cd/pytorch/mmd.py
Srceh: Good point. The default number of permutations then can be initialised in `/cd/mmd.py` when estimator is `'quad'`.

alibi_detect/cd/mmd.py
Srceh: Indeed, will modify the tests.

Srceh: Simply rewrite the test to go through different `backend` and `estimator` options, should do the job.

PR #479, opened 2022-04-06
Replace all occurences of "en/latest" with "en/stable".

jklaise:
This is to ensure that our docs always point to the stable release
instead of latest to avoid potential discrepancy between someone reading
the latest docs but installing a stable version.


jklaise:
To note - this includes retroactive changes to `CHANGELOG.md`. However, the previous `latest` wasn't very accurate either as wouldn't point to the actual release tag.

There are a couple of non-alibi changed, e.g. for `numpy`, `mamba` , `myst` docs. Have checked that these work with the change.

ascillitoe (approved):
I've randomly selected some links to check and they look good. Agree that this is a good change to make.

Issue #453, opened 2022-02-25
Python 3.10 support


jklaise:
The newly released `tensorflow` 2.8 has Python 3.10 support. We can proceed with adding CI tests for 3.10 after merging #444.

PR #485, opened 2022-04-17
Add Python 3.10

ascillitoe:
Addresses #453

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,9 @@ repos:
     hooks:
       - id: flake8
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.812
+    rev: v0.942
     hooks:
       - id: mypy
+        additional_dependencies: [
+          types-requests>=2.25.0,
+        ]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,12 +30,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [ '3.7', '3.8', '3.9' ]
-        include: # Run macos and windows tests on python 3.9 only
+        python-version: [ '3.7', '3.8', '3.9', '3.10' ]
+        include: # Run macos and windows tests on python 3.10 only
           - os: windows-latest 
-            python-version:  '3.9'
+            python-version:  '3.10'
           - os: macos-latest 
-            python-version:  '3.9'
+            python-version:  '3.10'
 
     steps:
       - name: Checkout code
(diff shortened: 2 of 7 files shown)


ascillitoe:
CI failing on Python 3.10 due to issue installing `pystan`. This might be due to the rather old `pystan` version not playing nicely with new Python versions (we are locked in to `pystan<3.0` due to `Prophet` (https://github.com/SeldonIO/alibi-detect/pull/194).

Hopefully, Prophet might become easier to install soon with the move to `cmdstanpy` (see https://github.com/facebook/prophet/issues/2041). In the meantime, we could skip the Prophet tests on Python 3.10.

ascillitoe:
PyTorch is very unhappy on Windows with Python 3.10 (lots of `RuntimeError: Could not infer dtype of numpy.float32`). Requires further investigation.

ascillitoe:
From [this](https://pytorch.org/get-started/locally/#windows-python) PyTorch documentation:

*"Currently, PyTorch on Windows only supports Python 3.7-3.9"*

Hence the above errors are expected. Will move the Windows CI back to 3.9 for now.

ascillitoe:
Moving off draft to run CI.

ascillitoe:
@jklaise  @mauicv  This should be good to go, assuming we don't want to add anything to the docs regarding PyTorch limiting us to Python 3.9 on Windows. (I'm tempted not to add anything since it is yet another thing to keep updated, and Windows usage still appears to be quite an edge case).

.pre-commit-config.yaml
```diff
@@ -4,6 +4,9 @@ repos:
     hooks:
       - id: flake8
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.812
+    rev: v0.942
     hooks:
       - id: mypy
+        additional_dependencies: [
+          types-requests>=2.25.0,
+        ]
```
jklaise, line 12: I thought changing `mypy` to `>0.900` was quite a big change? Surprised it doesn't require more type stub packages... In any case, we should split this out and do a PR for both libraries in sync to avoid incompatible dev environments.

jklaise (approved):
Looks good, just splitting out `mypy` upgrade in a different PR would be preferable.

.pre-commit-config.yaml
ascillitoe: Yeh tbh I was surprised it was this easy. I simply bumped to `>0.900` and then only added the type stub packages necessary to run successfully. The full list of stdlib type stub packages is [here](https://github.com/python/typeshed/tree/master/stubs) I think. I'm surpised we only seem to need `type-requests` (and `type-urllib3` which is a dep of `type-requests`. But it seems to run OK!

Good idea separating out, I didn't think about compatible dev requirements. I'll do them both together!

jklaise: Great, thanks. I think when I looked into bumping for `alibi` it seemed more involved with `numpy` stuff etc. but things might have improved since 0.900 dropped.

ascillitoe: Think you might be right, I started to have a go with `alibi`; the type stubs seemed fine but it did raise some additional `numpy` errors. I'll have another look!

PR #489, opened 2022-04-22
Refactor drift detector score methods

ascillitoe:
This PR moves the `distance_threshold` computation from `.predict()` to `.score()` methods, for all drift detectors that perform permutation tests. 

This simplifies detector `predict` methods, and aids consistency between the existing `MMDDrift` detectors and the new linear-time MMD detectors introduced by #475.

### DIFF
--- a/alibi_detect/cd/tensorflow/mmd.py
+++ b/alibi_detect/cd/tensorflow/mmd.py
@@ -89,7 +89,7 @@ class MMDDriftTF(BaseMMDDrift):
         kernel_mat = tf.concat([tf.concat([k_xx, k_xy], 1), tf.concat([tf.transpose(k_xy, (1, 0)), k_yy], 1)], 0)
         return kernel_mat
 
-    def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
+    def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, float]:
         """
         Compute the p-value resulting from a permutation test using the maximum mean discrepancy
         as a distance measure between the reference data and the data to be tested.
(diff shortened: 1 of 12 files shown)


alibi_detect/cd/tensorflow/mmd.py
```diff
@@ -115,4 +115,8 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
              for _ in range(self.n_permutations)]
         )
         p_val = (mmd2 <= mmd2_permuted).mean()
-        return p_val, mmd2, mmd2_permuted
+        # compute distance threshold
+        idx_threshold = int(self.p_val * len(mmd2_permuted))
+        distance_threshold = np.sort(mmd2_permuted)[::-1][idx_threshold]
```
arnaudvl, line 120: Not too bothered here but the torch version uses `torch.sort`, any reason why tensorflow doesn't use `tf.sort`?

alibi_detect/cd/tensorflow/mmd.py
```diff
@@ -115,4 +115,8 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
+        print(type(p_val), type(mmd2), type(distance_threshold))
```
arnaudvl, line 121: Remove print statements

alibi_detect/cd/tensorflow/mmd.py
ascillitoe: No good reason. The TF version worked fine with just copying across the legacy `np.sort` code from `predict`, whereas the torch version required updating. Agree though we should use the backend-specific versions for both!

arnaudvl (approved):

PR #490, opened 2022-04-25
Update to mypy 0.9

ascillitoe:
Update to mypy 0.900. This is a precursor for #485.

This requires the installation of third-party type stubs, since they are no longer included in mypy (see [here](http://mypy-lang.blogspot.com/2021/06/mypy-0900-released.html)). 

The type stub packages can be installed with `mypy --install-types`, but this PR installs them via `requirements/dev.txt` so that the stub package versions can be bounded for reproducibility.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,9 @@ repos:
     hooks:
       - id: flake8
   - repo: https://github.com/pre-commit/mirrors-mypy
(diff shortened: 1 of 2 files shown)


ascillitoe:
We should wait until https://github.com/SeldonIO/alibi/pull/638 is also approved before merging.

jklaise (approved):

Issue #491, opened 2022-04-25
training autoencoder with batch size 2 causes error


aaronrmm:
### Summary
In alibi-detect 0.9.1
alibi_detect/models/tensorflow/trainer.py: line 77
`x, y = data if len(data) == 2 else (data, None)`  where data is a minibatch, is an incorrect assumption that causes an error if my minibatch has size 2 (either because my batch size is 2 or because I have an remainder mini-batch of 2. I found this using a batch size of 16 and a training set of size 18).

e.g.:
```
od.fit(
    X_train,
    loss_fn=tf.keras.losses.mse,
    epochs=5,
    verbose=True,
    batch_size=2,
)
```

### To reproduce
: run od_vae_adult.ipynb, adding `batch_size=2` to the training params as in above. Alternatively the below example is a concise script version of that notebook:
```
import alibi.datasets
import tensorflow as tf
from alibi_detect.od import OutlierVAE
from tensorflow.keras import Sequential
from tensorflow.keras.activations import relu
from tensorflow.keras.layers import (
    Dense,
    InputLayer,
)

if __name__ == "__main__":

    adult = alibi.datasets.fetch_adult()
    X, y = adult.data, adult.target

    n_train = 25000
    n_valid = 5000
    X_train, y_train = X[:n_train, :], y[:n_train]
    n_features = X_train.shape[1]
    latent_dim = 2

    encoder_net = Sequential(
        [
            InputLayer(input_shape=(n_features,)),
            Dense(25, activation=tf.nn.relu),
            Dense(10, activation=tf.nn.relu),
            Dense(5, activation=tf.nn.relu),
        ]
    )

    decoder_net = Sequential(
        [
            InputLayer(input_shape=(latent_dim,)),
            Dense(5, activation=relu),
            Dense(10, activation=relu),
            Dense(25, activation=relu),
            Dense(n_features, activation=None),
        ]
    )

    # initialize outlier detector
    od = OutlierVAE(
        threshold=None,  # threshold for outlier score
        score_type="mse",  # use MSE of reconstruction error for outlier detection
        encoder_net=encoder_net,  # can also pass VAE model instead
        decoder_net=decoder_net,  # of separate encoder and decoder
        latent_dim=latent_dim,
        samples=5,
    )

    # train
    od.fit(
        X_train,
        loss_fn=tf.keras.losses.mse,
        epochs=1,
        verbose=True,
        batch_size=2, # comment out to train
    )
```

[...]

PR #492, opened 2022-04-27
bug fix tensorflow trainer

arnaudvl:
Addresses #491 . Requires that, when an explicit dataset is passed to the TensorFlow trainer, it returns `(x, y)`. This behaviour is the same as what is expected for the PyTorch trainer where we pass a dataloader returning `(x, y)`. Compatible with the usage of the trainer in the outlier, adversarial and drift detectors.

### DIFF
--- a/alibi_detect/models/tensorflow/trainer.py
+++ b/alibi_detect/models/tensorflow/trainer.py
@@ -35,7 +35,7 @@ def trainer(
     y_train
         Training labels.
     dataset
-        Training dataset.
+        Training dataset which returns (x, y).
     optimizer
         Optimizer used for training.
     loss_fn_kwargs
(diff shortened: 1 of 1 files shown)


alibi_detect/models/tensorflow/trainer.py
```diff
@@ -57,6 +57,7 @@ def trainer(
     callbacks
         Callbacks used during training.
     """
+    return_xy = False if not isinstance(dataset, tf.keras.utils.Sequence) and y_train is None else True
```
jklaise, line 60: This can be probably simplified to `... if not dataset and ...` (would ignore `None` and empty `dataset`) ? Similarly could write ` ... and not y_train ...` to use the same pattern. (That being said I see `if not isinstance(...)` is used on the line below too, so happy to keep this to avoid further changes).

jklaise (approved):

PR #496, opened 2022-05-03
Random seed utilities

ascillitoe:
This PR adds utility functions to get and set seeds for the Python, Numpy, TensorFlow and PyTorch random number generators. 

### Motivation

Some detectors such as `MMDDrift` and `LSDDDrift` involve operations using NumPy, TensorFlow and/or PyTorch random number generators. The resulting non-deterministic behavior is a problem for testing, as it prevents us performing checks such as:

```
random.seed(42)
np.random.seed(42)
tf.random.set_seed(42)

detector1 = MMDDrift(X_ref, ...)
preds1 = detector1.predict(x_h0)

detector2 = MMDDrift(X_ref, ...)
preds2 = detector2.predict(x_h0)

assert preds1['data']['p_val'] == preds2['data']['p_val']
```
Setting all random seeds (with `np.random.seed(), tf.random.set_seed()` etc) at the start of the test is insufficient to provide determinism here, since the operations on the first detector will still cycle through the RNG states such that the second detector would see a different RNG state. To ensure determinism we would instead have to set seeds again before the init and prediction of the second detector.

Setting global random seeds in individual `pytest` functions is also undesriable in general, and leads to a difficult-to-maintain and unpredictable test suite.

### Solution

This PR addresses the above through two changes:

1. Introduces the use of [pytest-randomly](https://github.com/pytest-dev/pytest-randomly) for testing. This package manages the random seeds used for each test. The seed is randomly changed for each test, which avoids us being tricked by tests passing due to a "friendly" hardcoded random state. Importantly, the random seed is reported for each test e.g.:

   `Using --randomly-seed=1553614239`

   and a test can then be repeated with a given random seed for debugging purposes:

   `pytest --randomly-seed=1234 alibi_detect -k <name_of_failed_test>`

2. A new submodule `alibi_detect.utils.random` is introduced. This contains functionality to set and get RNG seeds (the relevent seeds for NumPy, TensorFlow, PyTorch etc are all set at once). A context manager `fixed_seed` is also introduced to enable select lines of code to be run with a given seed in an isolated fashion, e.g.:

   ```python
   with fixed_seed(42):
      detector1 = MMDDrift(X_ref, ...)
      preds1 = detector1.predict(x_h0)

   with fixed_seed(42):
      detector2 = MMDDrift(X_ref, ...)
      preds2 = detector2.predict(x_h0)

   assert preds1['data']['p_val'] == preds2['data']['p_val']
   ```
   The `fixed_seed`'s above set all the random seeds to 42 within the `with` block, but they are reset to their original values upon exit of the block.

The `alibi_detect.utils.random` has been made public as it might be quite useful for use in example notebooks, experiments etc.

**Note**: If we decided we didn't want the new utilities public, we could hide it within `tests` instead, however we would not be able to set `set_seeds` as an `entry_point` for `pytest`, meaning `pytest` wouldn't use our custom seed setter internally (our custom one also sets the TF and torch random seeds, whereas the `pytest` `reseed` function only sets Python and Numpy seeds).

### DIFF
--- a/alibi_detect/utils/random.py
+++ b/alibi_detect/utils/random.py
@@ -1,0 +1,65 @@
+from contextlib import contextmanager
+import random
+import numpy as np
+import os
+from alibi_detect.utils.frameworks import has_pytorch, has_tensorflow
+
+if has_tensorflow:
+    import tensorflow as tf
+if has_pytorch:
+    import torch
+
+
+def set_seed(seed):
+    """
+    Sets the Python, NumPy, TensorFlow and PyTorch random seeds (if installed).
+    """
+    os.environ['PYTHONHASHSEED'] = str(seed)
+    random.seed(seed)
+    np.random.seed(seed)
+    if has_tensorflow:
+        tf.random.set_seed(seed)
+    if has_pytorch:
+        torch.manual_seed(seed)
+
+
+def get_seed() -> int:
+    """
+    Gets the seed set by :func:`set_seed`.
+
+    Example
+    -------
+    from alibi_detect.utils import set_seed, get_seed
+    set_seed(42)
+    get_seed()
+    >>> 42
+    """
+    return np.random.get_state()[1][0]  # type: ignore[index]
+
+
+@contextmanager
+def fixed_seed(seed: int):
+    """
+    A context manager to run with a requested random seed (applied to all the RNG's set by
+    :func:`alibi_detect.utils.random.set_seeds`).
+
+    Example
+    -------
+    set_seed(0)
+    with fixed_seed(42):
+        dd = cd.LSDDDrift(X_ref)  # seeds equal 42 here
+        p_val = dd.predict(X_h0)['data']['p_val']
+    # seeds equal 0 here
+
+    Warning
+    -------
+    To ensure random seeds are reset to their original values upon exit of the context manager, it is
+    recommended to use :func:`alibi_detect.utils.random.set_seed` rather than `tf.random.set_seed` etc
+    individually when using this context manager.
+    """
+    orig_seed = get_seed()
+    set_seed(seed)
+    try:
+        yield
+    finally:
+        set_seed(orig_seed)
--- a/alibi_detect/utils/tests/test_random.py
+++ b/alibi_detect/utils/tests/test_random.py
@@ -1,0 +1,29 @@
+from alibi_detect.utils.random import set_seed, get_seed, fixed_seed
+
+
+def test_set_get_seed(seed):
+    """
+    Tests the set_seed and get_seed fuctions.
+    """
+    # Check initial seed within test is the one set by pytest-randomly
+    current_seed = get_seed()
+    assert current_seed == seed
+
+    # Set another seed and check
+    new_seed = seed + 42
+    set_seed(new_seed)
+    current_seed = get_seed()
+    assert current_seed == new_seed
+
+
+def test_fixed_seed(seed):
+    """
+    Tests the fixed_seed context manager.
+    """
+    tmp_seed = seed + 42
+    with fixed_seed(tmp_seed):
+        # Check seed inside of context manager
+        assert get_seed() == tmp_seed
+
+    # Check seeds were reset upon exit of context manager
+    assert get_seed() == seed
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -7,6 +7,7 @@ types-requests>=2.25.0, <2.28.0
 pytest>=5.3.5, <8.0.0
 pytest-cov>=2.6.1, <4.0.0
 pytest-xdist>=1.28.0, <3.0.0 # for distributed testing, currently unused (see setup.cfg)
+pytest-randomly>=3.5.0, <4.0.0
 pytest-timeout>=1.4.2, <3.0.0 # for notebook tests
 jupytext>=1.12.0, <2.0.0 # for notebook tests
 ipykernel>=5.1.0, <6.0.0 # for notebook tests
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,7 @@
+[options.entry_points]
+pytest_randomly.random_seeder =
+    reseed = alibi_detect.utils.random:set_seed
+
 [aliases]
 test=pytest
 
@@ -6,6 +10,7 @@ addopts =
     --tb native
     -W ignore
     --cov=alibi_detect
+    --randomly-dont-reorganize
     #-n auto
     #--forked
 
--- a/alibi_detect/utils/tests/conftest.py
+++ b/alibi_detect/utils/tests/conftest.py
@@ -1,0 +1,9 @@
+import pytest
+
+
+@pytest.fixture
+def seed(pytestconfig):
+    """
+    Returns the random seed set by pytest-randomly.
+    """
+    return pytestconfig.getoption("randomly_seed")


setup.cfg
```diff
@@ -6,6 +10,7 @@ addopts =
     --tb native
     -W ignore
     --cov=alibi_detect
+    --randomly-dont-reorganize
```
jklaise, line 13: add comment that this is to keep test ordering

setup.cfg
```diff
@@ -1,3 +1,7 @@
+[options.entry_points]
+pytest_randomly.random_seeder =
+    reseed = alibi_detect.utils.random:set_seed
```
jklaise, line 3: probably need a comment explaining this

alibi_detect/utils/random.py
```diff
@@ -0,0 +1,65 @@
...
+import os
+from alibi_detect.utils.frameworks import has_pytorch, has_tensorflow
+
+if has_tensorflow:
+    import tensorflow as tf
+if has_pytorch:
+    import torch
+
+
+def set_seed(seed):
```
jklaise, line 13: `seed: int`
Is `None` valid/desirable here (e.g. you can do `np.random.seed(None)` do seed with an unknown state).

alibi_detect/utils/random.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Sets the Python, NumPy, TensorFlow and PyTorch random seeds (if installed).
+    """
+    os.environ['PYTHONHASHSEED'] = str(seed)
+    random.seed(seed)
+    np.random.seed(seed)
```
jklaise, line 19: Not sure if relevant, but what was the deal with using new-style `numpy` random number generators?

alibi_detect/utils/random.py
```diff
@@ -0,0 +1,65 @@
...
+    Gets the seed set by :func:`set_seed`.
+
+    Example
+    -------
+    from alibi_detect.utils import set_seed, get_seed
+    set_seed(42)
+    get_seed()
+    >>> 42
+    """
+    return np.random.get_state()[1][0]  # type: ignore[index]
```
jklaise, line 37: Wonder if good enough to rely on `numpy` only for this but not the other libraries? Possible to get the state/seed from others too?

jklaise: Probably overkill/unnecessary but one could also have a global `alibi` variable which holds the seed set by `set_seed`.

ascillitoe: Just had a look and I could get the `tensorflow` and `torch` seeds with [tf.compat.v1.get_seed](https://www.tensorflow.org/api_docs/python/tf/compat/v1/get_seed) and [get_rng_state](https://pytorch.org/docs/stable/generated/torch.get_rng_state.html#torch.get_rng_state). What do you think we should do in the event of conflict though, just raise a warning saying that `set_seed` must have been called before `get_seed` can be used?

alibi_detect/utils/random.py
ascillitoe: Yes I wanted that, but annoyingly [torch.random.manual_seed](https://pytorch.org/docs/stable/generated/torch.manual_seed.html) doesn't support `None`. I guess we could take `None`, and then generate a random seed ourselves using system time etc, but I didn't want to complicate matters too much for now, especially since `random`, `np.random` and `tf.random` all randomly generate seeds in different ways if `None` is passed.

ascillitoe: On second thoughts, in the event of `None`, we could call `np.random.seed(None)` first, and then get that seed to use to set the others?

alibi_detect/utils/random.py
```diff
@@ -0,0 +1,65 @@
...
+    To ensure random seeds are reset to their original values upon exit of the context manager, it is
+    recommended to use :func:`alibi_detect.utils.random.set_seed` rather than `tf.random.set_seed` etc
+    individually when using this context manager.
+    """
+    orig_seed = get_seed()
+    set_seed(seed)
+    try:
+        yield
+    finally:
+        set_seed(orig_seed)
```
jklaise, line 65: I'm not sure resetting seed (or rng state if that was the intention) is that easy (see also above). E.g. what happens in the following scenario where the different library rng's are out of sync?

`set_seed(42) -> some numpy calls -> set_seed(other)`

alibi_detect/utils/random.py
jklaise: I don't think it's worth complicating at this point, the utility of a "random random seed" is also not clear in a function specifically for setting a known seed.

alibi_detect/utils/random.py
ascillitoe: AFAIK, this shouldn't affect the numpy generators, in that they don't appear to be affected by the `np.random.seed()` at all. A very crude example:

```python
from alibi_detect.utils.random import set_seed, get_seed

rng = np.random.default_rng()
set_seed(0)
a = rng.random()
set_seed(0)
b = rng.random()
assert a == b  # False

rng = np.random.default_rng(0)
a = rng.random()
rng = np.random.default_rng(0)
b = rng.random()
assert a == b  # True
```

I've changed `utils.random` to private for now, since going forward we might want to more deeply consider how we can move from the legacy `numpy` RNG to generators.

ascillitoe:
The FET and CVM detector tests needed to be updated since they were setting `np.random.seed` outside of test function definitions, so the addition of `pytest-randomly` introduced randomness in terms of the seeds used in the tests. 

`np.random.seed(0)` is still used in numerous places in other tests, meaning `pytest-randomly` isn't currently operating as intended (we end up with seed=0 for most tests), but since the seeds are set inside the tests the behavior of these tests is unchanged.

I'm tempted to leave replacing the remaining hard-coded seeds in tests as a future project. If there are any unintended issues with existing tests failing due to random seed setting, we could hardcode `--randomly-seed=0` in `setup.cfg` in the interim.

alibi_detect/utils/random.py
ascillitoe: I think the new global variable approach fixes this? I've extended the tests in `tests_random.py` to check that the seed returned by `get_seed` is unchanged after RNG calls.

alibi_detect/utils/_random.py
```diff
@@ -0,0 +1,80 @@
+from contextlib import contextmanager
+import random
+import numpy as np
+import os
+import sys
+from alibi_detect.utils.frameworks import has_pytorch, has_tensorflow
+
+# Pointer to the alibi_detect module. Used to set a global seed variable.
+_module_pointer = sys.modules[__name__]
+setattr(_module_pointer, '_seed', None)
```
jklaise, line 10: Not sure why this roundabout way instead of just `SEED = None` and then refer to `global SEED` in functions?

jklaise: Perhaps also a more clear vairable name and docstring, e.g. `ALIBI_SEED` or something like that.

alibi_detect/utils/_random.py
```diff
@@ -0,0 +1,80 @@
...
+    -------
+    >>> from alibi_detect.utils._random import set_seed, get_seed
+    >>> set_seed(42)
+    >>> get_seed()
+    42
+    """
+    if _module_pointer._seed is not None:
+        return _module_pointer._seed
+    else:
+        raise RuntimeError('`set_seed` must be called before `get_seed` can be called.')
```
jklaise, line 52: What's the reason for not accepting `None` as valid? Think I'm a bit lost in the motivation of these utilities again...

alibi_detect/utils/tests/test_random.py
```diff
@@ -20,10 +23,39 @@ def test_fixed_seed(seed):
     """
     Tests the fixed_seed context manager.
     """
+    n = 5  # Length of random number sequences
+
+    nums0 = np.empty((n, 3))
```
jklaise, line 28: parametrize both `n` and `3` to make it clearer/more flexible?

alibi_detect/utils/tests/test_random.py
```diff
@@ -20,10 +23,39 @@ def test_fixed_seed(seed):
...
     tmp_seed = seed + 42
     with fixed_seed(tmp_seed):
-        # Check seed inside of context manager
+        # Generate a sequence of random numbers
+        for i in range(n):
+            nums0[i, 0] = np.random.normal([1])
+            nums0[i, 1] = tf.random.normal([1])
+            nums0[i, 2] = torch.normal(torch.tensor([1.0]))
+
```
jklaise, line 36: Mixing fixed and parametrised indices...

alibi_detect/utils/tests/test_random.py
jklaise: Ah I guess `3` is the number of distinct frameworks. But is a `numpy` container the most natural one for heterogenous contents...?

alibi_detect/utils/_random.py
ascillitoe: WRT to motivation, the primary motivation of this PR now is to provide the `fixed_seed` context manager for use in tests. We want to be able to run detectors deterministically, i.e. 

```
with fixed_seed(42):
   detector1 = MMDDrift(X_ref, ...)
   preds1 = detector1.predict(x_h0)
```

But want to avoid setting global seeds, since these transfer into the next `pytest` test (I've checked this with a simple demo). This is quite an issue since in some tests we don't want the seeds fixed (see #250). If we move to use generators (e.g. `np.random.Generator`) in the future, and have `MMDDrift(x_ref, ..., random_seed=...)`, we can probably get rid of this entire `_random` module. 

The `set_seed` and `get_seed` methods might be helpful to a user in some way, but they're primarily for use with `fixed_seed` so I've kept the submodule private for now.

Does the above help or am I barking up the wrong tree? 🙂

alibi_detect/utils/_random.py
jklaise: You can never truly hide anything in Python... I'm not so sure it's worth the cognitive overhead of interpreting these lines rather than just having a global variable with a clear name (can make it private too e.g. `_ALIBI_SEED`).

jklaise (changes requested):
See individual comments.

alibi_detect/utils/tests/test_random.py
ascillitoe: Yeh its a fair point. I guess I could just unroll them all into a list? i.e. simply do `num0.append(...)` on all lines?

ascillitoe:
Rebased onto `feature/config_driven_detectors" since this functionality is initially intended for use in tests within #497.

requirements/dev.txt
```diff
@@ -2,12 +2,14 @@
 # linting and static type-checking
 flake8>=3.7.7, <5.0.0
 mypy~=0.900
-types-requests>=2.25.0, <2.28.0
+types-requests~=2.25
+types-toml~=0.10
```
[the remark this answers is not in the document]
ascillitoe: To clarify, `types-requests~=2.25` means `>= 2.25, == 2.*`, so I think we are OK, but have changed to `types-requests>=2.25, <3.0` for consistency with elsewhere.

jklaise (approved):
LGTM

PR #497, opened 2022-05-05
Config driven detectors - part 4

ascillitoe:


ascillitoe:
Making "Ready for review" to run CI.

ascillitoe:
**Note**: This PR currently duplicates the new random seed utilities in #496 for use in `saving/tests/test_saving.py`. Once (if) #496 is merged, we can rebase this PR and reduce the file diff.

.readthedocs.yml
```diff
@@ -5,21 +5,21 @@
...
 version: 2
 
+# Set the version of Python and other tools you might need
 build:
-  image: latest # Python 3.8 available on latest
-  apt_packages:
-    - pandoc
+  os: ubuntu-20.04
+  tools:
+    python: "3.9"
```
jklaise, line 12: What happened with `pandoc`?

ascillitoe: Thanks, this was an ommision. Weirdly `nbsphinx` still appeared to work OK, but I've added it back in to be safe.

ascillitoe:
This has now been synced with base (`feature/config_driven_detectors`) following the #496 PR, and is once again ready for review.

requirements/dev.txt
```diff
@@ -10,6 +10,7 @@ pytest-cov>=2.6.1, <4.0.0
 pytest-xdist>=1.28.0, <3.0.0 # for distributed testing, currently unused (see setup.cfg)
 pytest-randomly>=3.5.0, <4.0.0
 pytest_cases>=3.6.8, <4.0.0
+pytest-randomly>=3.5.0, <4.0.0
```
jklaise, line 13: Already listed 2 lines above?

requirements/docs.txt
```diff
@@ -12,3 +12,7 @@ ipykernel>=5.1.0, <6.0.0 # required for executing notebooks via nbsphinx
 ipython>=7.2.0, <8.0.0 # required for executing notebooks nbsphinx
 # pandoc
 # pandoc==1.16.02 # NB: as this is not a Python library, it should be installed manually on the system or via a package manager such as `conda`
+# alibi_detect dependencies (these are installed on ReadTheDocs so are not mocked)
+numpy>=1.16.2, <2.0.0
+pandas>=0.23.3, <2.0.0
+matplotlib>=3.0.0, <4.0.0
```
jklaise, line 18: Can you remind me the reason we want these on RTD? To render types better or another reason?

jklaise: Realized that this is explained in the `conf.py` - RTD 3.9 includes these deps.

doc/source/conf.py
```diff
@@ -205,6 +207,8 @@
     'preamble': r''' 
         \DeclareUnicodeCharacter{2588}{=}
         \DeclareUnicodeCharacter{258E}{|} 
+        \DeclareUnicodeCharacter{274C}{$\times$} 
+        \DeclareUnicodeCharacter{2705}{$\checkmark$}
```
jklaise, line 211: What are these symbols that latex can't parse so they have to be replaced?

alibi_detect/base.py
```diff
@@ -93,40 +93,89 @@ def infer_threshold(self, X: np.ndarray) -> None:
...
+        return detector
+
+    def _set_config(self, inputs):  # TODO - move to BaseDetector once config save/load implemented for non-drift
+        # Set config metadata
         name = self.__class__.__name__
         # strip off any backend suffix
         backends = ['TF', 'Torch', 'Sklearn']
         for backend in backends:
             if name.endswith(backend):
                 name = name[:-len(backend)]
```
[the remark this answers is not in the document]
jklaise: Ah is it because `self.meta = self._detector.meta` in the public classes? Confusing!

alibi_detect/base.py
```diff
@@ -93,40 +93,89 @@ def infer_threshold(self, X: np.ndarray) -> None:
...
+                'version': self.meta['version'],
+                'config_spec': __config_spec__,
+            }
+        }
 
-        # Add preprocess_fn field
-        if self.preprocess_fn is not None:
-            cfg.update({'preprocess_fn': self.preprocess_fn})
+        # args and kwargs
+        pop_inputs = ['self', '__class__', '__len__']
```
jklaise, line 169: Where do the `__class__` and `__len__` come from? Is this an exhaustive list of things to get rid of?

alibi_detect/base.py
```diff
@@ -93,40 +93,89 @@ def infer_threshold(self, X: np.ndarray) -> None:
...
+        pop_inputs += self.config.keys()  # Adding self.config.keys() avoids overwriting existing config
```
jklaise, line 170: Is this purely to avoid overwriting `name` and `meta` which we just created?

alibi_detect/base.py
```diff
@@ -93,40 +93,89 @@ def infer_threshold(self, X: np.ndarray) -> None:
...
+        config
+            A config dictionary matching the schema's in :class:`~alibi_detect.saving.schemas`.
+        """
+        # Check for exisiting version_warning. meta is pop'd as don't want to pass as arg/kwarg
+        version_warning = config.pop('meta', {}).pop('version_warning', False)
+        # Init detector
+        detector = cls(**config)
+        # Add version_warning
+        detector.meta['version_warning'] = version_warning  # type: ignore[attr-defined]
+        detector.config['meta']['version_warning'] = version_warning
```
jklaise, line 148: Not sure if there's much we can do, but noticed that this (and similar patterns) of duplication seem quite common now where we want to keep both `meta` and `config` in sync.

Perhaps for another PR, but what if said `config` is the source of truth and the `meta` attribute just points to the relevant `config` entry (could be a property)?

alibi_detect/saving/tensorflow/_loading.py
```diff
@@ -146,37 +149,27 @@ def load_kernel_config(cfg: dict) -> Callable:
     cfg
         A kernel config dict. (see pydantic schema's).
 
+    print(cfg_kernel)
```
jklaise, line 152: rogue line?

alibi_detect/cd/base.py
```diff
@@ -19,7 +19,7 @@
 logger = logging.getLogger(__name__)
 
 
-class BaseClassifierDrift(BaseDetector, DriftConfigMixin):
+class BaseClassifierDrift(BaseDetector):
```
jklaise, line 22: Is the intended paradigm to inherit from the `DriftConfigMixin` at the highest (public class) level only?

alibi_detect/cd/base.py
```diff
@@ -147,12 +142,15 @@ def preprocess(self, x: Union[np.ndarray, list]) -> Tuple[Union[np.ndarray, list
...
-            x_ref = self.preprocess_fn(self.x_ref) if self.preprocess_at_pred else self.x_ref
-            return x_ref, x
+            if not self.preprocess_at_init and not self.x_ref_preprocessed:
+                x_ref = self.preprocess_fn(self.x_ref)
+            else:
+                x_ref = self.x_ref
+            return x_ref, x  # type: ignore[return-value]
         else:
-            return self.x_ref, x
+            return self.x_ref, x  # type: ignore[return-value]
```
jklaise, line 153: This seems to be the same logic repeated for all detectors, perhaps we can pull that out in a follow-up PR?

Similar comment to the pre-process logic at `__init__` as well.

alibi_detect/cd/base.py
```diff
@@ -732,38 +660,12 @@ def predict(self, x: Union[np.ndarray, list], return_p_val: bool = True, return_
...
-
-        return cfg
-
-
-class BaseLSDDDrift(BaseDetector, DriftConfigMixin):
+class BaseLSDDDrift(BaseDetector):
     # TODO: TBD: this is only created when _configure_normalization is called from backend-specific classes,
     # is declaring it here the right thing to do?
     _normalize: Callable
+    _unnormalize: Callable
```
jklaise, line 668: Why do we need the inverse transform?

alibi_detect/cd/cvm.py
```diff
@@ -71,6 +71,8 @@ def __init__(
             input_shape=input_shape,
             data_type=data_type
         )
+        # Set config
+        self._set_config(locals())
```
jklaise, line 75: Noticed that in some cases you do `self._set_config(locals())` whilst in others it's
```
inputs = locals().copy()
self._set_config(inputs)
```

What's the difference?

alibi_detect/cd/lsdd.py
```diff
@@ -76,6 +77,9 @@ def __init__(
         """
         super().__init__()
```
jklaise, line 78: Not part of this PR, but noticed that these `super` calls don't seem to do anything as there is no superclass constructor. This is probably a relic of when we inherited the public classes from `BaseDetector` - conscious that we should revisit this again to enforce expectations on the class hierarchy - but don't remember if there's an issue in place?

alibi_detect/cd/lsdd.py
```diff
@@ -134,12 +140,17 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
...
         Returns
         -------
         The detector's configuration dictionary.
         """
-        return self._detector.get_config()
+        cfg = super().get_config()
+        # Unnormalize x_ref
+        if self._detector.preprocess_at_init or self._detector.preprocess_fn is None \
+                or self._detector.x_ref_preprocessed:
+            cfg['x_ref'] = self._detector._unnormalize(cfg['x_ref'])
```
jklaise, line 155: Rather confused by this, what happens if there is a user pre-processing function that is different from `_normalize`? Won't we end up with `x_ref` that is completely different from the original `x_ref`?

alibi_detect/cd/model_uncertainty.py
```diff
@@ -27,6 +29,7 @@ def __init__(
             device: Optional[str] = None,
             tokenizer: Optional[Callable] = None,
             max_len: Optional[int] = None,
+            input_shape: Optional[tuple] = None,
```
jklaise, line 32: This change seems unrelated to this PR?

alibi_detect/cd/model_uncertainty.py
```diff
@@ -173,6 +182,7 @@ def __init__(
...
```
jklaise, line 185: This change seems unrelated to this PR?

alibi_detect/cd/tensorflow/lsdd.py
```diff
@@ -80,7 +80,7 @@ def __init__(
         )
         self.meta.update({'backend': 'tensorflow'})
 
-        if self.preprocess_at_init:
+        if self.preprocess_at_init or self.preprocess_fn is None or self.x_ref_preprocessed:
```
jklaise, line 83: Struggling to understand this logic (applies to all detectors)...

alibi_detect/cd/tensorflow/lsdd.py
```diff
@@ -132,7 +133,7 @@ def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray]:
         """
         x_ref, x = self.preprocess(x)
 
-        if not self.preprocess_at_init:
+        if self.preprocess_fn is not None and self.preprocess_at_init is False and not self.x_ref_preprocessed:
```
jklaise, line 136: Struggling to understand this logic (applies to all detectors)...

alibi_detect/saving/loading.py
```diff
@@ -28,60 +28,38 @@
...
 
-REQUIRES_BACKEND = [
-    'ClassifierDrift',
-    'ClassifierUncertaintyDrift',
-    'LearnedKernelDrift',
-    'LSDDDrift',
-    'MMDDrift',
-    'RegressorUncertaintyDrift',
-    'SpotTheDiffDrift'
-]
```
jklaise, line 39: Why is this no longer necessary?

alibi_detect/saving/loading.py
```diff
@@ -28,60 +28,38 @@
...
 
 # Fields to resolve in resolve_config ("resolve" meaning either load local artefact or resolve @registry, conversion to
 # tuple, np.ndarray and np.dtype are dealt with separately).
+# Note: For fields consisting of nested dicts, they must be listed in order from deepest to shallowest, so that the
+# deepest fields are resolved first. e.g. 'preprocess_fn.src' must be resolved before 'preprocess_fn'.
```
jklaise, line 35: However, `['preprocess_fn', 'src']` is in the opposite order?

alibi_detect/saving/loading.py
```diff
@@ -28,60 +28,38 @@
...
     ['kernel', 'kernel_b', 'src'],
-    ['initial_diffs']
-]
-
-# Directories to amend before resolving config (fields to prepend config file dir to).
-# e.g. in a `config.toml` stored in `filepath='my_detector'` we might have `preprocess_fn.model.src = 'model/'.
-# Before resolution, we need to update `preprocess_fn.model.src` to be relative to runtime dir i.e. 'my_detector/model/'
-# Note some fields in FIELDS_TO_RESOLVE are missing from DIR_FIELDS, since some fields cannot be spec'd as directories.
-# e.g. embedding cannot be spec'd directly as a .dill file reference, it must be an EmbeddingConfig.
-DIR_FIELDS = [
```
jklaise, line 68: Can you summarize the new logic and how it's different/superior to the old one having `DIR_FIELDS` separately?

alibi_detect/saving/saving.py
```diff
@@ -203,12 +211,19 @@ def write_config(cfg: dict, filepath: Union[str, os.PathLike]):
...
+    # Create directory if it doesn't exist
     filepath = Path(filepath)
     if not filepath.is_dir():
         logger.warning('Directory {} does not exist and is now created.'.format(filepath))
         filepath.mkdir(parents=True, exist_ok=True)
+    # Convert pathlib.Path's to str's
     cfg = _path2str(cfg)
+    # Validatre config before final tweaks
+    validate_config(cfg)  # Must validate here as replacing None w/ str will break validation
+    # Replace None with "None", and dicts with integer keys with str keys (TODO: S2C depending on toml library updates)
```
jklaise, line 223: What's `S2C`?

alibi_detect/saving/schemas.py
```diff
@@ -78,7 +95,7 @@ class DetectorConfig(CustomBaseModel):
 
 class ModelConfig(CustomBaseModel):
     """
-    Unresolved schema for (ML) models. Note that the model "backend" e.g. 'tensorflow', 'pytorch', 'sklearn', is set
+    Schema for (ML) models. Note that the model "backend" e.g. 'tensorflow', 'pytorch', 'sklearn', is set
```
jklaise, line 98: What's the reason for changing the description from "Unresolved schema" to just "schema"?

alibi_detect/saving/schemas.py
```diff
@@ -261,61 +278,12 @@ class PreprocessConfig(CustomBaseModel):
     """
 
 
-class PreprocessConfigResolved(CustomBaseModel):
```
jklaise, line 264: What happened to this schema?

alibi_detect/saving/schemas.py
```diff
@@ -786,24 +757,26 @@ class SpotTheDiffDriftConfig(DriftDetectorConfig):
...
-    optimizer: Optional[Union[str, dict]] = None  # dict as can pass dict to tf.keras.optimizers.deserialize
+    optimizer: Optional[Union[str, OptimizerConfig]] = None
     learning_rate: float = 1e-3
     batch_size: int = 32
     preprocess_batch_fn: Optional[str] = None
     epochs: int = 3
     verbose: int = 0
     train_kwargs: Optional[dict] = None
-    dataset: str = '@alibi_detect.utils.tensorflow.data.TFDataset'
+    dataset: str = None
```
jklaise, line 773: I'm guessing the `TFDataset` wasn't the right thing to put here?

alibi_detect/saving/schemas.py
```diff
@@ -889,6 +867,314 @@ class LearnedKernelDriftConfigResolved(DriftDetectorConfigResolved):
     train_kwargs: Optional[dict] = None
     dataset: Optional[Callable] = None
     device: Optional[Literal['cpu', 'cuda']] = None
+    dataloader: Optional[Callable] = None  # TODO: placeholder, will need to be updated for pytorch implementation
+
+
+class ContextMMDDriftConfig(DriftDetectorConfig):
+    """
+    Unresolved schema for the
```
jklaise, line 875: Remove "unresolved" since you've done it for others..?

alibi_detect/saving/schemas.py
```diff
@@ -889,6 +867,314 @@ class LearnedKernelDriftConfigResolved(DriftDetectorConfigResolved):
...
+    prop_c_held: float = 0.25
+    n_folds: int = 5
+    batch_size: Optional[int] = 256
+    verbose: bool = False
+    device: Optional[Literal['cpu', 'cuda']] = None
+
+
+class MMDDriftOnlineConfig(DriftDetectorConfig):
+    """
+    Unresolved schema for the
```
jklaise, line 920: Remove "unresolved" since you've done it for others..?

alibi_detect/saving/schemas.py
```diff
@@ -889,6 +867,314 @@ class LearnedKernelDriftConfigResolved(DriftDetectorConfigResolved):
...
+    kernel: Optional[Callable] = None
+    sigma: Optional[np.ndarray] = None
+    n_bootstraps: int = 1000
+    device: Optional[Literal['cpu', 'cuda']] = None
+    verbose: bool = True
+
+
+class LSDDDriftOnlineConfig(DriftDetectorConfig):
+    """
+    Unresolved schema for the
```
jklaise, line 956: Remove "unresolved" since you've done it for others..?

alibi_detect/saving/schemas.py
```diff
@@ -889,6 +867,314 @@ class LearnedKernelDriftConfigResolved(DriftDetectorConfigResolved):
...
+    n_bootstraps: int = 1000
+    n_kernel_centers: Optional[int] = None
+    lambda_rd_max: float = 0.2
+    device: Optional[Literal['cpu', 'cuda']] = None
+    verbose: bool = True
+
+
+class CVMDriftOnlineConfig(DriftDetectorConfig):
+    """
+    Unresolved schema for the
```
jklaise, line 994: Remove "unresolved" since you've done it for others..?

alibi_detect/saving/schemas.py
```diff
@@ -889,6 +867,314 @@ class LearnedKernelDriftConfigResolved(DriftDetectorConfigResolved):
...
+    n_features: Optional[int] = None
+    verbose: bool = True
+
+    # validators
+    _coerce_int2list = validator('window_sizes', allow_reuse=True, pre=True)(coerce_int2list)
+
+
+class FETDriftOnlineConfig(DriftDetectorConfig):
+    """
+    Unresolved schema for the
```
jklaise, line 1034: Remove "unresolved" since you've done it for others..?

alibi_detect/saving/schemas.py
```diff
@@ -889,6 +867,314 @@ class LearnedKernelDriftConfigResolved(DriftDetectorConfigResolved):
...
+
+    # validators
+    _coerce_int2list = validator('window_sizes', allow_reuse=True, pre=True)(coerce_int2list)
+
+
+# The uncertainty detectors don't inherit from DriftDetectorConfig since their kwargs are a little different from the
+# other drift detectors (e.g. no preprocess_fn). Subject to change in the future.
+class ClassifierUncertaintyDriftConfig(DetectorConfig):
+    """
+    Unresolved schema for the
```
jklaise, line 1080: Remove "unresolved" since you've done it for others..?

alibi_detect/saving/schemas.py
```diff
@@ -889,6 +867,314 @@ class LearnedKernelDriftConfigResolved(DriftDetectorConfigResolved):
...
+    device: Optional[str] = None
+    tokenizer: Optional[Union[str, Callable]] = None
+    max_len: Optional[int] = None
+    input_shape: Optional[tuple] = None
+    data_type: Optional[str] = None
+
+
+class RegressorUncertaintyDriftConfig(DetectorConfig):
+    """
+    Unresolved schema for the
```
jklaise, line 1132: Remove "unresolved" since you've done it for others..?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -33,7 +33,7 @@ class GaussianRBF(nn.Module):
     def __init__(
         self,
         sigma: Optional[torch.Tensor] = None,
-        init_sigma_fn: Callable = sigma_median,
+        init_sigma_fn: Optional[Callable] = None,
```
jklaise, line 36: Not sure why this change is needed, esp. since `None` will lead to `sigma_median`? Is it some issue with the `pydantic` models or serialization?

Would need to document that `None` leads to `sigma_median`.

Same for the `tensorflow` version.

doc/source/cd/background.md
```diff
@@ -274,8 +274,11 @@ the reconstruction loss, e.g. if MSE is used:
 $\phi, \psi = \text{arg} \min_{\phi, \psi}\, \lVert \mathbf{X}-(\phi \circ \psi)\mathbf{X}\rVert^2$.
 However, untrained (randomly initialised) autoencoders can also be used.
 For an example, a `pytorch` autoencoder can be incorporated into a
-detector by packaging it as a callable function using {func}`~alibi_detect.cd.pytorch.preprocess.preprocess_drift`
-and {func}`~functools.partial`: 
+detector by packaging it as a callable function using
+`preprocess_drift` (from
+[alibi_detect.cd.pytorch.preprocess](../api/alibi_detect.cd.pytorch.preprocess.rst))
+and `partial` (see
+[here](https://docs.python.org/3/library/functools.html#functools.partial)):
```
jklaise, line 281: Why the change to the link style here and elsewhere?

doc/source/overview/config_files.md
```diff
@@ -54,21 +54,17 @@ detector = load_detector(filepath)
...
+[pydantic](https://pydantic-docs.helpmanual.io/) to validate detector configurations.
 
-In what follows, the Alibi Detect config files are explored in some detail. To get a general idea 
-of the expected layout of a config file, readers can also skip ahead to [Example config files](examples) for examples of
-config files for some common use cases. Alternatively, to obtain a fully populated config file for reference, users
-can run one of the [example notebooks](../cd/examples.md) and generate a config file by passing an instantiated 
-detector to {func}`~alibi_detect.saving.save_detector`.
+To get a general idea of the expected layout of a config file, see the [Example config files](examples). Alternatively, 
+to obtain a fully populated config file for reference, users can run one of the [example notebooks](../cd/examples.md) 
+and generate a config file by passing an instantiated detector to {func}`~alibi_detect.saving.save_detector`.
```
jklaise, line 67: Note that the change to link style hasn't been applied to these ones, unlike above.

alibi_detect/saving/tests/test_saving.py
```diff
@@ -593,7 +921,7 @@ def test_load_absolute(data, tmp_path):
     # Write a new cfg file elsewhere, with x_ref reference inside it an absolute path to original x_ref location
     cfg = read_config(tmp_path.joinpath('config.toml'))
     x_ref_path = tmp_path.joinpath(Path(cfg['x_ref'])).resolve()  # Absolute path for x_ref
-    cfg['x_ref'] = x_ref_path
+    cfg['x_ref'] = x_ref_path.as_posix()  # we always write paths to config.toml as Posix not Windows paths
```
jklaise, line 924: Seems like this should be documented in the main saving implementation (also refer to the comment above about the one use of `as_posix` which I found - should there be more?).

jklaise (changes requested):
Thanks Ashley, please see the individual comments.

alibi_detect/cd/base.py
ascillitoe: Yes, in order to limit duplication, the convention is to inherit `DriftConfigMixin` in the relevent `cd.base` class.

However, this is complicated by the way we implement the "dispatch" wrapper classers for back-end specific detectors such as `MMDDrift`. These don't inherit from a base class so we have to inherit `DriftConfigMixin` there instead. Inheriting in `MMDDriftTF` etc would also complicate things since their kwargs are different to their public class.

Think we'd need to write up conventions here in `CONTRIBUTING.md`...

alibi_detect/cd/base.py
ascillitoe: Good idea! Issue opened here: https://github.com/SeldonIO/alibi-detect/issues/512

alibi_detect/cd/base.py
ascillitoe: The `LSDDDrift` preprocess logic is a little different to the other detectors since `self.x_ref` is normalized (regardless of whether `x_ref` is preprocessed by `preprocess_fn`):

https://github.com/SeldonIO/alibi-detect/blob/644ec298f73674175e605696ceeab5859864198a/alibi_detect/cd/tensorflow/lsdd.py#L83-L87

To avoid having to add another `x_ref_preprocessed` kwarg specific to whether `x_ref` has been normalized, the easiest thing to do is just unormalise `x_ref` before returning the config:

https://github.com/SeldonIO/alibi-detect/blob/644ec298f73674175e605696ceeab5859864198a/alibi_detect/cd/lsdd.py#L151-L156

alibi_detect/cd/cvm.py
ascillitoe: In the non-backend specific detectors we simply have `self._set_config(locals())` right after `__init__`, to ensure we get the "fresh" unedited args/kwargs. 

In the backend-specific detectors we can't call `self._set_config()` until `self.meta` has been set, therefore we have to call it at the end of the `__init__` method. `inputs = locals()` is still called immediately after `__init__` so that we capture the original args/kwargs (e.g. before `backend` etc edited). The `.copy()` was added as `items` being mutable meant it's items were modified before `_set_config` is reached.

The above works, but unsure if its the best approach... Would need documenting in `CONTRIBUTING.md` at the very least...

alibi_detect/cd/lsdd.py
ascillitoe: Yeh I can't recall. I did remove once as a test and think it was fine, but should try again. Also added to this PR: https://github.com/SeldonIO/alibi-detect/issues/512.

alibi_detect/cd/model_uncertainty.py
ascillitoe: The `input_shape` info is used when saving TensorFlow models. It was missing from detectors that didn't have save functionality with the legacy `save_detector`, so has now been added since all drift detectors are now "saveable".

alibi_detect/cd/model_uncertainty.py
ascillitoe: As above.

alibi_detect/saving/loading.py
ascillitoe: `_init_config()` has been simplified by the use of `from_config()`, and we no longer need to manually update `backend`:

https://github.com/SeldonIO/alibi-detect/blob/644ec298f73674175e605696ceeab5859864198a/alibi_detect/saving/loading.py#L160-L166

alibi_detect/saving/loading.py
ascillitoe: The way the config dict recursion works is such that:

```python
FIELDS_TO_RESOLVE = [
    ['preprocess_fn', 'src'],
    ['preprocess_fn', 'model'],
    ['preprocess_fn', 'embedding'],
    ['preprocess_fn', 'tokenizer'],
    ['preprocess_fn', 'preprocess_batch_fn'],
    ['preprocess_fn'],
```

means resolve `src` in `preprocess_fn`, then resolve `model` in `preprocess_fn`, etc etc, and finally resolve `preprocess_fn`. We need to have resolved all the possible artefacts in `preprocess_fn` before we resolve `preprocess_fn` itself. Hence why `['preprocess_fn', 'src']` must be listed before `['preprocess_fn']`. Don't know if that clarifies things?

alibi_detect/saving/loading.py
ascillitoe: Previously we did the following in `resolve_config`:

```python
        for key in DIR_FIELDS:
            src = _get_nested_value(cfg, key)
            if isinstance(src, str):
                src = config_dir.joinpath(Path(src))
                if src.is_file() or src.is_dir():
                    _set_nested_value(cfg, key, str(src))
        _prepend_cfg_filepaths(cfg, config_dir)
```

So we looped through `DIR_FIELDS` and prepended the `config.toml` directory to all the filepaths within it (so that we can do `x_ref = 'x_ref.npy'` in the `config.toml`, instead of having to do `x_ref = 'filepath/x_ref.npy'` where `filepath` is the path to the `config.toml` itself. 

`_prepend_cfg_filepaths` has now been modified so that it recurses through all fields in the config dict, attempts to prepend the `config.toml` filepath to any strings, and if the result of that gives a valid directory or file, it updates the config. In other words, we no longer have to specify `DIR_FIELDS` separatetly (the difference between this and `FIELDS_TO_RESOLVE` had caused some confusion previously, so not ideal to maintain both).

alibi_detect/saving/saving.py
ascillitoe: Sorry, subject to change. I've written it out properly now!

alibi_detect/saving/saving.py
```diff
@@ -409,14 +448,13 @@ def _save_tokenizer_config(tokenizer: PreTrainedTokenizerBase,
     cfg_token = {}
     logger.info('Saving tokenizer to {}.'.format(filepath))
     tokenizer.save_pretrained(filepath)
-    cfg_token.update({'src': path.joinpath('tokenizer')})
+    cfg_token.update({'src': path.joinpath('tokenizer').as_posix()})
```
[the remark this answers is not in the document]
ascillitoe: This is a Windows-specific issue. Basically if we write Windows paths to the `config.toml` (i.e. with `\` instead of `/`), it causes all sorts of pain upon loading.

There are only a few limited places where its neccesary:
- `cfg_model = {'src': local_path.joinpath('model').as_posix()}` in `tensorflow/_saving.py`.
- `cfg_embed.update({'src': local_path.as_posix()})` in `tensorflow/_saving.py`.
- `cfg_token.update({'src': path.joinpath('tokenizer').as_posix()})` in `saving.py`.

It isn't required for some other operations such as:

```python
save_path = filepath.joinpath(X_REF_FILENAME)
```

because currently `X_REF_FILENAME = 'x_ref.npy'`. (if we did e.g. `data/x_ref.npy` it'd need updating with `as_posix()`. 
 
I've just made a commit with `as_posix()` moved to `saving._path2str` so that any necessary paths are automatically converted, instead of us having to think about whether to do each in-place. Will see if it passes on the Windows build...

alibi_detect/saving/schemas.py
ascillitoe: It technically wasn't wrong because the config used to be associated with `SpotTheDiffTF`, whose kwarg was `dataset: Callable = TFDataset`. Now we (imo correctly) associate the config with `SpotTheDiff` class, whose kwarg is `dataset: Optional[Callable] = None`, hence the update. 

This is the same for all detectors with backends; the pydantic model fields now correspond to the top-level backend-agnostic class.

P.s. `Optional` was missing from the above. Have added!

alibi_detect/saving/schemas.py
ascillitoe: Detectors are different, as they do have an unresolved and resolved pydantic model.

alibi_detect/saving/schemas.py
ascillitoe: See above.

alibi_detect/saving/schemas.py
ascillitoe: https://github.com/SeldonIO/alibi-detect/pull/497#discussion_r879437444

alibi_detect/utils/pytorch/kernels.py
ascillitoe: Yep as you guessed, it's to make it play nice with `pydantic`, i.e. we can have `init_sigma_fn: Optional[str] = None` in `KernelConfig`, instead of needing to have a validator which defaults to `@alibi_detect.utils.tensorflow.kernels.sigma_median` or the `torch` equivalent, dependent on `backend`. The currently implemented behaviour is similar to how we handle `dataset` in `SpotTheDiff` etc (https://github.com/SeldonIO/alibi-detect/pull/497#discussion_r878198807) i.e. default is `None` and then set to backend-specific object internally.

doc/source/cd/background.md
ascillitoe: At a loss as to how this has snuck in. Will revert.

requirements/dev.txt
ascillitoe: Thanks!

doc/source/conf.py
ascillitoe: The cross and tick parks in the "Saving and Loading" page:

image

alibi_detect/base.py
ascillitoe: Yep. Because `inputs` is a dict, it has `self`, `__class__` and `__len__` attributes. We want to strip these so that we're left with only the args and kwargs. Also do `pop_inputs += self.config.keys()` so as to avoid overwriting existing args/kwargs.

alibi_detect/base.py
```diff
@@ -93,40 +93,89 @@ def infer_threshold(self, X: np.ndarray) -> None:
...
+        if self.config is not None:
+            # Get config (stored in top-level self)
+            cfg = self.config
+            # Get low-level nested detector (if needed)
+            detector = self._detector if hasattr(self, '_detector') else self  # type: ignore[attr-defined]
+            detector = detector._detector if hasattr(detector, '_detector') else detector  # type: ignore[attr-defined]
+            # Add large artefacts back to config
+            for key in LARGE_ARTEFACTS:
+                if key in cfg:  # self.config is validated, therefore if a key is not in cfg, it isn't valid to insert
+                    cfg[key] = getattr(detector, key)
```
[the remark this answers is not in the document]
ascillitoe: Good point. It's OK at the moment since the above assumption is currently met (`x_ref` always has the same name regardless, the `x_ref_preprocessed` boolean is used to signify if preprocessing has occurred). We might have to revisit if we adopt this strategy for more artefacts...

alibi_detect/base.py
ascillitoe: I like that idea, although we'll have to carefully think about how to handle artefacts in the config (`get_config` currently returns a config containing large artefacts, because returning a json-able type config would require the artefacts to be saved, and we had concerns with `get_config` writing to disk). Regardless, definitely worth thinking about what `meta` entails...

alibi_detect/saving/tests/test_saving.py
ascillitoe: Have added a note to the docs:
image

See https://github.com/SeldonIO/alibi-detect/pull/497#discussion_r879426796 regarding `as_posix`.

alibi_detect/saving/schemas.py
ascillitoe: https://github.com/SeldonIO/alibi-detect/pull/497#discussion_r879437444

ascillitoe:
P.s. @jklaise, these sections in the docs methods pages need updating:

image

I've added a TODO in the PR description to do this before merging into `master`, as don't want to add to the file diff in this PR.

alibi_detect/base.py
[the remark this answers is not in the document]
jklaise: @ascillitoe can you open an issue? This is a pretty confusing "bug" as the name should be the name of the wrapper class (unless we change conventions).

alibi_detect/base.py
jklaise: I'm confused, a `locals()` `dict` has many more attributes and methods than these, I'm not sure what's so special about e.g. `__len__`?

alibi_detect/cd/base.py
jklaise: I see, unrelated but to be honest I find the following condition very confusing:
```python
if self.preprocess_at_init or self.preprocess_fn is None or self.x_ref_preprocessed
```

So the data is normalized when either:
 - we specify that we want to pre-process at init (but then preprocessing function doesn't have to exist?)
 - we don't have a preprocessing function of our own (??)
 - the data is already preprocessed (which implies a preprocessing function was passed?)

It seems like there is something missing here...

alibi_detect/cd/cvm.py
jklaise: Yea, needs documenting.

alibi_detect/cd/model_uncertainty.py
jklaise: Does it mean the detectors are only saveable if the correct `input_shape` is passed by the user? Is this documented?

alibi_detect/cd/base.py
jklaise: I see it's explained below, will dig into that...

alibi_detect/saving/loading.py
jklaise: Alright, I was thinking resolving happens on every item in row, but actually it's row-by-row.

alibi_detect/utils/pytorch/kernels.py
jklaise: Perhaps an issue to make sure that the defaults are documented in the docstrings in all the places where `None` is going to be replaced by something default in the constructor?

jklaise (approved):
LGTM given the discussion from the previous review.

alibi_detect/base.py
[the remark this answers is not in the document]
ascillitoe: Done! https://github.com/SeldonIO/alibi-detect/issues/514

alibi_detect/cd/model_uncertainty.py
ascillitoe: The detector would be saveable without, but only if it didn't have a `model` defined as part of its `preprocess_fn`. It isn't documented. I'm now actually leaning towards introducing a new top-level drift detector base class to handle args/kwargs shared across all detectors, in which case we could add including `input_shape` (and `data_type` etc) to this issue: https://github.com/SeldonIO/alibi-detect/issues/512.

I've already done something similar for the pydantic models, where we have a top-level `DriftDetectorConfig`...

alibi_detect/base.py
jklaise: ok interesting...

alibi_detect/cd/classifier.py
```diff
@@ -142,6 +143,9 @@ def __init__(
         """
         super().__init__()
 
+        # Get args/kwargs to set config later
+        inputs = locals().copy()
```
mauicv, line 147: will this duplicate the memory during init even if it's later gc'd?

ascillitoe: P.s. IMO the above complexity is a symptom of the way we set detector backends i.e. `self._detector = MMDDriftTF(...)`. It is slightly odd how `MMDDrift` is not a detector itself, but only a container with a detector as an attribute. 

I've been discussing with @mauicv how we might want to refactor this, and will add some ideas to https://github.com/SeldonIO/alibi-detect/issues/512.

alibi_detect/saving/loading.py
```diff
@@ -607,3 +532,27 @@ def _replace(cfg: dict, orig: Optional[str], new: Optional[str]) -> dict:
...
+        The filepath to prepend to any filepaths in the dictionary.
+
+    Returns
+    -------
+    The updated config dictionary.
+    """
+    for k, v in cfg.items():
+        if isinstance(v, str):
+            v = prepend_dir.joinpath(Path(v))
+            if v.is_file() or v.is_dir():  # Update if prepending config_dir made config value a real filepath
```
mauicv, line 555: Do you iterate through recursively and if it's a string you prepend the directory and see if it exists in the file system? I'm trying to think if there's some way this might backfire?

alibi_detect/saving/saving.py
```diff
@@ -132,33 +134,39 @@ def _save_detector_config(detector: Detector, filepath: Union[str, os.PathLike])
         filepath.mkdir(parents=True, exist_ok=True)
 
     # Get the detector config (with artefacts still within it)
-    cfg = getattr(detector, 'get_config', None)
-    if cfg is not None:
-        cfg = cfg()  # TODO - can just do detector.get_config() once all detectors have a .get_config()
+    if hasattr(detector, 'get_config'):
+        cfg = detector.get_config()  # type: ignore[union-attr]  # TODO - remove once all detectors have get_config
```
mauicv, line 138: We should accumulate all these todo's in an issue somewhere once this is merged so that when stuff gets fixed we know what to update

alibi_detect/cd/context_aware.py
```diff
@@ -88,7 +100,7 @@ def __init__(
 
         kwargs = locals()
         args = [kwargs['x_ref'], kwargs['c_ref']]
-        pop_kwargs = ['self', 'x_ref', 'c_ref', 'backend', '__class__']
+        pop_kwargs = ['self', 'x_ref', 'c_ref', 'backend', '__class__', 'inputs']
```
mauicv, line 103: Why do we ned `inputs` here?

alibi_detect/cd/learned_kernel.py
```diff
@@ -125,7 +129,7 @@ def __init__(
 
         kwargs = locals()
         args = [kwargs['x_ref'], kwargs['kernel']]
-        pop_kwargs = ['self', 'x_ref', 'kernel', 'backend', '__class__']
+        pop_kwargs = ['self', 'x_ref', 'kernel', 'backend', '__class__', 'inputs']
```
mauicv, line 132: Can `inputs` be removed?

alibi_detect/cd/lsdd.py
```diff
@@ -85,7 +89,7 @@ def __init__(
 
         kwargs = locals()
         args = [kwargs['x_ref']]
-        pop_kwargs = ['self', 'x_ref', 'backend', '__class__']
+        pop_kwargs = ['self', 'x_ref', 'backend', '__class__', 'inputs']
```
mauicv, line 92: Can inputs be removed?

alibi_detect/cd/mmd.py
```diff
@@ -84,7 +88,7 @@ def __init__(
 
         kwargs = locals()
         args = [kwargs['x_ref']]
-        pop_kwargs = ['self', 'x_ref', 'backend', '__class__']
+        pop_kwargs = ['self', 'x_ref', 'backend', '__class__', 'inputs']
```
mauicv, line 91: Can inputs be removed?

alibi_detect/cd/mmd_online.py
```diff
@@ -74,7 +83,7 @@ def __init__(
 
         kwargs = locals()
         args = [kwargs['x_ref'], kwargs['ert'], kwargs['window_size']]
-        pop_kwargs = ['self', 'x_ref', 'ert', 'window_size', 'backend', '__class__']
+        pop_kwargs = ['self', 'x_ref', 'ert', 'window_size', 'backend', '__class__', 'inputs']
```
mauicv, line 86: Can inputs be removed?

alibi_detect/cd/spot_the_diff.py
```diff
@@ -130,7 +134,7 @@ def __init__(
 
         kwargs = locals()
         args = [kwargs['x_ref']]
-        pop_kwargs = ['self', 'x_ref',  'backend', '__class__']
+        pop_kwargs = ['self', 'x_ref',  'backend', '__class__', 'inputs']
```
mauicv, line 137: Can inputs be removed?

alibi_detect/cd/tabular.py
```diff
@@ -146,27 +148,3 @@ def _get_counts(self, x: np.ndarray, categories: Dict[int, List[int]]) -> Dict[i
...
-        """
-        Get the detector's configuration dictionary.
-
-        Returns
-        -------
-        The detector's configuration dictionary.
-        """
-        cfg = super().get_config()
-
-        # Prep categories_per_feature. NOTE - toml can't write dict with int keys. See loading.read_detector_config
```
mauicv, line 160: Where is this done now that it's not done here?

alibi_detect/cd/context_aware.py
ascillitoe: It was there from where we had `inputs = locals().copy()` but you're right it should be removed now.

alibi_detect/cd/tabular.py
ascillitoe: Its done in `_int2str_keys` in `saving.saving` ([here](https://github.com/ascillitoe/alibi-detect/blob/ab3562fc2ade48cf88eb8a81dd38e35b33602925/alibi_detect/saving/saving.py#L226)).

At load time, `pydantic` does the reverse for us (coerces `str` keys back to `int`), since we have the following in the `pydantic` models:
```python
categories_per_feature: Dict[int, Union[int, List[int]]] = None
```

ascillitoe: Its done in `_int2str_keys` in `saving.saving` ([here](https://github.com/ascillitoe/alibi-detect/blob/ab3562fc2ade48cf88eb8a81dd38e35b33602925/alibi_detect/saving/saving.py#L226)).

At load time, `pydantic` does the reverse for us (coerces `str` keys back to `int`), since we have the following in the `pydantic` models:
```python
categories_per_feature: Dict[int, Union[int, List[int]]] = None
```

alibi_detect/base.py
```diff
@@ -93,40 +93,86 @@ def infer_threshold(self, X: np.ndarray) -> None:
...
+        Returns
+        -------
+        The detector's configuration dictionary.
+        """
+        if self.config is not None:
+            # Get config (stored in top-level self)
+            cfg = self.config
+            # Get low-level nested detector (if needed)
+            detector = self._detector if hasattr(self, '_detector') else self  # type: ignore[attr-defined]
+            detector = detector._detector if hasattr(detector, '_detector') else detector  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
ascillitoe: Yes. The `SpotTheDiffDrift` detector. It sets `self._detector = SpotTheDiffDriftTF`, which itself sets `self._detector = ClassifierDriftTF`. 🙈

alibi_detect/saving/loading.py
ascillitoe: Yes exactly that. Then only update the string in the config (with the new prepended file path) if it is a file or dir. 

The two ways I can think of which it might backfire are:
1. If a config contains a field which a `str` but shouldn't be a filepath (e.g. a regsitry str), whose string is such that when prepending the config filepath to it the new string is actually a filepath. This field would then incorrectly be updated. I can't think of a realistic example of where this would happen, but it is theoretically possible.
2. The bigger disadvantage is that if a filepath has been spec'd in the config but it is missing/incorrect, it won't be updated with the prepended path. The `FileNotFoundError` error raised later would then be confusing if it prints out the problem directory. The easiest thing to do is probs just go through and check no `FileNotFoundError`'s print out the given directory...

ascillitoe: Following up on this, I've actually decided not to remove the filepath included in `FileNotFoundError`'s such as that in `saving.tensorflow._loading.load_model`:

```python
raise FileNotFoundError(f'No .h5 file found in {model_dir}.')
```

Although `model_dir` wouldn't have the config file dir prepended to it (see 2. above), the filepath would still match that in `config.toml`, which might actually be a nice feature to aid the user's debugging...

@mauicv shout if you disagree, otherwise think we're good to go!

PR #498, opened 2022-05-06
Revert "Replace autodoc with autoapi (#482)"

ascillitoe:
This reverts commit bd0fa0051a7ce8b836a7b33bce0cec33c5443d4f. The move to sphinx autoapi (#482) has caused a few problems; for all classes autoapi automatically includes the docstring of any super class if they can be discovered (i.e. if their packages are installed at build time). For example, see the api doc for [alibi_detect.base.NumpyEncoder](https://seldon--482.org.readthedocs.build/projects/alibi-detect/en/482/api/alibi_detect/base/index.html#alibi_detect.base.NumpyEncoder):

image

It has included the entire docstring for [json.JSONEncoder](https://docs.python.org/3/library/json.html#json.JSONEncoder). Since there isn't currently a way to turn off this docstring inheritance, and compatibility with some `sphinx` versions has proved to be flaky, this PR reverts back to `autodoc` and `sphinx-apidoc` for now. Moving to `sphinx-autoapi` was also to aid building of the new `pydantic` api docs in #469, but tweaking which modules are mocked and moving to Python 3.9 on docker seems to have resolved the majority of issues here (see #497).

**Note**: A small number of changes to links in `cd/background.md` have been retained, as these match the (IMO nicer) directive format mentioned in [CONTRIBUTING.md](https://github.com/SeldonIO/alibi/blob/master/CONTRIBUTING.md).

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -44,6 +44,7 @@ release = __version__
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
+    "sphinx.ext.autodoc",
     "sphinx.ext.doctest",
     "sphinx.ext.intersphinx",
     "sphinx.ext.todo",
@@ -52,12 +53,12 @@ extensions = [
     "sphinx.ext.ifconfig",
     "sphinx.ext.viewcode",
     "sphinx.ext.napoleon",
-    "sphinx.ext.autodoc.typehints",  # still used with autoapi
+    "sphinx_autodoc_typehints",
+    "sphinxcontrib.apidoc",  # automatically generate API docs, see https://github.com/rtfd/readthedocs.org/issues/1139
     "sphinxcontrib.bibtex",
     "nbsphinx",
     "myst_parser",
     "sphinx_design",
-    "autoapi.extension"
 ]
 
 # -- nbsphinx settings -------------------------------------------------------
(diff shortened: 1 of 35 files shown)


doc/source/conf.py
```diff
@@ -52,12 +53,12 @@
     "sphinx.ext.ifconfig",
     "sphinx.ext.viewcode",
     "sphinx.ext.napoleon",
-    "sphinx.ext.autodoc.typehints",  # still used with autoapi
+    "sphinx_autodoc_typehints",
```
jklaise, line 56: I know this is a reversion, but what's the difference between this and the `sphinx.ext` version?

requirements/docs.txt
```diff
@@ -1,9 +1,9 @@
 # dependencies for building docs, separate from dev.txt as this is also used for builds on readthedocs.org
 # core dependencies
 sphinx>=4.2.0, <5.0.0
-sphinx-autodoc-typehints>=1.12.0, <2.0.0
-sphinx-autoapi>=1.8.0, <2.0.0
-sphinx-rtd-theme>=1.0.0, <2.0.0
+sphinx-autodoc-typehints>=1.12.0, <=2.0.0
+sphinx-rtd-theme>=1.0.0, <=2.0.0
```
jklaise, line 5: Strict upper bounds for these two should be upheld as equality was a mistake/typo.

jklaise (approved):

doc/source/conf.py
ascillitoe: I completely missed this difference. It seems they are in fact two separate extensions: https://github.com/tox-dev/sphinx-autodoc-typehints/issues/167.

ascillitoe: Comparision of the outputs from both here:
https://github.com/tox-dev/sphinx-autodoc-typehints/issues/167#issuecomment-815968866

jklaise: Ah nice, yes the 3rd party is superior, good to keep it.

PR #499, opened 2022-05-06
Install alibi-detect during RTD docs build

ascillitoe:
With the ReadTheDocs V2 config file it appears we can install `alibi-detect` (and its dependencies) as part of the docs build, so that they don't have to be mocked. This is a bit of a waste of compute whilst we have heavy core deps like `tensorflow`, but once the optional deps project is complete we can investigate installing `alibi-detect` on RTD and only mocking heavy optional deps (e.g. `tensorflow`, `torch`). This would ease pain points experienced when building the API docs involving some mocked modules such as `pydantic`. 

Opening this PR now to come back to once the optional deps project is complete.

If going ahead with this we'd likely have to configure `:inherited-members:` to prevent docstrings of inherited modules being included when we don't want them to e.g. see example [here](https://seldon--499.org.readthedocs.build/projects/alibi-detect/en/499/api/alibi_detect.models.tensorflow.html#alibi_detect.models.tensorflow.AE).

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -25,5 +25,3 @@ python:
     - requirements: requirements/docs.txt
     - method: pip
       path: .
-      extra_requirements:
-        - torch
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -13,13 +13,17 @@ build:
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
-   configuration: doc/source/conf.py
+  configuration: doc/source/conf.py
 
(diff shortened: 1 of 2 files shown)


ascillitoe:
This change will also allow us to [https://autodoc-pydantic.readthedocs.io/en/stable/index.html](https://autodoc-pydantic.readthedocs.io/en/stable/index.html) to prettify our pydantic model api docs. This package doesn't appear to work properly if `pydantic`, or any modules used as types in model fields, are mocked..

Issue #493, opened 2022-04-27
Factor out PyTorch device setting


arnaudvl:
Various detectors set `torch.device` as follows:

```python
       # set backend
        if device is None or device.lower() in ['gpu', 'cuda']:
            self.device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
            if self.device.type == 'cpu':
                print('No GPU detected, fall back on CPU.')
        else:
            self.device = torch.device('cpu')
```
As a result, this can be factored out in a simple utility function.

PR #503, opened 2022-05-13
Factor out PyTorch device setting

kuutsav:
Addresses #493

### DIFF
--- a/alibi_detect/cd/pytorch/context_aware.py
+++ b/alibi_detect/cd/pytorch/context_aware.py
@@ -3,6 +3,7 @@ import numpy as np
 import torch
 from typing import Callable, Dict, Optional, Tuple, Union
 from alibi_detect.cd.base import BaseContextMMDDrift
+from alibi_detect.cd.pytorch.utils import get_torch_device
 from alibi_detect.utils.pytorch.kernels import GaussianRBF
 from alibi_detect.cd._domain_clf import _SVCDomainClf
 from tqdm import tqdm
@@ -94,12 +95,7 @@ class ContextMMDDriftTorch(BaseContextMMDDrift):
         self.meta.update({'backend': 'pytorch'})
 
         # set device
-        if device is None or device.lower() in ['gpu', 'cuda']:
-            self.device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
-            if self.device.type == 'cpu':
-                print('No GPU detected, fall back on CPU.')
-        else:
-            self.device = torch.device('cpu')
+        self.device = get_torch_device()
 
         # initialize kernel
         self.x_kernel = x_kernel(init_sigma_fn=_sigma_median_diag) if x_kernel == GaussianRBF else x_kernel
--- a/alibi_detect/cd/pytorch/context_aware.py
+++ b/alibi_detect/cd/pytorch/context_aware.py
@@ -95,7 +95,7 @@ class ContextMMDDriftTorch(BaseContextMMDDrift):
         self.meta.update({'backend': 'pytorch'})
 
         # set device
-        self.device = get_torch_device()
+        self.device = get_torch_device(device)
 
         # initialize kernel
         self.x_kernel = x_kernel(init_sigma_fn=_sigma_median_diag) if x_kernel == GaussianRBF else x_kernel
--- a/alibi_detect/cd/pytorch/context_aware.py
+++ b/alibi_detect/cd/pytorch/context_aware.py
@@ -3,7 +3,7 @@ import numpy as np
 import torch
 from typing import Callable, Dict, Optional, Tuple, Union
 from alibi_detect.cd.base import BaseContextMMDDrift
-from alibi_detect.cd.pytorch.utils import get_torch_device
+from alibi_detect.utils.pytorch import get_device
 from alibi_detect.utils.pytorch.kernels import GaussianRBF
 from alibi_detect.cd._domain_clf import _SVCDomainClf
 from tqdm import tqdm
@@ -95,7 +95,7 @@ class ContextMMDDriftTorch(BaseContextMMDDrift):
         self.meta.update({'backend': 'pytorch'})
 
         # set device
-        self.device = get_torch_device(device)
+        self.device = get_device(device)
 
         # initialize kernel
         self.x_kernel = x_kernel(init_sigma_fn=_sigma_median_diag) if x_kernel == GaussianRBF else x_kernel
--- a/alibi_detect/cd/pytorch/lsdd.py
+++ b/alibi_detect/cd/pytorch/lsdd.py
@@ -77,7 +77,7 @@ class LSDDDriftTorch(BaseLSDDDrift):
         self.meta.update({'backend': 'pytorch'})
 
         # set backend
-        self.device = get_torch_device()
+        self.device = get_torch_device(device)
 
         # TODO: TBD: the several type:ignore's below are because x_ref is typed as an np.ndarray
         #  in the method signature, so we can't cast it to torch.Tensor unless we change the signature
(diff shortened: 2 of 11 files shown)


ascillitoe:
Hi @kuutsav, welcome! and thanks for contributing 🙂

alibi_detect/cd/pytorch/lsdd_online.py
```diff
@@ -1,8 +1,10 @@
+from webbrowser import get
```
ascillitoe, line 1: Accidental addition?

kuutsav: Yes, vscode is playing tricks with me. Let me remove it.

ascillitoe (changes requested):
Thanks for this @kuutsav. The overall strategy looks good to me (assuming you are OK with the `get_torch_device` function name and location @arnaudvl?).

There are a few issues that need addressing though:
- The `flake8` linting issues. See `CONTRIBUTING.md` for more guidance, but essentially if you run `pip install -r requirements/dev.txt` you can then run `flake8 alibi_detect` and `mypy alibi_detect` to check these both pass locally before pushing.
- The new function can also be added to `pytorch/learned_kernel.py` and `pytorch/classifier.py`. (You've correctly left out `pytorch/spot_the_diff.py`, the logic is different for this one).

alibi_detect/cd/pytorch/lsdd_online.py
ascillitoe: Ah good old vscode!

alibi_detect/cd/pytorch/context_aware.py
```diff
@@ -94,12 +95,7 @@ def __init__(
         self.meta.update({'backend': 'pytorch'})
 
         # set device
-        if device is None or device.lower() in ['gpu', 'cuda']:
-            self.device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
-            if self.device.type == 'cpu':
-                print('No GPU detected, fall back on CPU.')
-        else:
-            self.device = torch.device('cpu')
+        self.device = get_torch_device()
```
arnaudvl, line 98: This would have to be changed to:
```python
self.device = get_torch_device(device)
```
since `device` is a kwarg of the detector. Similar for the other detectors.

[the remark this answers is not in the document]
ascillitoe: Good point!

arnaudvl:
Thanks for the contribution @kuutsav . Just left a small comment which is applicable to the various detectors.

alibi_detect/cd/pytorch/context_aware.py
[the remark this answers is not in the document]
kuutsav: Good catch! added

kuutsav:
> Thanks for the contribution @kuutsav . Just left a small comment which is applicable to the various detectors.

Have addressed the issue along with the linting issue. Have also added the method at `pytorch/learned_kernel.py` and `pytorch/classifier.py`.

Do we want to print or log here?
`print('No GPU detected, fall back on CPU.')`

ascillitoe:
> Do we want to print or log here?
> `print('No GPU detected, fall back on CPU.')`

Good question. IMO we should still with `logging`.

alibi_detect/cd/pytorch/lsdd.py
```diff
@@ -76,12 +77,7 @@ def __init__(
         self.meta.update({'backend': 'pytorch'})
 
         # set backend
```
ascillitoe, line 79: @kuutsav this isn't something you've introduced, but since you're already making changes to these sections, would you be able to change the `# set backend` to `# set device` here please?

alibi_detect/cd/pytorch/lsdd_online.py
```diff
@@ -80,12 +81,7 @@ def __init__(
         self.lambda_rd_max = lambda_rd_max
 
         # set backend
```
ascillitoe, line 83: Same here: `# set backend` -> `# set device`.

alibi_detect/cd/pytorch/utils.py
```diff
@@ -1,5 +1,10 @@
+import logging
```
arnaudvl, line 1: Even though the `get_torch_device` utility is currently only used for drift detectors, it would probably be used for outlier/adversarial detectors too as we'll be extending them to the PyTorch backend. Therefore it would make more sense to add `get_torch_device` to `alibi_detect.utils.pytorch` instead. There isn't a perfect "fit" for it yet but `misc.py` could make sense @ascillitoe ? Moreover, it could be fine to just call it `get_device` since it's imported from `alibi_detect.utils.pytorch` anyway. But that's a very tiny detail! :)

ascillitoe: So `get_device` in `alibi_detect.utils.pytorch`? 🙂

arnaudvl:
One more place where it makes sense to use this utility function is the [`alibi_detect.utils.pytorch.predict_batch`](https://github.com/SeldonIO/alibi-detect/blob/de4caa3b96a265007a14fd606b3294fa47c4481f/alibi_detect/utils/pytorch/prediction.py#L37) functionality. Note that the input argument type hint would need to be updated as well. Besides my comments, looks good!

alibi_detect/cd/pytorch/utils.py
kuutsav: I can make this change but i feel we should adopt a principle like YAGNI/KISS here and leave it as is till we are actually there.

kuutsav:
> One more place where it makes sense to use this utility function is the [`alibi_detect.utils.pytorch.predict_batch`](https://github.com/SeldonIO/alibi-detect/blob/de4caa3b96a265007a14fd606b3294fa47c4481f/alibi_detect/utils/pytorch/prediction.py#L37) functionality. Note that the input argument type hint would need to be updated as well. Besides my comments, looks good!

Made this change as well. Had to ignore the types at couple of places as mypy was complaining due to reassignment of `device`.

alibi_detect/cd/pytorch/utils.py
```diff
@@ -1,6 +1,7 @@
-from torch import nn
 from typing import Callable
 
+from torch import nn
```
ascillitoe, line 3: Best to revert this change since this file no longer needs to be changed.

alibi_detect/utils/pytorch/prediction.py
```diff
@@ -54,11 +54,11 @@ def predict_batch(x: Union[list, np.ndarray, torch.Tensor], model: Union[Callabl
                     preds = tuple([] for _ in range(len(preds_tmp)))
                     return_list = isinstance(preds_tmp, list)
                 for j, p in enumerate(preds_tmp):
-                    if device.type == 'cuda' and isinstance(p, torch.Tensor):
+                    if device.type == 'cuda' and isinstance(p, torch.Tensor):  # type: ignore
```
ascillitoe, line 57: What `mypy` error is raised here? It would be good to add an error code specific type ignore so that it is clear what is being ignored `# type: ignore[<error-code>]`. (this is something we're trying to do more going forward, as it leads to more readable code).

ascillitoe:
Sorry @kuutsav, after looking at this a bit more (after the most recent `mypy` errors), I've realised the current changes to `utils/pytorch/prediction.py` are not quite right. 

Changing `device: torch.device` to `device: Optional[str]` will raise more issues since `device` is type `torch.device` in `predict_batch_transformer` (which calls `predict_batch`), and in `cd.preprocess.preprocess_drift` (which calls both of these functions). 

Instead of adding multiple (i.e. 5 or 6) `type: ignore`'s, I've instead tweaked `get_device` so that it can take in a `torch.device` and simply return it untouched, for compatibility with `predict_batch`. @arnaudvl what do you think? I'll wait to get your thoughts before reviewing again...

arnaudvl (approved):

ascillitoe:
Congrats (and thanks) on your first contribution @kuutsav! 🎉

PR #505, opened 2022-05-16
Submodule to compute and visualise drift detection metrics

ascillitoe:
The `alibi_detect.cd.metrics` submodule implements utility functions to measure and visualize detector calibration and test power. These can be used for standalone experiments, and will also be used for our internal benchmarking.

How we check sufficient data has been provided in `_check_sufficient_data_size` is still an open question...

Example notebook:
https://gist.github.com/ascillitoe/4a23e156adbff72a3314d78846fbff83


alibi_detect/cd/metrics.py
```diff
@@ -0,0 +1,333 @@
...
+        for the detector. Should only be given for the :class:`~alibi_detect.cd.ContextMMDDrift` detector.
+    model
+        A PyTorch, TensorFlow or Sklearn model. Should only be given for detectors accepting a `model` arg, such
+        as :class:`~alibi_detect.cd.ClassifierDrift`.
+    detector_kwargs
+        Keyword arguments to pass to detector.
+    n_runs
+        The number of experiment runs. A larger number will give more stable p-value distributions, at the expense of
+        a longer runtime.
+    n_samples
```
ojcobb, line 44: Could we perhaps have separate args for ref and test here? Personally if I was using it I'd want n_test_samples = W and n_ref_samples = N-W

alibi_detect/cd/metrics.py
```diff
@@ -0,0 +1,333 @@
...
+
+        if model is not None:
+            detector_args.append(model)
+
+        # Init detector and predict
+        dd = detector(*detector_args, **detector_kwargs)
+        preds = dd.predict(*predict_args)
+        p_vals_list.append(preds['data']['p_val'])
+    p_vals = np.array(p_vals_list).flatten()
+    # TODO - is above OK? (to flatten over features for univariate detectors)
```
ojcobb, line 100: If so I think we should specify the calibration is at the feature rather than detector level. Alternatively we could have an arg corresponding to the correction and show compute the calibration at the detector level.

alibi_detect/cd/metrics.py
```diff
@@ -0,0 +1,333 @@
...
+
+    # QQ-plot
+    if qq_plot:
+        plot_qq(p_vals)
+    # Histogram plot
+    if hist_plot:
+        plot_hist(p_vals)
+    # Show plots
+    if qq_plot or hist_plot:
+        plt.show()
```
ojcobb, line 110: Worth allowing plots to be saved for those not using notebooks?

alibi_detect/cd/metrics.py
```diff
@@ -0,0 +1,333 @@
...
+            preds = dd.predict(*predict_args)
+            power[i] += preds['data']['is_drift']
+    power /= n_runs
+
+    if power_plot:
+        plot_power(sig_levels, power)
+        plt.show()
+
+    if return_auc:
+        auc = compute_auc(sig_levels, power)
```
ojcobb, line 189: Whilst this is, of course, *an* area under a curve, this is not AUC in the colloquial power-measuring sense (which is the area under a plot of FPRs vs TPRs at different significance levels). We should therefore use a different name or compute the AUC. I actually think both can be useful, although computing the true AUC of course additionally requires obtaining false positive rates.

alibi_detect/cd/metrics.py
```diff
@@ -0,0 +1,333 @@
...
+
+def eval_test_power(
+    detector: type,  # TODO: type w/ BaseDriftDetector if introduced.
+    X_ref: Union[np.ndarray, list],
+    X_drift: Union[np.ndarray, list],
+    C_ref: Optional[np.ndarray] = None,
+    C_drift: Optional[np.ndarray] = None,
+    model: Optional[Callable] = None,
+    detector_kwargs: Optional[dict] = None,
+    sig_levels: np.ndarray = np.linspace(0.05, 0.5, 10),
```
ojcobb, line 123: Not sure about this as a default. Maybe something more similar to np.linspace(0.01, 0.1, 10)?

ojcobb (commented):
Nice work - plots look great! Only minor comments except for the AUC one which needs sorting.

alibi_detect/cd/metrics.py
ascillitoe: `n_samples` is now a tuple `(n_ref_samples, n_test_samples)` like in `eval_test_power`. Does that work?

alibi_detect/cd/metrics.py
ascillitoe: Done!

alibi_detect/cd/metrics.py
ascillitoe: Good spot, thanks. I've renamed `return_auc` in `eval_test_power` to `return_power_auc`. We could even rename further if you think that's still confusing?

`eval_roc` now returns the true AUC.

ascillitoe:
Thanks for your comments @ojcobb, I've had a first go at addressing them.

One interesting thing to note is that in order to save some compute, in `eval_roc` I'm only calling `eval_calibration` for one significance level, and then compute the FPR's for each significance level using the same bunch of `p_vals`. Can you see any reason this is a bad idea?

alibi_detect/cd/metrics.py
ojcobb: Sounds good

alibi_detect/cd/metrics.py
ojcobb: I was thinking more along the lines of return the p-values as given when there is a p-value per feature and optionally allowing them to aggregate those multiple p-values using a correction to obtain instance level p-values.

ascillitoe: Thanks @ojcobb. So have `correction=None` as default and return the p-values as given (i.e. per feature) in that instance. Otherwise, aggregate to give instance-level p-values if `correction` is given?

Issue #518, opened 2022-05-26
Tests for `ClassifierDrift` with 'sklearn' backend


ascillitoe:
The `ClassifierDrift` detector has tests for all the backend-specific subclasses (in `cd/<backend>/tests/test_classifier_<backend>.py`). There are then tests for `ClassifierDrift` itself in `cd/tests/test_classifier.py`, which test that `ClassifierDrift` correctly dispatches the correct subclass depending on the value of the `backend` kwarg. However, the new `sklearn` backend is missing from these top-level tests.

PR #519, opened 2022-05-26
Minor fixes for config driven detectors

ascillitoe:
This PR fixes a number of issues uncovered in final testing:

- `input_shape` kwarg missing from `ClassifierDriftSklearn`. Discovered by notebook tests as relevent test missing from CI (see #518).
- Issue with `save_detector` in legacy mode when save directory not already created.
- Issue with `NDArray` not working on Python<3.9 with v. new NumPy (`1.22.4`) installed.


ascillitoe:
The build is now failing due to a `mypy` error with `ContextMMDDriftTorch` (`Tensor` is indexed with a NumPy array). This appears to be occurring now when new `mypy` (0.960) and `numpy` (1.22.4) versions are both installed. No idea why this `mypy` and `numpy` combination is being installed here but not in `master` CI.

We should probably refactor `ContextMMDDriftTorch` slightly in any case...

PR #537, opened 2022-06-21
Add optional dependency management

mauicv:
# TODO:

- [x] [feature branch](https://github.com/SeldonIO/alibi-detect/pull/538) for tensorflow and pytorch optional dependencies
- [ ] ~~feature branch for numba optional dependencies~~
- [x] [feature branch](https://github.com/SeldonIO/alibi-detect/pull/541) for prophet optional dependencies
- [ ] ~~Image branch (see [notion](https://www.notion.so/Alibi-detect-Solution-Note-2dfdfea2e6954dafa93e39f7e40f7021))~~
- [x] [docs](https://github.com/SeldonIO/alibi-detect/pull/542) branch
- [x] [Update Licenses](https://github.com/SeldonIO/alibi-detect/pull/562)
___

# What is this PR?

This PR addresses optional dependency import functionality for Alibi Detect. The core behaviour this adds is to throw errors informing the user if they haven't installed the necessary optional dependency for the functionality they wish to use. These errors are thrown at use time rather than at import.

There are three behaviours in general:

1. Optional dependency in a public object. In this case, we might have an object that can optionally use an optional dependency.
    - In this case, if the object is used in a way that doesn't use this dependency no error should be thrown. 
    - If The object is used such that the optional dependency is used then the error should be thrown.
2. Conditionally Optional dependency in an object. As an example, the value `backend` should be one of `'tensorflow'` or `'torch'` but not neither.
    - An error should be thrown if the user requests a backend that isn't installed.
3. Objects that are completely dependent on an optional dependency
    - If the user imports and uses these then an error should be thrown

**This PR addresses behaviours 1 and 3 above.**
___

# Note:
In order to manually test alibi-detect in each of the different environments developers can use:

```sh
make repl tox-env=<env>
```
and this will give them a REPL with the `<env>` optional dependencies installed. 

For example: `make repl tox-env=torch` will give them the torch optional dependency environment REPL.
___

The main changes are:

1. The MissingDependency and optional_import functionality, 
2. The refactoring of the codebase 
3. The tests.

## 1. MissingDependency

The MissingDependency metaclass is used to allow imports that have missing dependencies. If these imports are used later they will throw an error.

### Implementation

The basic pattern is to replace constructs that require uninstalled dependencies with something like:

```py
class MissingDependency:
    def __init__(self, missing_dependency: str, object_name: str):
        self.object_name = object_name
        self.missing_dependency = missing_dependency

    def __getattr__(self, key):
        raise ImportError()

    def __call__(self, *args, **kwargs):
        raise ImportError()
```

We also use an import function which should be used throughout by developers when importing constructs dependent on optional dependencies in order to ensure consistent behaviour. This looks roughly like:

```py
def import_optional(module_name: str):
    try:
        return import_module(module_name)
    except (ImportError, ModuleNotFoundError) as err:
        return MissingDependency(
            missing_dependency=err.name, 
            object_name=module_name)
```

The above is called with:

```py
UsefulClass = import_optional('alibi_detect.utils.useful_class')
```

The above raises an error if the user attempts to access an attribute or call the object. 
1. The error message informs the user that the `UsefulClass` object is missing optional dependencies. 
4. It tells the user how to resolve using `pip install alibi-detect[optional-dependency]`. 
5. And finally it links to the original error thrown.  

## 2. Refactoring:

The idea here is wherever the is an object that is dependent on an optional install the relevant functionality should be fenced off into a single file or collection of files. Access to that object should be via the `__init__.py` file. Within the `__init__.py` we will implement the `MissingDependency` pattern mentioned above.

### Note on Type checking issue:

- The `import_optional` function returns an object instance rather than an object. This will cause typechecking to fail 
if not all optional dependencies are installed. Because of this we also need to 1. Conditionally import the true object
dependent on `TYPE_CHECKING` and 2. Use forward referencing within typing constructs such as `Union`. We use forward 
referencing because in a user environment the optional dependency may not be installed in which case it'll be replaced 
with an instance of the MissingDependency class. This will throw an error when passed to `Union`. See `CONTRIBUTING.md` note for more details and example.

## 3. Tests:

The tests import all the named objects from the public API of alibi and test that they throw the correct errors if the relevant optional dependencies are not installed. If these tests fail, it is likely that:

1. The optional dependency relation hasn't been added to the test script. In this case, this test assumes that your
functionality should work for the default alibi-detect install. If this is not the case you should add the exported object name to the `dependency_map` in the relevant test.
6. The relevant export in the public API hasn't been protected with the `MissingDependency` class. In this case, see the docs string for the `utils.missing_dependency` m1odule.

## Notes:
1. The tests will be skipped in the normal test suite. To run correctly use tox. This should be as simple as running `tox` from the terminal. The environments will take a long time to build the first time but afterwards, this should be a quick process.
2. If you need to configure a new optional dependency you will need to update the `setup.cfg` file and add a `testenv` environment as well as to the `extra-requires.txt` file
7. Backend functionality may be unique to specific explainers/functions and so there may be multiple such modules that need to be tested separately.
8. We assume all public functionality is exposed in modules via the `__init__.py` files.
9. We assume all imports are top-level, if an import is nested within an object or function call this functionality will avoid being caught by the tests.

___

### See also [Further Notes](https://www.notion.so/Alibi-detect-Solution-Note-2dfdfea2e6954dafa93e39f7e40f7021)


README.md
```diff
@@ -117,7 +117,7 @@ We will use the [VAE outlier detector](https://docs.seldon.io/projects/alibi-det
 
 ```python
 from alibi_detect.od import OutlierVAE
-from alibi_detect.utils import save_detector, load_detector
+from alibi_detect.utils.saving import save_detector, load_detector
```
jklaise, line 120: Can be changed to use the public api `alibi_detect.saving`, although not strictly part of this PR (can be a different one).

alibi_detect/utils/missing_optional_dependency.py
```diff
@@ -0,0 +1,118 @@
...
+                 missing_dependency: str = 'all',):
+        """ Metaclass for MissingDependency classes
+        Parameters
+        ----------
+        object_name
+            Name of object we are replacing
+        missing_dependency
+            Name of missing dependency required for object
+        err
+            Error to be raised when the class is initialized or used
```
jklaise, line 49: Reorder docstring to align with the function signature.

alibi_detect/utils/missing_optional_dependency.py
```diff
@@ -0,0 +1,118 @@
...
+        module_name
+            The module to import
+        names
+            The names to import from the module. If None, all names are imported.
+    Returns
+    -------
+        The module or named objects within the modules if names is not None. If the import fails due to a
+        ModuleNotFoundError or ImportError then the requested module or named objects are replaced with instances of
+        the MissingDependency class above.
+    """
```
jklaise, line 87: Nitpick: I don't think numpy style docstrings give you indentend parameter lists (e.g. compare with `MissingDependency' class above)

alibi_detect/utils/frameworks.py
```diff
@@ -15,3 +18,91 @@
...
+
+def _iter_to_str(iterable):
+    """ Correctly format iterable of items to comma seperated sentence string."""
+    items = [f'`{option}`' for option in iterable]
+    last_item_str = f'{items[-1]}' if not items[:-1] else f' and {items[-1]}'
+    return ', '.join(items[:-1]) + last_item_str
+
+
+class BackendValidator:
+    def __init__(self, backend_options: Dict[Optional[str], List[str]], construct_name: str):
```
jklaise, line 38: btw I'm not sure `mypy` actually typechecks functions/methods with missing return type hints, can you verify and add `None` return types where necessary?

alibi_detect/utils/tests/test_backend_verify.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # If backend is an option, but multiple dependencies are missing ensure ImportError raised by BackendValidator
+        # including error message listing missing dependencies.
+        with pytest.raises(ImportError) as err:
+            BackendValidator(backend_options=options, construct_name='test').verify_backend('backend_3')
+        assert 'backend_D' in str(err.value).split('.')[0]
+        assert 'backend_E' in str(err.value).split('.')[0]
+        assert 'backend_3' in str(err.value)
+        assert 'backend_D' in str(err.value).split('pip install alibi-detect')[1]
+        assert 'backend_E' in str(err.value).split('pip install alibi-detect')[1]
```
jklaise, line 67: Any reason not to compare to the full string here?

alibi_detect/cd/classifier.py
```diff
@@ -147,12 +148,12 @@ def __init__(
...
-        if (backend == 'tensorflow' and not has_tensorflow) or (backend == 'pytorch' and not has_pytorch) or \
-                (backend == 'sklearn' and not has_sklearn):
-            raise ImportError(f'{backend} not installed. Cannot initialize and run the '
-                              f'ClassifierDrift detector with {backend} backend.')
-        elif backend not in ['tensorflow', 'pytorch', 'sklearn']:
-            raise NotImplementedError(f"{backend} not implemented. Use 'tensorflow', 'pytorch' or 'sklearn' instead.")
+        BackendValidator(
+            backend_options={'tensorflow': ['tensorflow'],
+                             'pytorch': ['pytorch'],
+                             'sklearn': ['sklearn']},
```
jklaise, line 154: Similarly `'sklearn': ['scikit-learn']`? Although I tested this and it seems they aliased `sklearn` to pull in `scikit-learn`.

jklaise: In fact, `scikit-learn` is a core dependency, not sure if this check is even here.

alibi_detect/cd/classifier.py
```diff
@@ -147,12 +148,12 @@ def __init__(
...
+            construct_name='ClassifierDrift'
```
jklaise, line 155: Could we get the class name dynamically here? Concerned that due to spelling mistakes the error message might be incorrect (although I think it should still validate the backends correctly).

alibi_detect/utils/saving/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+# TODO: config-driven-detectors moves this module out of utils. So
+#  will require refactoring!
+
+
+from alibi_detect.utils.missing_optional_dependency import import_optional
+
+save_detector, load_detector = import_optional(
+    'alibi_detect.utils.saving.saving',
+    names=['save_detector', 'load_detector']
+)
```
jklaise, line 17: Since these depend on optional dependencies (presumably tf/torch?) does that mean people can't save/load detectors that don't depend on optional dependencies? @ascillitoe @mauicv

alibi_detect/tests/test_dep_management.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+def test_cd_dependencies(opt_dep):
+    """Tests that the cd module correctly protects against uninstalled optional dependencies.
+    """
+
+    dependency_map = defaultdict(lambda: ['default'])
+    for dependency, relations in []:
+        dependency_map[dependency] = relations
+    if opt_dep != 'all':
```
jklaise, line 59: Noting that this conditional is missing in the alibi implementations of the test functions (https://github.com/SeldonIO/alibi/blob/master/alibi/tests/test_dep_management.py), what's the reason for having this here?

alibi_detect/utils/saving/__init__.py
ascillitoe: Mmmmn good point, and important to clarify I think. I originally designed the `saving` submodule with all "backend" functionality in submodules (i.e. `saving.tensorflow._saving.py`). The intention behind this is that the optional deps imports would be done in `saving.saving.py`, and `save_detector` would be imported with a standard import. That way `save_detector` and `load_detector` should be usable regardless of which optional deps are installed. 

I believe the above code snippet is a little different. This is importing the legacy `save_detector` and `load_detector`, which are in fact dependent on `tensorflow`. The import strategy for the new `save_detector` and `load_detector` should match my description above hopefully. I think we did test that `save_detector` works when `tensorflow` is not installed. @mauicv can you clarify? 

P.s. the above might all be a little confusing since many of the changes wrt to saving/loading will only be incorporated properly in #553.

alibi_detect/utils/missing_optional_dependency.py
```diff
@@ -0,0 +1,118 @@
...
+        if err.name is None:
+            raise TypeError()
+        if str(err.name) not in ERROR_TYPES:
+            raise err
+
+        missing_dependency = None
+        if not missing_dependency:
+            missing_dependency = ERROR_TYPES[err.name]
+
+        if names:
```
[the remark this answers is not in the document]
mauicv: Yep, this is functionality that was later [removed](https://github.com/SeldonIO/alibi-detect/commit/0b8f86904bca87b8ecd228acd780ca7130830f54). I've changed it to be in sync with alibi implimentation.

alibi_detect/utils/frameworks.py
mauicv: My understanding is that if a method is missing all type hints, so for arguments and return types then `mypy` doesn't type check it. So if `__init__` doesn't require any parameters then you have to specify the return type as `None` or it won't be checked. This isn't an issue here but might be elsewhere. I'll make sure to change it where necessary.

alibi_detect/utils/tests/test_backend_verify.py
mauicv: The order the dependencies were listed in `err.value`  could change. I've changed them to be deterministic.

alibi_detect/cd/classifier.py
```diff
@@ -147,12 +148,12 @@ def __init__(
...
         backend = backend.lower()
-        if (backend == 'tensorflow' and not has_tensorflow) or (backend == 'pytorch' and not has_pytorch) or \
-                (backend == 'sklearn' and not has_sklearn):
-            raise ImportError(f'{backend} not installed. Cannot initialize and run the '
-                              f'ClassifierDrift detector with {backend} backend.')
-        elif backend not in ['tensorflow', 'pytorch', 'sklearn']:
-            raise NotImplementedError(f"{backend} not implemented. Use 'tensorflow', 'pytorch' or 'sklearn' instead.")
+        BackendValidator(
+            backend_options={'tensorflow': ['tensorflow'],
+                             'pytorch': ['pytorch'],
```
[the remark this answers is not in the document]
mauicv: The dependencies' presence is checked by the `HAS_BACKEND` `dict` which has key `pytorch`. The `pip install` message on the other hand is generated from the  `ERROR_TYPES` which maps to `torch` which is also the name of our dependency bucket. So the error message is:

```
ImportError: `pytorch` not installed. Cannot initialize and run LSDDDrift with pytorch backend. The necessary missing dependencies can be installed using `pip install alibi-detect[torch]`.
```

I kept this in line with the error messages that were being generated prior to the addition of the BackendValidator.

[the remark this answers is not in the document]
jklaise: ok makes sense, but probably would be great to add some comments in the code around this

alibi_detect/cd/classifier.py
mauicv: I copied the functionality from [here](https://github.com/SeldonIO/alibi-detect/blob/6e4f5e0c86e562d0a92bb48d0a9690de0606a37b/alibi_detect/cd/classifier.py#L150) and [here](https://github.com/SeldonIO/alibi-detect/blob/6e4f5e0c86e562d0a92bb48d0a9690de0606a37b/alibi_detect/utils/frameworks.py#L15). But yeah, I don't think it makes sense to have the full import check so I've just set `HAS_BACKEND['sklearnn] = True`. I think It is simpler just to keep the general logic in place even though it's redundant. Although happy to change?

mauicv: Just that [here](https://github.com/SeldonIO/alibi-detect/blob/6e4f5e0c86e562d0a92bb48d0a9690de0606a37b/alibi_detect/utils/frameworks.py#L15) we try and import `sklearn` and then set `has_sklearn` conditional on the result which is redundant as `sklearn` should always be there!

alibi_detect/tests/test_dep_management.py
mauicv: Yeah, I'm not sure why I put this here but it's not necessary. I've removed!

alibi_detect/utils/saving/__init__.py
ascillitoe: My vote would be to not touch legacy saving, and just keep it `tensorflow` dependent. Removing the tf dep would involve significant reworking, which I don't think is appropriate for legacy functionality.

jklaise (approved):
LGTM

alibi_detect/cd/classifier.py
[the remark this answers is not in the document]
ascillitoe: @mauicv did you add the comments @jklaise is referring to here?

alibi_detect/tests/test_dep_management.py
```diff
@@ -0,0 +1,266 @@
...
+    that need to be tested separately.
+"""
+
+from types import ModuleType
+from collections import defaultdict
+
+import pytest
+
+
+def check_correct_dependencies(
```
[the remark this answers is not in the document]
ascillitoe: Should we open an issue for this?

alibi_detect/tests/test_dep_management.py
```diff
@@ -0,0 +1,274 @@
...
+            ("zero_diag", ['tensorflow']),
+            ("mutate_categorical", ['tensorflow']),
+            ("TFDataset", ['tensorflow'])
+            ]:
+        dependency_map[dependency] = relations
+    from alibi_detect.utils import tensorflow as tensorflow_utils
+    check_correct_dependencies(tensorflow_utils, dependency_map, opt_dep)
+
+
+def test_pytorch_utils_dependencies(opt_dep):
```
ascillitoe, line 252: The other test functions are all named `test_torch_...` rather than `test_pytorch_...`. Change for consistency?

alibi_detect/utils/missing_optional_dependency.py
```diff
@@ -0,0 +1,114 @@
+"""Functionality for optional importing
+This module provides a way to import optional dependencies. In the case that the user imports some functionality from
+alibi that is not usable due to missing optional dependencies this code is used to allow the import but replace it
```
ascillitoe, line 3: Change `alibi` to `alibi_detect`? Maybe worth a quick grep to check if any other places where comments have been copied from the `alibi` implementation?

ascillitoe:
Commenting here as can't comment on empty file. For `alibi_detect/utils/tests/__init__.py`, is it worth adding a comment to the file explaining why we need an empty `__init__.py`? Assuming its for duplicate `conftest.py`'s...

ascillitoe (approved):
Few final comments. LGTM once resolved!

Before merging we should remove [WIP] from the PR name, and decide on whether to squash before merging. @jklaise @mauicv

alibi_detect/cd/classifier.py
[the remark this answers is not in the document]
mauicv: I've updated the comment [here](https://github.com/SeldonIO/alibi-detect/blob/9c31d665074fd479b14d399af6cb235a64c094aa/alibi_detect/utils/frameworks.py#L41) to be clearer about the behaviour!

alibi_detect/tests/test_dep_management.py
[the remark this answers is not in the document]
mauicv: Opened an issue [here](https://github.com/SeldonIO/alibi-detect/issues/566)

bluepark-sk:
Hi. Do you have any plan to add saving/loading functionality for PyTorch model/backend?
If you have, when will it be developed?

ascillitoe:
> Hi. Do you have any plan to add saving/loading functionality for PyTorch model/backend? If you have, when will it be developed?

Hi @bluepark-sk, we sure do. Time permitting, I will be beginning work on PyTorch save/load this week (for drift detectors only)!

bluepark-sk:
@ascillitoe Thank you! I'll be waiting~~

Issue #443, opened 2022-02-02
Conda install stuck on "Solving environment"


ascillitoe:
Our new conda install gets stuck on "Solving environment". We can still install from conda-forge with `mamba` (see https://github.com/SeldonIO/alibi-detect/issues/436#issuecomment-1026663095), but we should still work on addressing the above for maximum compatibility. 

Making deps such as opencv and TensorFlow may help here, but we might also need to adjust deps version requirements...

candalfigomoro:
Moreover, by default the tensorflow requirement usually brings the GPU version of tensorflow which has very heavy dependencies such as the cuda toolkit

ascillitoe:
Additional sub-issues to consider:

- [ ] The `transformers` recipe includes `pytorch`, therefore our recipe implicitly includes `pytorch` and it cannot be optional. We should add `pytorch` explicitly to the `run` list to constrain the versions.
- [ ] See #537 for more discussion of optional deps related to conda. Any optional deps not included in the recipe should be added to the `run_constrained` list (e.g. `fbprophet`).
- [ ] As @candalfigomoro mentioned above, we should consider whether to include `tensorflow` in our recipe since it is soon to be an optional deps for `alibi-detect`. This would allow the user to chose between `tensorflow-cpu` or `tensorflow-gpu`, although I wonder if the same is achieved via `CONDA_CUDA_OVERRIDE` (see [here](https://conda-forge.org/blog/posts/2021-11-03-tensorflow-gpu/#installation)).

PR #538, opened 2022-06-21
Feature Optional backends functionality

mauicv:
# Todo:
- [x] Document BackendValidator
- [x] Tox env tests
- [x] Document BackendValidator tests
- [x] Update Readme
- [x] Update docs/notebooks
- [x] Comments for large file changes
___

## What is this

This work branches off [`feature/dep-management`](https://github.com/SeldonIO/alibi-detect/pull/501) and will make `tensorflow`, `pytorch` and `tensorflow_probability` optional dependencies of `alibi-detect`.

Two main objectives are:

## 1. Make `tensorflow`, `pytorch` and `tensorflow_probability` optional dependencies:

I've combined these into this single PR because they're all quite interdependent and so It was hard to do them separately without writing code to manage functionality implemented in one but not the others.

## 2. Add BackendValidation for all detectors:

This wasn't such an issue in alibi as only CFRL implemented multiple backends. In `detect` however there are multiple cases where we have a `tensorflow`, `PyTorch` or `sklearn` backend and on top of this sometimes the tensorflow backend is dependent on `tensorflow-probability` as well. Previously this was handled using duplicated code in each detector. I've added a `BackenValidator` that replaces this and issues more specific error messages for optional dependencies. 

___

### Outstanding Questions:

1. Do we need separate dependency buckets for tensorflow and tensorflow-probability? I think in nearly all cases these are both required at the same time. (see [Notion page](https://www.notion.so/Dependency-list-4bff20fe048d4e3183ddf5e5a69a805e))?

### What this PR doesn't include:

1. The install options in the docs. I think it makes sense to do this in a later PR on the feature branch for all the optional deps all at once, once this branch has been merged in.
2. The install options in the `README.md` for same reason as 1.
3. The config-driven work in the [release branch](https://github.com/SeldonIO/alibi-detect/tree/release/0.10.0). This will be done in a separate PR in order to manage Release/Master branch complexity issues.
4. Make Licenses updates. This will be done in a PR on the feature branch.

___

### Circular imports:

I'm not sure why this wasn't an issue with alibi but in alibi-detect often we have the following scenario:
    - Import model from models
    - In importing the above model it needs a function from `utils.distance` 
    - Importing this function means the code in `utils.__init__` is executed
    - This code imports `utils.fetching` which in turn imports all the models from `models`
    - We get a circular import error

This is an issue that's introduced by us trying to bundle all the functionality into the init file and the correct approach is further scoping and modularization. The solution to the above is a `utils/fetching/__init__.py` as separate from the `utils/__init__.py`

### Changing import statements:

This PR results in lots of changes to the notebooks primarily in the form of changes to the notebooks to reflect the new intended public API. Note that these shouldn't be breaking changes as the objects in question should still be in the old locations the only difference is that where they need to be protected from `ImportErrors` we import them into the `__init__.py` file in a module instead of from the file directly. So old code will still run but the object imported won't be protected as per `import_optional`. This might mean code will break if users are running default installs as they won't have all the relevant dependencies. Any release should incorporate a note that a quick fix is updating their installs to `pip install alibi[x,y,z]` for relevant `x, y, z`. A longer fix is to update the import statements to the public API. Most of the time this won't even be necessary, just where imports have changed as:

```
from alibi_detect.utils.pytorch.kernels import DeepKernel
```

to

```
from alibi_detect.utils.pytorch import DeepKernel
```

The full list of changes is as follows:

1. `from alibi_detect.utils.tensorflow.kernels import DeepKernel` -> `from alibi_detect.utils.tensorflow import DeepKernel`
2. `from alibi_detect.utils.tensorflow.prediction import predict_batch` -> `from alibi_detect.utils.tensorflow import predict_batch`
3. `from alibi_detect.utils.pytorch.data import TorchDataset` -> `from alibi_detect.utils.pytorch import TorchDataset`
4. `from alibi_detect.models.pytorch.trainer import trainer` -> `from alibi_detect.models.pytorch import trainer`
5. `from alibi_detect.models.tensorflow.resnet import scale_by_instance` -> `from alibi_detect.models.tensorflow import scale_by_instance`
6. `from alibi_detect.models.tensorflow.resnet import scale_by_instance` -> `from alibi_detect.models.tensorflow import scale_by_instance`
7. `from alibi_detect.utils.pytorch.kernels import DeepKernel` -> `from alibi_detect.utils.pytorch import DeepKernel`
8. `from alibi_detect.models.tensorflow.autoencoder import eucl_cosim_features` -> `from alibi_detect.models.tensorflow import eucl_cosim_features`
9. `from alibi_detect.utils.tensorflow.prediction import predict_batch` -> `from alibi_detect.utils.tensorflow import predict_batch`
10. `from alibi_detect.models.tensorflow.losses import elbo` -> `from alibi_detect.models.tensorflow import elbo`
11. `from alibi_detect.models import PixelCNN` -> `from alibi_detect.models.tensorflow import PixelCNN`
12. `from alibi_detect.utils.tensorflow.data import TFDataset` -> `from alibi_detect.utils.tensorflow import TFDataset`
12. `from alibi_detect.utils.pytorch.data import TorchDataset` -> `from alibi_detect.utils.pytorch import TorchDataset`

also see [this](https://github.com/SeldonIO/alibi-detect/pull/538#discussion_r914862049) related comment!


ascillitoe:
Noting down a few thoughts:

1. **Number of optional deps groups**: The number is getting quite large, and might grow further in future. We might want to reassess the motivation behind each group. Specifically, is the `numba` group needed, and can  we merge `tensorflow` and `tensorflow-probability` (apologies, separating the last two was largely down to my recommendation). 
2. **conda optional deps groups**: We need to consider what is included (in `run`) and what is optional (in `run_constrained`) in our [conda recipe](https://github.com/conda-forge/alibi-detect-feedstock/blob/main/recipe/meta.yaml). This is made more complex by the need to consider what is included in recipes of our dependencies i.e. the `transformers` recipe includes `pytorch`, so we cannot have `pytorch` as optional in our conda recipe. Based on offline discussion with @mauicv, the latest thinking here is that we should keep our conda recipe "fat" for now (i.e. include `tensorflow` and `pytorch`), and only have `fbprophet` as an optional addition here. Looking at how to slim down the conda recipe and speed up the install should be kept as a separate task (#443).

mauicv:
w.r.t. 

> 1. **Number of optional deps groups**: The number is getting quite large, and might grow further in future. We might want to reassess the motivation behind each group. Specifically, is the `numba` group needed, and can  we merge `tensorflow` and `tensorflow-probability` (apologies, separating the last two was largely down to my initial input, which I am now thinking was maybe misguided).

Do you mean having a tensorflow bucket and a TensorFlow-probability bucket that also has tensorflow or just having a single tensorflow bucket which includes tensorflow-probability? I think I'm generally in favour of the former.

> 2. **conda optional deps groups**: We need to consider what is included (in `run`) and what is optional (in `run_constrained`) in our [conda recipe](https://github.com/conda-forge/alibi-detect-feedstock/blob/main/recipe/meta.yaml). This is made more complex by the need to consider what is included in recipes of our dependencies i.e. the `transformers` recipe includes `pytorch`, so we cannot have `pytorch` as optional in our conda recipe. Based on offline discussion with @mauicv, the latest thinking here is that we should keep our conda recipe "fat" for now (i.e. include `tensorflow` and `pytorch`), and only have `fbprophet` as an optional addition here. Looking at how to slim down the conda recipe and speed up the install should be kept as a separate task [Conda install stuck on "Solving environment" #443](https://github.com/SeldonIO/alibi-detect/issues/443).

agreed!

mauicv:
> > Do you mean having a tensorflow bucket and a TensorFlow-probability bucket that also has tensorflow or just having a single tensorflow bucket which includes tensorflow-probability? I think I'm generally in favour of the former.
> 
> I was thinking just have a single bucket. I'm wary of how many buckets we're going to have since especially if we introduce new backends like keops...

Ok, I've removed tensorflow-probability completely and incorporated into the tensorflow bucket.

alibi_detect/datasets.py
```diff
@@ -9,7 +9,7 @@
 import pandas as pd
 import requests
 from alibi_detect.utils.data import Bunch
-from alibi_detect.utils.fetching import _join_url
+from alibi_detect.utils.fetching.url import _join_url
```
ascillitoe, line 12: Please can you remind me why `_join_url` has been moved?

alibi_detect/models/tensorflow/__init__.py
```diff
@@ -11,7 +47,14 @@
     "VAE",
     "VAEGMM",
     "resnet",
+    "scale_by_instance",
```
ascillitoe, line 50: Just to double check; is `scale_by_instance` promoted from `tensorflow.resnet` to `tensorflow` so that we can do the `import_optional` check in `__init__`? This makes sense I think, although a minor point is that we would ordinarily want to raise a `DeprecationWarning` if the old import usage `from alibi_detect.models.tensorflow.resnet import scale_by_instance` is used. Perhaps not important in this case as `scale_by_instance` is not used a lot...

alibi_detect/tests/test_dep_management.py
```diff
@@ -57,61 +57,117 @@ def test_cd_dependencies(opt_dep):
     for dependency, relations in []:
         dependency_map[dependency] = relations
     if opt_dep != 'all':
-        with pytest.raises(ImportError):
-            from alibi_detect import cd
-            check_correct_dependencies(cd, dependency_map, opt_dep)
+        from alibi_detect import cd
```
ascillitoe, line 60: Are there any functional differences between the `alibi_detect` and `alibi` tests, or does the intention of the tests carry over?

alibi_detect/utils/fetching/fetching.py
```diff
@@ -1,40 +1,57 @@
...
-    LLR,
-    Mahalanobis,
-    OutlierAEGMM,
-    OutlierAE,
-    OutlierProphet,
-    OutlierSeq2Seq,
-    OutlierVAE,
-    OutlierVAEGMM,
-    SpectralResidual
+    'BaseDetector',
```
ascillitoe, line 42: What happens if forward refs aren't used?

alibi_detect/utils/frameworks.py
```diff
@@ -15,3 +18,90 @@
...
+
+
+HAS_BACKEND = {
+    'tensorflow': has_tensorflow,
+    'sklearn': has_sklearn,
+    'pytorch': has_pytorch,
+}
+
+
+def iter_to_str(iterable):
```
ascillitoe, line 30: Might be worth making this a private function if only intended to be used here. Otherwise needs a docstring.

alibi_detect/utils/perturbation.py
```diff
@@ -1,11 +1,18 @@
+# ###### CHANGED in TF-TFP-TORCH OPTIONAL DEPS PR ####################
+# ###### these comments should be removed prior to PR merge. #########
+# ####################################################################
+#
+# 1. The function `mutate_categorical` has been moved from here to
+# `utils/tensorflow/perturbation.py` to avoid tensorflow optional
```
ascillitoe, line 6: Makes sense 👍🏻

alibi_detect/utils/pytorch/__init__.py
```diff
@@ -16,5 +40,6 @@
     "predict_batch_transformer",
     "get_device",
     "quantile",
-    "zero_diag"
+    "zero_diag",
+    "TorchDataset"
```
ascillitoe, line 44: General point here: Where we have promoted (public!) imports up to a higher submodule (or just moved things around entirely), IMO we at a minimum need to document the changes properly in the release notes. Better yet we should raise `DeprecationWarning`'s if the old imports are used. 

An example is the old `from alibi_detect.utils.saving import save_detector` has been moved to `from alibi_detect.saving import save_detector`, however calling the former will still import `save_detector` and raise a `DeprecationWarning`.

alibi_detect/utils/saving/__init__.py
```diff
@@ -0,0 +1,17 @@
+# ###### CHANGED in TF-TFP-TORCH OPTIONAL DEPS PR ####################
+# ###### these comments should be removed prior to PR merge. #########
+# ####################################################################
+#
+# 2. Moved `saving` from `/utils/saving.py` to here: `/utils/saving/`.
```
ascillitoe, line 5: 👍🏻

alibi_detect/datasets.py
ascillitoe: Nevermind, becomes clear later. Please see later comment on this.

alibi_detect/utils/tensorflow/__init__.py
```diff
@@ -17,5 +45,6 @@
     "predict_batch_transformer",
     "quantile",
     "subset_matrix",
-    "zero_diag"
+    "zero_diag",
+    "mutate_categorical"
```
ascillitoe, line 49: See previous comments wrt to promoting import level.

doc/source/cd/methods/chisquaredrift.ipynb
```diff
@@ -112,7 +112,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "Python 3",
```
ascillitoe, line 115: Usual comment wrt to notebooks; ideally would reset these changes to reduce file diffs.

ascillitoe (commented):
Hey @mauicv, done a first pass. Looking good I think, but will hold off from approving until a few queries are resolved.

alibi_detect/models/tensorflow/__init__.py
```diff
@@ -1,8 +1,44 @@
-from .autoencoder import AE, AEGMM, VAE, VAEGMM, Seq2Seq
-from .embedding import TransformerEmbedding
-from .pixelcnn import PixelCNN
-from .resnet import resnet
-from .trainer import trainer
+# ###### CHANGED in TF-TFP-TORCH OPTIONAL DEPS PR ####################
+# ###### these comments should be removed prior to PR merge. #########
+# ####################################################################
+#
+# 1. loss functions that require tensorflow-probability have been
```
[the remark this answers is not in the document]
mauicv: Ah, yeah good point. I think I must have left it in from before when I combined tensorflow and tensorflow-probability! I'll put it back in `losses.py`!

alibi_detect/models/tensorflow/__init__.py
mauicv: Yeah, that's exactly correct. I'll add the relevent `DepreciationWarnings`

alibi_detect/tests/test_dep_management.py
mauicv: I'm not sure what you mean exactly?

alibi_detect/utils/fetching/fetching.py
```diff
@@ -1,40 +1,57 @@
...
+import tensorflow as tf
+from tensorflow.python.keras import backend
+
 from alibi_detect.models.tensorflow import PixelCNN
-from alibi_detect.od import (IForest, LLR, Mahalanobis, OutlierAE, OutlierAEGMM, OutlierProphet,
-                             OutlierSeq2Seq, OutlierVAE, OutlierVAEGMM, SpectralResidual)
-from alibi_detect.utils.saving import load_detector  # type: ignore
+
+if TYPE_CHECKING:
+    # Import the true objects directly for typechecking. If we import at the module level the import_optional method
```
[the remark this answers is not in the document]
mauicv: Yeah, I mean from `alibi_detect.ad` instead of `alibi_detect.ad.adversarialae`. It's not clear though so I'll reword it!

alibi_detect/utils/fetching/fetching.py
mauicv: The logic is a bit complex. If we remove the forward refs:
1. If not `TYPE_CHECKING` these objects won't be imported so an error will be thrown as they're not defined.
2. If we remove the import conditional on `TYPE_CHECKING` then if the user doesn't have all the optional dependencies installed anything imported that depends on an optional dependency will fail and throw an error.
3. If we use optional import instead (i.e. import from the `__init__.py`) then If the user doesn't have all optional dependencies installed then import optional will replace a `Class` Object with an instance of the `MissingDependency` class. This will throw an error when passed to `Union`. 
4. As well as this the import optional return type has to be `Any` and so this basically makes any type checking meaningless.

alibi_detect/tests/test_dep_management.py
ascillitoe: Just trying to understand if the `alibi-detect` dep management tests do pretty much the same thing as the `alibi` tests (obviously with different buckets). Or if there are any notable differences.

setup.py
```diff
@@ -30,9 +30,9 @@ def readme():
         "fbprophet>=0.5, <0.7",
         "holidays==0.9.11",
         "pystan<3.0",
-        "numba>=0.50.0, !=0.54.0, <0.56.0",
-        # "tensorflow_probability>=0.8.0, <0.18.0",
-        # "tensorflow>=2.2.0, !=2.6.0, !=2.6.1, <2.10.0",  # https://github.com/SeldonIO/alibi-detect/issues/375 and 387
+        "tensorflow_probability>=0.8.0, <0.18.0",
```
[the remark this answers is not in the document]
mauicv: I think `dependabot` does track optional deps. It made a [PR](https://github.com/SeldonIO/alibi/pull/702) for `shap` in Alibi

alibi_detect/tests/test_dep_management.py
ascillitoe: Got you, thanks! 👍🏻

alibi_detect/utils/fetching/fetching.py
ascillitoe: Ah yes I remember now, thanks!

alibi_detect/utils/pytorch/__init__.py
mauicv: Me and @ascillitoe discussed options: 

The best way of providing depreciation warnings is to move the functionality in `alibi_detect.utils.pytorch.misc` to `alibi_detect.utils.pytorch._misc`. We then update the `__init__.py` file to import from `_misc.py`:

```py
get_device, quantile, zero_diag = import_optional(
    'alibi_detect.utils.pytorch._misc',
    names=['get_device', 'quantile', 'zero_diag']
)
```

In the `misc.py` file we replace the relevant functions with a depreciation warning similar to in the [config work](https://github.com/SeldonIO/alibi-detect/blob/release/0.10.0/alibi_detect/utils/saving.py). I think this makes sense but the result would be inconsistencies in what is private and what is public in a wider sense throughout detect. I could try and make this consistent but this would mean renaming lots of files with the underscore...

README.md
```diff
@@ -117,7 +117,7 @@ We will use the [VAE outlier detector](https://docs.seldon.io/projects/alibi-det
 
 ```python
 from alibi_detect.od import OutlierVAE
-from alibi_detect.utils import save_detector, load_detector
+from alibi_detect.utils.saving import save_detector, load_detector
```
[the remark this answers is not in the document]
mauicv: I've put it in my todos, but hopefully, this should be flagged in conflicts when we merge no?

alibi_detect/models/tensorflow/__init__.py
mauicv: see [this](https://github.com/SeldonIO/alibi-detect/pull/538#discussion_r914862049) related comment!

setup.py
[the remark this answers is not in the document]
ascillitoe: Oh interesting! That seems to be the first time its ever made a `shap` PR (or any optional dep for that matter?): https://github.com/SeldonIO/alibi/issues?q=label%3Adependencies+is%3Aclosed. Wondering if that is just a coincidence or something has changed...

alibi_detect/utils/pytorch/__init__.py
ascillitoe: IMO moving backend-specific files to private (to avoid imports without optional deps checks) might be a worthwhile exercise in a subsequent PR (to include in `0.10.0` release). However, personally I would only vote for this change if it doesn't screw up the git history for these files too much i.e. if we can get the git history to just track the filename change instead of registering it as a complete deletion and rewrite of each file.

alibi_detect/models/tensorflow/losses.py
```diff
@@ -1,7 +1,8 @@
 import tensorflow as tf
-from tensorflow.keras.layers import Flatten
-from tensorflow.keras.losses import kld, categorical_crossentropy
 import tensorflow_probability as tfp
+from tensorflow.keras.losses import kld, categorical_crossentropy
```
ascillitoe, line 3: Uber pendantic nitpick, could just remove these changes and changes to `test_losses_tf.py`...

doc/source/cd/methods/spotthediffdrift.ipynb
```diff
@@ -123,7 +123,7 @@
     "```python\n",
     "import tensorflow as tf\n",
     "from tensorflow.keras.layers import Conv2D, Flatten, Input\n",
-    "from alibi_detect.utils.tensorflow.kernels import DeepKernel\n",
+    "from alibi_detect.utils.tensorflow import DeepKernel\n",
```
ascillitoe, line 126: Change also needs to be made in main text: `The DeepKernel class found in either alibi_detect.utils.tensorflow.kernels or alibi_detect.utils.pytorch.kernels`

Might be worth grep-ing all the import statements that have been changed?

alibi_detect/models/tensorflow/losses.py
mauicv: i'm not sure I understand?

doc/source/cd/methods/spotthediffdrift.ipynb
mauicv: Hmm, good idea. I found `alibi_detect.utils.pytorch.data.TorchDataset` and `alibi_detect.utils.tensorflow.data.TFDataset` as well, These are referenced in a parameter docstring. Are they intended to be public? Would a user import and use them? They're not used explicitly anywhere within the examples!

mauicv: In offline discussion agreed the `TorchDataset` and `TFDataset`'s should be public so I've protected them with `import_optional` and fixed private import paths in the docs.

PR #539, opened 2022-06-22
Feature Optional numba

mauicv:
# Todo:
- [x] Tox env tests
___

## What is this

This work branches off [`feature/dep-management`](https://github.com/SeldonIO/alibi-detect/pull/501) and will make `numba` optional dependency of `alibi-detect`.

___

### What this PR doesn't include:

1. The install options in the docs. I think it makes sense to do this in a later PR on the feature branch for all the optional deps all at once, once this branch has been merged in.
2. The install options in the `README.md` for same reason as 1.
3. The config-driven work in the [release branch](https://github.com/SeldonIO/alibi-detect/tree/release/0.10.0). This will be done in a separate PR in order to manage Release/Master branch complexity issues.
4. Make Licenses updates. This will be done in a PR on the feature branch.

___

### Notes:

- Should be merged after #538, 
    - [ ] check tests functionality is correct as well as `config.py`
- For some reason, this work caused some typing errors. I'm not sure why exactly as they didn't seem related. One of them I ignored with a comment (see #540) the others I fixed


ascillitoe:
@mauicv re the typing errors, the `pytorch/kernels.py` and `pytorch/context_aware.py` ones only seem to occur with `numpy>=1.23.0`. `numba 0.55.2` requires `numpy<1.23` hence these only became apparent when you removed `numba` from the CI install env. Still good you've fixed them imo. 

Re #540 I'm still looking into that one...

Issue #540, opened 2022-06-23
Intermittent typing error in pytorch mmd.py


mauicv:
I get the following error intermittently both locally and in CI:

```
alibi_detect/cd/pytorch/mmd.py:120: error: Item "List[Any]" of "Union[ndarray[Any, Any], List[Any]]" has no attribute "shape"  [union-attr]
Found 1 error in 1 file (checked 182 source files)
```

Initially, I first saw it only locally when first setting up alibi-detect but then later it disappeared (Can't recall exactly why?). Subsequently while working on #539 it turned up in CI. I couldn't recreate it locally until I deleted and rebuilt the conda environment at which point it cropped up again?

The underlying error seems to be a result of us ignoring assignments earlier in the code block. For the purposes of completing #539  I've just ignored type-checking for that line.

ascillitoe:
@mauicv I can't figure out why this error only appears intermittently (I actually can't reproduce locally), but it is part of a larger issue we have with `x_ref/x` being typed as `Union[np.ndarray, list]`, but then `np.ndarray` being expected after preprocessing (we already have a lot of type ignore comments elsewhere to deal with this i.e. see `score` in `LSDDDrift`). 

IMO we either need to enable `--allow-redefinition` with `mypy`, or rework the preprocessing variable names, for example change:

```python
x_ref = torch.from_numpy(x_ref).to(self.device)
```

to

```python
x_ref_torch = torch.from_numpy(x_ref).to(self.device)
```

(or change `x_ref` to `x_ref_orig` before preprocessing etc...)

PR #548, opened 2022-07-06
Add KeOps MMD detector

arnaudvl:
Add MMD detector using the [KeOps](https://github.com/getkeops/keops) (PyTorch) backend to further accelerate drift detection and scale up to larger datasets. This PR needs to be made compatible with the optional dependency management (incl. #538 and related).

This PR includes:
- [x] MMD detector implementation using KeOps
- [x] GaussianRBF kernel using KeOps
- [ ] Tests
- [ ] Docs
- [ ] Basic benchmarking example vs. PyTorch MMD

Once this PR is merged, it will be followed up by a similar implementation for the Learned (Deep) Kernel detector.


ascillitoe:
@arnaudvl I shall resolve these conflicts and then review once we have #537 merged.

I'm also adding @mauicv for review specifically to check my additions wrt to optional dependency handling.

ascillitoe:
Additional note: Need to check suitable error is raised when passed to `save_detector`. Implementing save/load functionality can be left to a future PR.

ascillitoe:
@arnaudvl I have merged in 0.10.0 associated changes from `master` into this PR. This primarily involves: 

- Updating the preprocessing kwarg's (`preprocess_x_ref` etc) in code, docs and tests.
- Incorporating @mauicv's new optional deps functionality. `keops` support is now installed with `pip install alibi-detect[keops]`. @mauicv could you check this please, and also add some tox tests for the `keops` backend.
- Raising a `NotImplemented` error when the keops based `MMDDrift` is passed to `save_detector`.

I seem to have introduced a `mypy` error which I shall investigate now...

ascillitoe:
@arnaudvl re the `mypy` error, I have simply added a type ignore:

https://github.com/SeldonIO/alibi-detect/blob/0db2239e53996e1c56b01758dfb1dc2f7eafa2a0/alibi_detect/cd/keops/mmd.py#L164-L171

This error has started arising in the same place in the pytorch version too (https://github.com/SeldonIO/alibi-detect/issues/540), and we are ignoring it in the same way. The issue is that `mypy` loses track of the types for `x_ref` and `x` due to the preceding `# type: ignore[assignment]`'s. IMO we should ignore for now and think about how we handle the type annotations for `x_ref` and `x` more generally in all our `score` and `predict` methods (as part of addressing #540).

ascillitoe:
`test_mmd_keops` passes on ubuntu builds but fails on macOS. We need to explore further or just say we don't support `keops` with macOS for now, and skip the tests...

doc/source/examples/cd_mmd_keops.ipynb
```diff
@@ -0,0 +1,513 @@
...
+    "\n",
+    "As illustrated in the experiments, KeOps allows you to drastically speed up and scale up drift detection to larger datasets without running into memory issues. The speed benefit of KeOps over the PyTorch (or TensorFlow) MMD detector decreases as the number of features increases. Note though that it is not advised to apply the (untrained) MMD detector to very high-dimensional data in the first place."
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python [conda env:detect]",
+   "language": "python",
+   "name": "conda-env-detect-py"
```
ascillitoe, line 496: This causes an error in the notebook CI.

mauicv:
@ascillitoe
> Incorporating @mauicv's new optional deps functionality. keops support is now installed with pip install alibi-detect[keops]. @mauicv could you check this please, and also add some tox tests for the keops backend (pretty please slightly_smiling_face).

As talked about offline, because `keops` is a backend it's not included in the `test_dep_managment` tests however it makes sense to have a tox environment for testing regardless. I've also added `torch` to the `keops` dependency bucket as they're both required in the MMD detector with the keops backend. This has required making some changes to the other torch `test_dep_management` tests. 

Wil do a review now as well

mauicv:
In the cd_mmd_drift notebook we have:

> The notebook requires [PyTorch](https://pytorch.org/) and KeOps to be installed. Once PyTorch is installed, KeOps can be installed via pip


Could change that to `pip install alibi-detect[keops]` now.

mauicv (approved):
LGTM

ascillitoe:
@arnaudvl I will wait for us to decide on what to do about Windows and macOS support before reviewing, so that I can review once tests are passing :)

mauicv:
Just noticed we haven't added `keops` to the `all` optional dependency bucket. Is this intentional?

doc/source/examples/cd_mmd_keops.ipynb
jklaise: `display_name` should also be reset to the default (either `Python 3` or `Python 3 (ipykernel)`) to not cause an error for a user opening the notebook.

alibi_detect/cd/mmd.py
```diff
@@ -82,27 +89,35 @@ def __init__(
...
         args = [kwargs['x_ref']]
         pop_kwargs = ['self', 'x_ref', 'backend', '__class__']
+        if backend == 'tensorflow' and has_tensorflow:
+            pop_kwargs += ['device', 'batch_size_permutations']
+            detector = MMDDriftTF
+        elif backend == 'pytorch' and has_pytorch:
+            pop_kwargs += ['batch_size_permutations']
+            detector = MMDDriftTorch
+        else:
+            pop_kwargs += ['configure_kernel_from_x_ref']
```
jklaise, line 107: The docstring for this parameter probably needs updating now that it's not used for the KeOps backend (including also a brief explanation why it's not relevant for KeOps).

alibi_detect/cd/mmd.py
```diff
@@ -66,6 +70,9 @@ def __init__(
             Whether to already configure the kernel bandwidth from the reference data.
         n_permutations
             Number of permutations used in the permutation test.
+        batch_size_permutations
+            KeOps computes the n_permutations of the MMD^2 statistics in chunks of batch_size_permutations.
+            Only relevant for 'keops' backend.
```
jklaise, line 75: One thing to probably think about now rather than later is naming conventions for parameters which are only relevant for certain backends. I suppose relying on docstring explanation is not the worst thing, but if you have a number of different options for each backend it may be cleaner to organize them in a dict, e.g. `keops_options = {'batch_size_permutations': 1000000}`.

On the other hand, it's possible that some parameter may be used for more than 1 backend but not all, in which case the dictionary convention may not be the simplest, nor pre-fixing parameters with the appropriate backend (e.g. `keops_batch_size_permutations` would work here, but not if the parameter applied to >=2 different backends).

alibi_detect/cd/pytorch/mmd.py
```diff
@@ -89,7 +89,7 @@ def __init__(
         # initialize kernel
         sigma = torch.from_numpy(sigma).to(self.device) if isinstance(sigma,  # type: ignore[assignment]
                                                                       np.ndarray) else None
-        self.kernel = kernel(sigma) if kernel == GaussianRBF else kernel
+        self.kernel = kernel(sigma).to(self.device) if kernel == GaussianRBF else kernel
```
jklaise, line 92: Seems like a bug, probably worth opening a separate PR to get this in `master` quickly and not clutter this PR?

alibi_detect/utils/keops/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .kernels import GaussianRBF
```
jklaise, line 1: Should this be protected with `import_optional`? @mauicv

alibi_detect/cd/mmd.py
jklaise: Seems like the explanation on the method page is a lot more comprehensive, we should keep them in sync as much as possible.

alibi_detect/utils/keops/__init__.py
mauicv: ah! yeah, perhaps it should! Is it part of the public api?

ascillitoe:
I have skipped the Windows tests since keops does not currently support Windows (https://github.com/getkeops/keops/issues/43). We still need to decide on MacOS...

alibi_detect/cd/mmd.py
```diff
@@ -82,27 +89,35 @@ def __init__(
...
         kwargs = locals()
         args = [kwargs['x_ref']]
         pop_kwargs = ['self', 'x_ref', 'backend', '__class__']
+        if backend == 'tensorflow' and has_tensorflow:
+            pop_kwargs += ['device', 'batch_size_permutations']
+            detector = MMDDriftTF
+        elif backend == 'pytorch' and has_pytorch:
+            pop_kwargs += ['batch_size_permutations']
+            detector = MMDDriftTorch
+        else:
```
ascillitoe, line 106: The `else` here is not consistent with the previous `if backend == 'tensorflow' and has_tensorflow:` and `elif backend == 'pytorch' and has_pytorch:`, since it does not check `has_keops` like the other two. That being said, I don't believe we need any `has_<backend>` checks here (See https://github.com/SeldonIO/alibi-detect/issues/576).

alibi_detect/utils/missing_optional_dependency.py
```diff
@@ -25,7 +25,8 @@
     "tensorflow_probability": 'tensorflow',
     "tensorflow": 'tensorflow',
     "torch": 'torch',
-    "pytorch": 'torch'
+    "pytorch": 'torch',
+    "keops": 'keops',
```
ascillitoe, line 29: One for @mauicv here. I've had to swap `pykeops` for `keops` here (in https://github.com/SeldonIO/alibi-detect/pull/548/commits/9c2da77734c1d30cbb8da8f9312da23a5441c249), as I was unable to import `MMDDrift` when `pykeops` was not installed, due to `BackendValidator` raising the error `KeyError: 'keops'`. 

However, this means the error message is slightly incorrect:
```
ImportError: `keops` not installed. Cannot initialize and run MMDDrift with keops backend. The necessary missing dependencies can be installed using `pip install alibi-detect[keops]`.
```
Since the library is actually called `pykeops` despite our bucket being called `keops`. Is this a limitation of `BackendValidator`?

On a side note, this is an example of why I still wonder if we should have optional deps tests to check the wrapper classes such as `MMDDrift` (`test_mmd.py` was failing when `pykeops` was not installed, however this was not picked up since we don't test this with the optional deps `tox` based tests).

mauicv: As spoken about offline the fix is to add `"pykeops": 'keops'` to the `ERROR_TYPES` dictionary.

ascillitoe: Just pushed a fix. Thanks Alex!

ascillitoe:
The `test_changed_notebooks` tests are failing because all notebooks are skipped. This is a known issue: https://github.com/SeldonIO/alibi/issues/647

alibi_detect/utils/missing_optional_dependency.py
```diff
@@ -25,7 +25,9 @@
...
+    "pykeops": 'keops',
```
jklaise, line 30: I don't actually understand why we need both `"keops": 'keops',` and `"pykeops": 'keops',` (or `"torch": 'torch',` and `    "pytorch": 'torch',` for that matter). Is this necessary? Perhaps a docstring or a comment for `ERROR_TYPES` is needed to clarify for the future (separate PR). @mauicv

alibi_detect/tests/test_dep_management.py
```diff
@@ -66,8 +66,8 @@ def test_cd_torch_dependencies(opt_dep):
 
     dependency_map = defaultdict(lambda: ['default'])
     for dependency, relations in [
-        ("HiddenOutput", ['torch']),
-        ("preprocess_drift", ['torch'])
+        ("HiddenOutput", ['torch', 'keops']),
```
jklaise, line 69: I don't really understand why `keops` is added for all these tests that are supposed to test `torch` dependencies?

alibi_detect/cd/tests/test_mmd.py
```diff
@@ -3,10 +3,13 @@
 from alibi_detect.cd import MMDDrift
 from alibi_detect.cd.pytorch.mmd import MMDDriftTorch
 from alibi_detect.cd.tensorflow.mmd import MMDDriftTF
+from alibi_detect.utils.frameworks import has_keops
+if has_keops:
+    from alibi_detect.cd.keops.mmd import MMDDriftKeops
```
jklaise, line 8: Not sure we need to protect this import as tests should have all dependencies, unless the concern is that Windows/Mac tests would not run because of missing `keops` support? @ascillitoe

alibi_detect/utils/keops/kernels.py
```diff
@@ -0,0 +1,91 @@
+import numpy as np
+from pykeops.torch import LazyTensor
+import torch
+import torch.nn as nn
+from typing import Callable, Optional
+
+
+def sigma_mean(x: LazyTensor, y: LazyTensor, dist: LazyTensor) -> torch.Tensor:
```
jklaise, line 8: Curious why `torch` and `tensorflow` backends have `sigma_median` heuristic, but this one has `sigma_mean`?

alibi_detect/utils/keops/kernels.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    Bandwidth estimation using the mean heuristic.
+
+    Parameters
+    ----------
+    x
+        LazyTensor of instances with dimension [Nx, 1, features].
+    y
+        LazyTensor of instances with dimension [1, Ny, features].
```
jklaise, line 17: Perhaps this is KeOps 101, but maybe worth adding here why there's an extra dimension for these tensors (compare to `torch` and `tensorflow` implementation without it. Perhaps link to https://www.kernel-operations.io/keops/_auto_tutorials/a_LazyTensors/plot_lazytensors_a.html ?

doc/source/examples/cd_mmd_keops.ipynb
```diff
@@ -0,0 +1,513 @@
+{
```
jklaise, line 1: <p>Worth being more explicit about the GPU requirement to run this notebook? Could add a warning/callout box so it's clear that it's long-running and requires/recommended GPU.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/548/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

jklaise, line 1: <p>Replace <code>&gt;&gt;</code> with <code>\gg</code> everywhere in the notebook for a better symbol of "much greater than".</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/SeldonIO/alibi-detect/pull/548/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

alibi_detect/utils/keops/kernels.py
```diff
@@ -0,0 +1,91 @@
...
+    The computed bandwidth, `sigma`.
+    """
+    n = x.shape[0]
+    if x.shape == y.shape and (dist.min(axis=1) == 0.).all() and \
+            (torch.arange(n) == dist.argmin(axis=1).cpu().view(-1)).all():
+        n_mean = n * (n - 1)
+    else:
+        n_mean = np.prod(dist.shape)
+    sigma = (.5 * dist.sum(1).sum().unsqueeze(-1) / n_mean) ** .5
+    return sigma
```
jklaise, line 32: Not sure I follow the logic, come comments would be appreciated!

1. What are the two different branch conditions?
2. Doesn't `x.shape == y.shape` imply that `Nx == Ny ==1` (from docstring shapes)? But then the following don't make sense to me as `dist` would have shape `(1, 1)`.

alibi_detect/utils/keops/kernels.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class GaussianRBF(nn.Module):
+    def __init__(
+        self,
+        sigma: Optional[torch.Tensor] = None,
+        init_sigma_fn: Callable = sigma_mean,
```
jklaise, line 39: Should we be consistent with `torch` and `tensorflow` and set this to be `None` by default? Not sure of the reasons for having done it for `torch` and `tensorflow`, perhaps for serialization? @ascillitoe

alibi_detect/utils/keops/kernels.py
```diff
@@ -0,0 +1,91 @@
...
+            self.log_sigma = nn.Parameter(sigma.log(), requires_grad=trainable)
+            self.init_required = False
+        self.init_sigma_fn = init_sigma_fn
+        self.trainable = trainable
+
+    @property
+    def sigma(self) -> torch.Tensor:
+        return self.log_sigma.exp()
+
+    def forward(self, x: LazyTensor, y: LazyTensor, infer_sigma: bool = False) -> LazyTensor:
```
jklaise, line 74: For curiosity, noting that whilst `keops` and `tensorflow` backends here take only native tensors, the `torch` backend also allows `numpy` arrays (which are converted to tensors immediately). To me it seems like the `torch` implementation may need rewriting to only deal with native tensors (e.g. `numpy` arrays would not be put on the GPU after conversion to tensors currently).

alibi_detect/utils/keops/kernels.py
```diff
@@ -0,0 +1,91 @@
...
+
+        dist = ((x - y) ** 2).sum(-1)
+
+        if infer_sigma or self.init_required:
+            if self.trainable and infer_sigma:
+                raise ValueError("Gradients cannot be computed w.r.t. an inferred sigma value")
+            sigma = self.init_sigma_fn(x, y, dist)  # .to(x.device)
```
jklaise, line 81: What's happening with the device here?

alibi_detect/utils/keops/kernels.py
```diff
@@ -0,0 +1,91 @@
...
+            with torch.no_grad():
+                self.log_sigma.copy_(sigma.log().clone())
+            self.init_required = False
+
+        gamma = 1. / (2. * self.sigma ** 2)
+        gamma = LazyTensor(gamma[None, None, :]) if len(dist.shape) == 2 else LazyTensor(gamma[None, None, None, :])
```
jklaise, line 87: What rank can `dist` be other than 2 (as specified in docstring)? This conditional needs some clarification.

alibi_detect/tests/test_dep_management.py
mauicv: The `keops` bucket includes `torch` as the `keops` functionality needs it. Therefore any functionality dependent on `torch` will also pass if the user has the `keops` bucket installed.

alibi_detect/utils/keops/kernels.py
```diff
@@ -0,0 +1,91 @@
...
+            self.init_required = False
+        self.init_sigma_fn = init_sigma_fn
+        self.trainable = trainable
+
+    @property
+    def sigma(self) -> torch.Tensor:
+        return self.log_sigma.exp()
+
+    def forward(self, x: LazyTensor, y: LazyTensor, infer_sigma: bool = False) -> LazyTensor:
+
```
jklaise, line 75: Noticed also that none of our kernel implementations have docstrings for `forward`, should we add some?

alibi_detect/utils/keops/tests/test_kernels_keops.py
```diff
@@ -0,0 +1,45 @@
...
+    return tests_gk[request.param]
+
+
+@pytest.mark.skipif(not has_keops, reason='Skipping since pykeops is not installed.')
+@pytest.mark.parametrize('gaussian_kernel_params', list(range(n_tests_gk)), indirect=True)
+def test_gaussian_kernel(gaussian_kernel_params):
+    sigma, n_features, n_instances, trainable = gaussian_kernel_params
+    xshape, yshape = (n_instances[0], n_features), (n_instances[1], n_features)
+
+    print(sigma, xshape, yshape, trainable)
```
jklaise, line 29: Unnecessary

alibi_detect/utils/keops/tests/test_kernels_keops.py
```diff
@@ -0,0 +1,45 @@
...
+
+    sigma = sigma if sigma is None else torch.from_numpy(sigma).float()
+    x = torch.from_numpy(np.random.random(xshape)).float()
+    y = torch.from_numpy(np.random.random(yshape)).float()
+
+    kernel = GaussianRBF(sigma=sigma, trainable=trainable)
+    infer_sigma = True if sigma is None else False
+    if trainable and infer_sigma:
+        with pytest.raises(Exception):
```
jklaise, line 38: `Exception` -> `ValueError` for tighter control. I note that the same could be done for the `tensorflow` and `torch` tests.

alibi_detect/tests/test_dep_management.py
jklaise: So are these tests now run twice, once with `[torch]` and once with `[keops]` installed?

alibi_detect/utils/missing_optional_dependency.py
mauicv: Yeah, good idea.  

Basically in the code we import from `pykeops` whereas when setting the backend the user specifies `keops` so somewhere we need to map between these two and it's here in the `ERROR_TYPES` dict. If the entry above is `keops` then the `impot_optional` will fail to catch the error when it imports from `pykeops`. If we change the above to `pykeops` then `import_optional` will catch the error but the Backend Validator will get a key error as `keops` is what the user specifies. So we could try and set a convention to only use one of `pykeops` or `keops` everywhere but I felt it would be more robust to just use a surjective mapping `ERROR_TYPES`.  

I'll make an issue and create another PR fixng this and some other stuff!

alibi_detect/utils/keops/tests/test_kernels_keops.py
```diff
@@ -0,0 +1,45 @@
...
+            kernel(LazyTensor(x[:, None, :]), LazyTensor(y[None, :, :]), infer_sigma=infer_sigma)
+    else:
+        k_xy = kernel(LazyTensor(x[:, None, :]), LazyTensor(y[None, :, :]), infer_sigma=infer_sigma)
+        k_xx = kernel(LazyTensor(x[:, None, :]), LazyTensor(x[None, :, :]), infer_sigma=infer_sigma)
+        assert k_xy.shape == n_instances and k_xx.shape == (xshape[0], xshape[0])
+        assert (torch.arange(xshape[0]) == k_xx.argmax(axis=1).cpu().view(-1)).all()
+        assert (k_xx.min(axis=1) >= 0.).all() and (k_xy.min(axis=1) >= 0.).all()
```
jklaise, line 45: Are these tests slightly different to the `torch` and `tensorflow` counterparts? Or is it just manipulating axis?

alibi_detect/tests/test_dep_management.py
mauicv: The `test_dep_management` tests are run once for every optional dependency bucket. All they do is check if the correct object is imported dependent on the optional dependencies present. So this one says that we should receive a `MissingOptionalDependecy` object for everything except `all`, `torch` and `keops`. As they're redundant in the full test suite we skip them there.

mauicv: It's just these tests though and they're reasonably fast. The only thing that takes time is building the tox environments.

alibi_detect/cd/keops/mmd.py
```diff
@@ -0,0 +1,179 @@
...
+        self.kernel = kernel(sigma).to(self.device) if kernel == GaussianRBF else kernel
+
+        # set the correct MMD^2 function based on the batch size for the permutations
+        self.batch_size = batch_size_permutations
+        self.n_batches = 1 + (n_permutations - 1) // batch_size_permutations
+
+        # infer the kernel bandwidth from the reference data
+        if self.infer_sigma or isinstance(sigma, torch.Tensor):
+            x = torch.from_numpy(self.x_ref).to(self.device)
+            _ = self.kernel(LazyTensor(x[:, None, :]), LazyTensor(x[None, :, :]), infer_sigma=self.infer_sigma)
```
jklaise, line 102: If `self.infer_sigma = False` and `sigma` is passed by user then isn't this call redundant?

alibi_detect/cd/keops/mmd.py
```diff
@@ -0,0 +1,179 @@
...
+            Batch of instances.
+
+        Returns
+        -------
+        p-value obtained from the permutation test, the MMD^2 between the reference and test set,
+        and the MMD^2 threshold above which drift is flagged.
+        """
+        x_ref, x = self.preprocess(x)
+        x_ref = torch.from_numpy(x_ref).float()  # type: ignore[assignment]
+        x = torch.from_numpy(x).float()  # type: ignore[assignment]
```
jklaise, line 166: I note that in the `torch` implementation this are also sent to `self.device`

alibi_detect/cd/keops/tests/test_mmd_keops.py
```diff
@@ -0,0 +1,100 @@
...
+    (preprocess_drift, {'model': HiddenOutput, 'layer': -1}),
+    (preprocess_list, None)
+]
+preprocess_at_init = [True, False]
+n_permutations = [10]
+batch_size_permutations = [10, 1000000]
+configure_kernel_from_x_ref = [True, False]
+tests_mmddrift = list(product(n_features, n_enc, preprocess, n_permutations, preprocess_at_init,
+                              batch_size_permutations, configure_kernel_from_x_ref))
+n_tests = len(tests_mmddrift)
```
jklaise, line 45: Noting that `update_x_ref = [{'last': 750}, {'reservoir_sampling': 750}, None]` is missing from the parameters compared to the `torch` and `tensorflow` tests.

alibi_detect/cd/keops/mmd.py
```diff
@@ -0,0 +1,179 @@
...
+            x = torch.from_numpy(self.x_ref).to(self.device)
+            _ = self.kernel(LazyTensor(x[:, None, :]), LazyTensor(x[None, :, :]), infer_sigma=self.infer_sigma)
+            self.infer_sigma = False
+        else:
+            self.infer_sigma = True
+
+    def _mmd2(self, x_all: torch.Tensor, perms: List[torch.Tensor], m: int, n: int) \
+            -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        Batched (across the permutations) MMD^2 computation for the original test statistic and the permutations.
```
jklaise, line 110: Is it worth writing a unit test for this?

jklaise (changes requested):
Please refer to my comments.

alibi_detect/cd/tests/test_mmd.py
ascillitoe: Yep, exactly that. The Windows (and potentially MacOS) issues mean that we can no longer assume **all** backends are always installed during CI.

alibi_detect/utils/keops/kernels.py
arnaudvl: As far as I know it is not possible to easily infer the median pairwise distance using KeOps (not a reduction operation on the symbolic tensor). As a result I opted for the mean instead of median pairwise distance. Realistically either could work well/terrible for a given use case.

alibi_detect/utils/keops/kernels.py
arnaudvl: Sure, can add that it's necessary to indicate how broadcasting takes place.

alibi_detect/utils/keops/tests/test_kernels_keops.py
arnaudvl: Slightly different due to KeOps restrictions.

alibi_detect/utils/keops/kernels.py
arnaudvl: 1. Checks if you're comparing x with itself (y == x) to take zero diagonal into account. You need to do it in a roundabout way given we're working on LazyTensors here.
2. Same as 1., if `x!=y` then you go to the else statement. Note that dist is also broadcasted and does not have dims of 1 anymore.

arnaudvl: Note that it isn't entirely foolproof: if for instance x takes consecutive elements which are the same there might be more 0's in the distance matrix where the argmin comes before the diagonal (the last check). Can avoid this I think by extracting the diagonal from `dist.argmin(axis=1)` and checking `torch.arange(n)` against that.

alibi_detect/utils/keops/kernels.py
arnaudvl: Should be removed when the `GaussianRBF` is already placed on device.

alibi_detect/utils/keops/kernels.py
arnaudvl: Good point, you can add the batch dim as well as explained in `keops.mmd.py`, will add clarification in the docstrings.

alibi_detect/utils/keops/kernels.py
arnaudvl: Could do, currently the forward pass docstrings are actually already described in the init, e.g. for PyTorch:
``` Gaussian RBF kernel: k(x,y) = exp(-(1/(2*sigma^2)||x-y||^2). A forward pass takes a batch of instances x [Nx, features] and y [Ny, features] and returns the kernel matrix [Nx, Ny].```

File: alibi_detect/cd/keops/mmd.py (hunk 24 above)
Comment (arnaudvl) [reply alibi_detect/cd/keops/mmd.py:166]: Yes here we don't want to do that as we're stacking all the permutations together and then optionally operate on slices of permutations which are placed on the device. Check the `_mmd2` function for the logic.

File: alibi_detect/cd/keops/mmd.py (hunk 23 above)
Comment (arnaudvl) [reply alibi_detect/cd/keops/mmd.py:102]: Yes good catch. Will change to:
```python
if isinstance(sigma, torch.Tensor):
   self.infer_sigma = False
elif self.infer_sigma:
   x = torch.from_numpy(self.x_ref).to(self.device)
    _ = self.kernel(LazyTensor(x[:, None, :]), LazyTensor(x[None, :, :]), infer_sigma=self.infer_sigma)
    self.infer_sigma = False
else:
   self.infer_sigma = True
```

alibi_detect/cd/keops/mmd.py
arnaudvl: Yes can do.

alibi_detect/utils/keops/__init__.py
mauicv: just for reference, this has been protected as of [#71142d7](https://github.com/SeldonIO/alibi-detect/pull/548/commits/71142d79b88c1b98085c86fbf5bf8e027fd6870f)

alibi_detect/utils/keops/kernels.py
jklaise: Would be great to add these comments in the code as it simplifies the comprehension a lot!

I'm still not sure about `x.shape == y.shape` because the docstring specifies shapes to be `[Nx, 1, features]` and `[1, Ny, features]`. Does it mean that in the special case that the same tensor is passed in both we actually support shapes`[Nx, 1, features]` and `[Nx, 1, features]`? If so, should this be documented (if it's intentional to make it easier for users to write `sigma_mean(x, x)` rather than reshuffling the axis on the second `x`)?

alibi_detect/cd/keops/mmd.py
jklaise: Should the same change be made for the `torch` and `tensorflow` ones (different PR)? I'm questioning this now as it's present in these two so perhaps there was some reason to do it this way, otherwise looks like a bug.

arnaudvl: I don't think so since there are 2 reasons to do this:
1. Infer a kernel bandwidth sigma.
2. Obtain kernel matrix `k_xx`.
These can be grouped for PyTorch/TensorFlow but not KeOps since there is no benefit from computing a symbolic `k_xx` in advance, so you only do it to infer sigma which as you noticed made the statement a bit redundant.

alibi_detect/cd/keops/mmd.py
```diff
@@ -0,0 +1,179 @@
...
+            x = torch.cat([x_all[perm[:m]][None, :, :] for perm in perms[i:j]], 0)
+            y = torch.cat([x_all[perm[m:]][None, :, :] for perm in perms[i:j]], 0)
+            if batch == 0:
+                x = torch.cat([x_all[None, :m, :], x], 0)
+                y = torch.cat([x_all[None, m:, :], y], 0)
+            x, y = x.to(self.device), y.to(self.device)
+
+            # batch-wise kernel matrix computation over the permutations
+            k_xx.append(self.kernel(
+                LazyTensor(x[:, :, None, :]), LazyTensor(x[:, None, :, :])).sum(1).sum(1).squeeze(-1))
```
mauicv, line 140: Is there a reason we don't use the `LazyTensor` within the kernel? That way we'd keep the `KeOps` dependency isolated to the kernel code.

alibi_detect/tests/test_dep_management.py
arnaudvl: @jklaise happy with this?

alibi_detect/cd/keops/mmd.py
arnaudvl: We could but would have to add an additional kwarg to the forward pass that all keops kernels (incl. user defined ones) need to comply with to specify what type of dimension expanding for broadcasting needs to be done: `[n_x, n_y, feature_dim]` or `[batch_dim, n_x, n_y, feature_dim]`. Keeping this out of the kernel avoids this piece of custom logic. I don't feel very strongly about this (had it inside the kernel initially actually) but tried to avoid putting logic which has little to do with kernel matrix computation in the kernel forward pass. Further thoughts @jklaise @ascillitoe ?

alibi_detect/utils/keops/kernels.py
arnaudvl: Opened a separate issue for this #583

alibi_detect/utils/keops/kernels.py
arnaudvl: @jklaise Happy keeping it in init?

doc/source/examples/cd_mmd_keops.ipynb
arnaudvl: From the introduction:

```text
In this notebook we will run a few simple benchmarks to illustrate the speed and memory improvements from using KeOps over vanilla PyTorch on the GPU (1x RTX 2080 Ti).
```
Not clear enough?

alibi_detect/cd/keops/mmd.py
mauicv: ok, the reason I ask is just that I'm thinking about the implementation w.r.t. the outlier detectors. It probably doesn't make sense to consider it within this PR. I'm writing up my notes on this [here](https://www.notion.so/Keops-b57624f16c01416b9a1589031202079d).

alibi_detect/utils/keops/kernels.py
arnaudvl: If not, I would prefer to open an issue to revise general docstring strategy since it's not isolated to this PR.

alibi_detect/cd/keops/tests/test_mmd_keops.py
arnaudvl: Done

alibi_detect/cd/keops/mmd.py
ascillitoe: A bit late to this one, but if it were the case that `MMDDriftKeops` and `MMDDriftTorch`detectors were almost identical, then I think it might be worth looking into isolating the KeOps functionality to the kernel (so that the KeOps backend would essentially be the torch backend with a special keops kernel). However, since`MMDDriftKeops` has many other differences (batching etc), I'm not quite sure what benefit isolating keops to the kernel code would bring? It doesn't save us much wrt to optional deps management does it? Since we're already inside `keops/` anyway... 

Deffo worth continuing the discussion wrt to outliers over on your notes!

arnaudvl: It's a bit different as @mauicv for outlier detectors since they can be treated as more vanilla models and don't have to deal with batching across permutations etc.

ascillitoe: Yep makes sense 👍🏻

ascillitoe (approved):
Generally looks very nice! 

Just a few minor comments on my end, plus I have not investigated the final task of testing `sigma_mean` in batch and non-batch settings.

README.md
```diff
@@ -180,8 +185,8 @@ The following tables show the advised use cases for each algorithm. The column *
 
 #### TensorFlow and PyTorch support
```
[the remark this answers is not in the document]
arnaudvl: Yes could adjust this once we've got a bit more general KeOps support (other drift + new outlier detectors).

doc/source/cd/methods/mmddrift.ipynb
```diff
@@ -73,23 +73,22 @@
...
     "```\n",
     "\n",
-    "The same detector in PyTorch:\n",
-    "\n",
-    "```python\n",
-    "cd = MMDDrift(x_ref, backend='pytorch', p_val=.05)\n",
-    "```\n",
-    "\n",
-    "We can also easily add preprocessing functions for both frameworks. The following example uses a randomly initialized image encoder in PyTorch:\n",
+    "We can also easily add preprocessing functions for the *TensorFlow * and *PyTorch* frameworks. The following example uses a randomly initialized image encoder in PyTorch:\n",
```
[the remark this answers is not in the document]
arnaudvl: Technically you could use any mixture of backends between the preprocessing and following drift detection computations, incl. backends which are not even explicitly supported by us. Of course I personally wouldn't encourage mixing these in practice since it can become painful for deployment, but we're not restricting anyone. I will update this sentence though.

alibi_detect/cd/keops/tests/test_mmd_keops.py
```diff
@@ -0,0 +1,121 @@
...
+        x_h1 = [_[None, :] for _ in x_h1]
+    preds = cd.predict(x_h1, return_p_val=True)
+    if preds['data']['is_drift'] == 1:
+        assert preds['data']['p_val'] < preds['data']['threshold'] == cd.p_val
+        assert preds['data']['distance'] > preds['data']['distance_threshold']
+    else:
+        assert preds['data']['p_val'] >= preds['data']['threshold'] == cd.p_val
+        assert preds['data']['distance'] <= preds['data']['distance_threshold']
+
+    # ensure the keops MMD^2 estimate matches the pytorch implementation for the same kernel
```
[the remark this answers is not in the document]
arnaudvl: Yes agreed. I had a chat expressing this exact sentiment with @jklaise, and discussed how we could modularise the types of tests in a few ways:
1. Basic tests, making sure shapes, types etc. are all fine.
2. Functionality tests, e.g. for `_mmd2` here.
3. Cross-backend tests: making sure similar functions between backends give the same result where desired.
These could share artefacts (e.g. models, input data) but be split out in very small test functions. It might be beneficial to first flesh out how we would approach this (if we want to do this) across the whole library.

alibi_detect/utils/frameworks.py
```diff
@@ -13,12 +13,18 @@
 except ImportError:
     has_pytorch = False
 
+try:
+    import pykeops  # noqa
```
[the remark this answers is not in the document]
mauicv: Yes, good catch! 

Agree w.r.t. the idea of automating these checks. I'll open an issue and create a PR!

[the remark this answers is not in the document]
ascillitoe: Nice one @mauicv!

@arnaudvl we should still make the change in this PR too I think.

jklaise (approved):
LGTM!

PR #553, opened 2022-07-08
Merge config into opt deps

mauicv:
# What is this

This PR updates the config functionality to work with optional dependencies. Currently, this is just protecting the tensorflow saving backend. Note this only really affects loading as it's impossible to instantiate and save a detector that depends on missing optional dependencies.


alibi_detect/saving/schemas.py
```diff
@@ -24,24 +24,23 @@
...
-#    import sklearn
-#    SupportedModels_list.append()  # TODO
-# SupportedModels is a tuple of possible models (conditional on installed deps). This is used in isinstance() etc.
-SupportedModels = tuple(SupportedModels_list)
-# SupportedModels_types is a typing Union for use with pydantic. We include all optional deps in here so that they
-# are all documented in the api docs (where the optional deps are not installed at build time)
-SupportedModels_types = Union['tf.keras.Model', UAE_tf, HiddenOutput_tf]
+    SupportedModels_tf = (tf.keras.Model, UAE_tf, HiddenOutput_tf)  # type: ignore
+if has_pytorch:
+    # import torch
```
mauicv, line 36: this raised a fake error which is why I've commented it out

alibi_detect/saving/typing.py
```diff
@@ -0,0 +1,41 @@
+from abc import ABC
```
jklaise, line 1: Would be good to add a docstring about the purpose of this module and the `ConfigDetector` class.

alibi_detect/saving/loading.py
```diff
@@ -68,7 +62,7 @@
 ]
 
 
-def load_detector(filepath: Union[str, os.PathLike], **kwargs) -> Detector:
+def load_detector(filepath: Union[str, os.PathLike], **kwargs) -> ConfigDetector:
```
jklaise, line 65: I'm assuming the name change from `Detector` to `ConfigDetector` is to signify that detectors that have config saving implemented are returned? But would this be misleading since legacy detectors are still returned from the `load_detector` function call? Should we just stick with `Detector` or is there something I'm missing?

alibi_detect/saving/loading.py
```diff
@@ -474,7 +470,14 @@ def resolve_config(cfg: dict, config_dir: Optional[Path]) -> dict:
                 if src in registry.get_all():
                     obj = registry.get(src)
                 else:
-                    raise ValueError("Can't find {} in the custom function registry".format(src))
+                    raise ValueError(
+                        "Can't find {} in the custom function registry, It may be misspelled or missing "
+                        "if you have incorrect optional dependencies installed. Make sure the loading environment"
+                        " is the same as the saving environment. For more information, check the Installation "
+                        "documentation at "
+                        "https://docs.seldon.io/projects/alibi-detect/en/latest/overview/getting_started.html."
```
jklaise, line 478: stick with `stable` links as most users will have a `stable` version installed instead of `latest`

alibi_detect/saving/loading.py
```diff
@@ -474,7 +470,14 @@ def resolve_config(cfg: dict, config_dir: Optional[Path]) -> dict:
...
+                        .format(src)
```
jklaise, line 479: Was there a reason not to use f-strings in error messages?

alibi_detect/saving/saving.py
```diff
@@ -49,7 +48,7 @@ def save_detector(detector: Detector, filepath: Union[str, os.PathLike], legacy:
 
     # TODO: Replace .__args__ w/ typing.get_args() once Python 3.7 dropped (and remove type ignore below)
     detector_name = detector.__class__.__name__
-    if detector_name not in [detector.__name__ for detector in Detector.__args__]:  # type: ignore[attr-defined]
+    if detector_name not in [detector for detector in ALLOWED_DETECTORS]:  # type: ignore[attr-defined]
```
jklaise, line 51: the `type: ignore` can probably be removed now

alibi_detect/saving/schemas.py
```diff
@@ -53,6 +52,20 @@ def coerce_int2list(value: int) -> List[int]:
...
+def validate_model(model: Callable, values: dict) -> Callable:
+    """Validator to check the model is compatible with the given backend"""
+    backend = values['backend']
+    if backend == 'tensorflow':
+        assert isinstance(model, SupportedModels_tf)
+    elif backend == 'pytorch':
+        assert isinstance(model, SupportedModels_torch)
+    elif backend == 'sklearn':
+        assert isinstance(model, SupportedModels_sklearn)
+    return model
```
jklaise, line 64: I'm not too familiar with `pydantic` validators, but shouldn't the `assert` statements be replaced with raising errors?

alibi_detect/saving/schemas.py
```diff
@@ -53,6 +52,20 @@ def coerce_int2list(value: int) -> List[int]:
...
+
+
+# TODO - we could add another validator to check given "backend" against what optional deps are installed?
```
jklaise, line 67: Not sure what you mean, if the specified backend is missing some dependencies then the user wouldn't be able to instatiate the detector which happens way before saving?

alibi_detect/saving/tensorflow/_loading.py
```diff
@@ -219,7 +181,7 @@ def load_embedding(src: str, embedding_type, layers) -> TransformerEmbedding:
 #######################################################################################################
 # TODO: Everything below here is legacy loading code, and will be removed in the future
 #######################################################################################################
-def load_detector_legacy(filepath: Union[str, os.PathLike], suffix: str, **kwargs) -> Detector:
+def load_detector_legacy(filepath: Union[str, os.PathLike], suffix: str, **kwargs) -> BaseDetector:
```
jklaise, line 184: Are we distinguishing between legacy detector formats and config detector formats via `BaseDetector` and `ConfigDector`? If so, isn't `BaseDetector` a bit misleading (as it's a base class; further, don't config-based detectors inherit from `BaseDetector` anyway?)?

alibi_detect/saving/tensorflow/_loading.py
```diff
@@ -266,14 +228,14 @@ def load_detector_legacy(filepath: Union[str, os.PathLike], suffix: str, **kwarg
         raise NotImplementedError('Detectors with PyTorch backend are not yet supported.')
 
     detector_name = meta_dict['name']
-    if detector_name not in [detector.__name__ for detector in Detector.__args__]:  # type: ignore[attr-defined]
+    if detector_name not in [detector for detector in ALLOWED_DETECTORS]:  # type: ignore[attr-defined]
```
jklaise, line 231: can probably remove `type: ignore` now

alibi_detect/saving/tests/test_validate.py
```diff
@@ -3,9 +3,12 @@
 from pydantic import ValidationError
 
 from alibi_detect.saving import validate_config
-from alibi_detect.saving.saving import X_REF_FILENAME
+# from alibi_detect.saving.saving import X_REF_FILENAME
```
jklaise, line 6: Remove line?

alibi_detect/utils/missing_optional_dependency.py
```diff
@@ -96,6 +96,7 @@ def import_optional(module_name: str, names: Optional[List[str]] = None) -> Any:
             return objs if len(objs) > 1 else objs[0]
         return module
     except (ImportError, ModuleNotFoundError) as err:
+        # print(err, err.name)
```
jklaise, line 99: Remove?

setup.cfg
```diff
@@ -39,6 +39,7 @@ ignore_errors = True
 # tox test environments for generating licenses and testing optional dependencies
 [tox:tox]
 envlist=
+    validate
```
jklaise, line 42: What is this for?

jklaise (changes requested):
Left some comments, mostly was struggling to follow the logic without the background information. Hopefully adding some info to the comments and comments in the source files would help here.

alibi_detect/saving/loading.py
ascillitoe: Yeh admittedly `ConfigDetector` is a bit of a hack (hope you don't mind me saying @mauicv!). This was necessary because I implemented the `get_config` and `from_config` methods via `DriftConfigMixin` instead of as abstract methods of `BaseDetector`. The reason for this was because outlier and adversarial detectors do not yet support config save/load. 

IMO the most principled solution would either be to create a separate `BaseDriftDetector`, and define `get_config` and `from_config` in there, or alternatively, define the methods in `BaseDetector` and have the methods raise `NotImplementedError` in outlier and adversarial detectors. The `DriftConfigMixin` was intended to be a temporary measure whilst we decided what exactly the revamped outlier detectors would look like...

alibi_detect/saving/schemas.py
ascillitoe: Answering as I implemented this: 

I copied in `assert`'s from the [`pydantic` docs](https://pydantic-docs.helpmanual.io/usage/validators/) without thinking about it enough. You are right though, in their docs they point out "If you make use of assert statements, keep in mind that running Python with the [-O optimization flag](https://docs.python.org/3/using/cmdline.html#cmdoption-o) disables assert statements, and validators will stop working.". To avoid commit clashes would you be able to change these to `raise ValueError(...)` please @mauicv ?

alibi_detect/saving/loading.py
mauicv: Yeah, I agree w.r.t. `DriftConfigMixin` feeling hacky. It's frustrating that [intersection types](https://github.com/python/typing/issues/213) aren't implemented but it seems there quite tricky.

If we decide to stick with it however we can also use [`TypeGuards`](https://mypy.readthedocs.io/en/latest/type_narrowing.html#user-defined-type-guards). So we define a function:

```py
def _is_config_detector(detector: Union[BaseDetector, ConfigDetector]) -> TypeGuard[ConfigDetector]:
    return hasattr(detector, 'get_config')
```

That narrows the detector type. This is simpler than changing the inheritance structure of all the detectors but perhaps still less principled.

Perhaps the method of making the above more "Proper" is to use [protocols](https://docs.python.org/3/library/typing.html#typing.Protocol). See also [this](https://stackoverflow.com/a/70816342) and [this](https://stackoverflow.com/a/50320982). I haven't tried this yet though.

jklaise: I like the protocol approach and I think it is also the modern/recommended way of doing this sort of thing. As @ascillitoe mentioned, would need to check if `Protocol` and `runtime_checkable` are available in `typing_extensions` for Python 3.7 support.

alibi_detect/saving/loading.py
mauicv: I don't think so, i've replaced with `f-strings`

alibi_detect/saving/tensorflow/_loading.py
mauicv: So the `ConfigDetector` class (Now the `ConfigurableDetector` protocol) distinguishes which detectors can be saved using the config approach. But basically, I'm just using it to check the detectors have the correct methods required for config-driven saving. In the case of legacy saving, it's not a question of what methods the detector has it's whether or not the function to save the detector exists. I can add a protocol for those detectors if we think that's a good idea?

alibi_detect/saving/tests/test_validate.py
mauicv: It's a mistake, I've removed!

setup.cfg
mauicv: Ah, I was using a tox environment to validate some of the saving stuff but it proved unessesery. I've removed it now.

alibi_detect/saving/schemas.py
jklaise: Yea the main thing was that I didn't understand the comment :)

alibi_detect/saving/tensorflow/_loading.py
jklaise: I'm not sure we need it unless there is a clear protocol definition of what constitutes a legacy detector? I'm just confused why the return type is `BaseDetector`, espcially because e.g. `MMDDrift` doesn't inherit from `BaseDetector`.

mauicv: Ah good point, does it make sense to define two protocols in that case? So:

```py
@runtime_checkable
class Detector(Protocol):
    meta: Dict

    def predict(self): ...


@runtime_checkable
class ConfigurableDetector(Detector):
    def get_config(self): ...

    def from_config(self): ...

    def _set_config(self): ...
```

Otherwise, we'll have to use a Union of all of them as there's no common base class. I think this will reintroduce the tensorflow import issues that we're trying to circumvent.

jklaise: Sounds reasonable to me if @ascillitoe agrees

alibi_detect/saving/loading.py
```diff
@@ -68,7 +62,7 @@
 ]
 
 
-def load_detector(filepath: Union[str, os.PathLike], **kwargs) -> Detector:
+def load_detector(filepath: Union[str, os.PathLike], **kwargs) -> Union[BaseDetector, ConfigurableDetector]:
```
mauicv, line 65: `BaseDetector` -> `Detector`?

alibi_detect/saving/saving.py
```diff
@@ -49,7 +51,7 @@ def save_detector(detector: Detector, filepath: Union[str, os.PathLike], legacy:
 
     # TODO: Replace .__args__ w/ typing.get_args() once Python 3.7 dropped (and remove type ignore below)
     detector_name = detector.__class__.__name__
-    if detector_name not in [detector.__name__ for detector in Detector.__args__]:  # type: ignore[attr-defined]
+    if detector_name not in [detector for detector in ALLOWED_DETECTORS]:
```
mauicv, line 54: Rename to `VALID_DETECTORS`

alibi_detect/utils/saving/saving.py
```diff
@@ -25,7 +29,7 @@ def save_detector(detector: Detector, filepath: Union[str, os.PathLike], legacy:
     return _save_detector(detector, filepath, legacy)
 
 
-def load_detector(filepath: Union[str, os.PathLike], **kwargs) -> Detector:
+def load_detector(filepath: Union[str, os.PathLike], **kwargs) -> Union[BaseDetector, ConfigurableDetector]:
```
jklaise, line 32: Replace `BaseDetector` with the `Detector` protocol?

alibi_detect/saving/saving.py
```diff
@@ -27,7 +26,10 @@
 C_REF_FILENAME = 'c_ref.npy'
 
 
-def save_detector(detector: Detector, filepath: Union[str, os.PathLike], legacy: bool = False) -> None:
+def save_detector(
+        detector: Union[BaseDetector, ConfigurableDetector],
```
jklaise, line 30: Replace `BaseDetector` with the `Detector` protocol?

jklaise (approved):

alibi_detect/saving/schemas.py
ascillitoe: Yep good idea 👍🏻

alibi_detect/base.py
```diff
@@ -3,8 +3,10 @@
 import json
 import numpy as np
 from typing import Dict, Any, Optional
+from typing_extensions import Protocol, runtime_checkable
```
ascillitoe, line 6: For Python version dependent `typing` imports we usually add some logic to `utils._types` to import from `typing` or `typing_extensions` depending on the current Python version i.e. see how we handle `Literal`. Should we not do the same for `Protocol`?

jklaise: Agreed, especially since `typing_extensions` is only conditionally installed depending on Python version.

alibi_detect/saving/loading.py
ascillitoe: I am also confused about this one. Are we not meant to be using the protocol's (i.e. `Detector`) rather than the `BaseDetector` class in the return type now?

alibi_detect/saving/loading.py
```diff
@@ -404,10 +398,12 @@ def _set_dtypes(cfg: dict):
             lib, dtype, *_ = val.split('.')
             # val[0] = np if val[0] == 'np' else tf if val[0] == 'tf' else torch if val[0] == 'torch' else None
             # TODO - add above back in once optional deps are handled properly
-            lib = np if lib == 'np' else tf if lib == 'tf' else None
             if lib is None:
                 raise ValueError("`dtype` must be in format np.<dtype>, tf.<dtype> or torch.<dtype>.")
-            _set_nested_value(cfg, key, getattr(lib, dtype))
+            {
```
ascillitoe, line 403: This is pretty clever 🙂

alibi_detect/base.py
jklaise: Interesting, didn't know the extensions did that check itself.

Now I'm concerned that `typing_extensions` is missing from the installation as I can't find a reference to it, but then our 3.7 tests shouldn't be passing... https://github.com/SeldonIO/alibi-detect/search?q=typing_extensions

For `alibi` it used to be conditional on the Python version, but that was changed in this PR, not sure if it was a conscious decision or we missed it: https://github.com/SeldonIO/alibi/pull/603/files

All-in-all I think `typing_extensions` is a lightweight dependency so I'm happy to just have it included as a core dependency and remove our conditional logic regarding types.

alibi_detect/saving/saving.py
ascillitoe: Agreed.

alibi_detect/base.py
jklaise: Re tests passing for 3.7 but us not having it as a dep - probably another one of our deps installs it. Regardless we should add it to `setup.py`.

alibi_detect/saving/tensorflow/_loading.py
```diff
@@ -281,9 +241,9 @@ def load_detector_legacy(filepath: Union[str, os.PathLike], suffix: str, **kwarg
         vae = load_tf_vae(filepath, state_dict)
         detector = init_od_vae(state_dict, vae)
     elif detector_name == 'Mahalanobis':
-        detector = init_od_mahalanobis(state_dict)
+        detector = init_od_mahalanobis(state_dict)  # type: ignore[assignment]
```
ascillitoe, line 244: Wondering if we could change the earlier type hint:

```python
 detector = None  # type: Optional[Detector]  # to avoid mypy errors
```

to avoid needing these type ignore's? i.e. change it to `Optional[ALLOWED_DETECTORS]`?

alibi_detect/saving/validate.py
```diff
@@ -5,6 +5,41 @@
 from alibi_detect.version import __config_spec__, __version__
 
 
+ALLOWED_DETECTORS = [
```
ascillitoe, line 8: Nitpick - is `validate` the right place for this? I'd argue its more to do with typing than validating configs. We should at least add a comment saying what its for IMO.

alibi_detect/base.py
ascillitoe: How strange!

mauicv: I think I removed the `<python3.8` condition in that PR because we need [`TypeAlias`](https://docs.python.org/3.8/library/typing.html#type-aliases) from typing-extensions for python3.8. Also seems that [huggingface-hub](https://github.com/huggingface/huggingface_hub/blob/b6accae2ebfe2d0986934f2933b7bf2fb3811726/setup.py#L19) is the dependency that installs `typing-extensions`

alibi_detect/saving/loading.py
mauicv: Yeah, I missed some :face_exhaling:

alibi_detect/saving/tensorflow/_loading.py
mauicv: `ALLOWED_DETECTORS` (now `VALID_DETECTORS`) is of type: `List[str]` now? Doing the above gives me `mypy` errors. Is that what you meant?

alibi_detect/tests/test_dep_management.py
```diff
@@ -188,18 +188,26 @@ def test_fetching_utils_dependencies(opt_dep):
...
 
 
-def test_saving_utils_dependencies(opt_dep):
+def test_saving_tf_dependencies(opt_dep):
     """Tests that the saving utils module correctly protects against uninstalled optional dependencies.
     """
 
     dependency_map = defaultdict(lambda: ['default'])
     for dependency, relations in [
-        ('load_detector', ['tensorflow']),
```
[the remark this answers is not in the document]
mauicv: Good point, i've added that test!

alibi_detect/saving/tensorflow/_loading.py
ascillitoe: Ooops ignore me, I had in my head that `VALID_DETECTORS` was a `Union` of ForwardRef's when I said this. I am happy to leave these type ignore's here, especially since it is a legacy function.

ascillitoe (commented):
LGTM!

PR #555, opened 2022-07-14
FEAT: pass through preprocessing_func to training

WilliamJudge94:
The option to create ones own preprocessing function for each epoch was hidden. This pull request gives that option back to the users for a Tensorflow model.

### DIFF
--- a/alibi_detect/od/vae.py
+++ b/alibi_detect/od/vae.py
@@ -108,6 +108,8 @@ class OutlierVAE(BaseDetector, FitMixin, ThresholdMixin):
             Whether to print training progress.
         log_metric
             Additional metrics whose progress will be displayed if verbose equals True.
+        preprocess_fn
+            Preprocessing function applied to each training batch.
         callbacks
             Callbacks used during training.
         """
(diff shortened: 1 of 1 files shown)


ascillitoe:
Hi @WilliamJudge94, welcome, and thank you for contribution! I'm afraid we will not be able to accept this particular PR since we are intentionally avoiding adding `preprocess_fn` support to our outlier detectors for now. We're doing this for a number of reasons:

- We include `preprocess_fn` support for drift detectors to allow them to work for more complex data modalities such as text and graph data. Outlier detectors are not really as applicable to these modalities.
- We will soon be revamping and our `trainer`'s to allow for pytorch `DataLoader`'s and the tensorflow equivalent, as well as revamping our outlier detectors, so adding more complexity at this point isn't desirable.

If you can tell us more about your use case we'll be happy to recommend workarounds given the above. If you would like to contribute to the library we can also recommend possible issues to tackle 🙂

WilliamJudge94:
@ascillitoe thank you for getting back to me! 

The reason I need the pre processing function is to add random Poisson noise to each input batch when training. My group has an incredibly small dataset to work with and have found this is the only way to train an efficient outlier detector. We also have limited memory allocations so storing addition noisy datasets in memory is not possible. 

Is there another way to add random noise to each input dataset for each training batch?

PR #557, opened 2022-07-16
Tabular classifier drift example - dummy trap

RobertSamoilescu:
Minor fix in `ClassifierDrift` sklearn example. Included `drop='first'` in OHE to address dummy trap, which is a standard procedure.

### DIFF
--- a/doc/source/examples/cd_clf_adult.ipynb
+++ b/doc/source/examples/cd_clf_adult.ipynb
@@ -206,7 +206,7 @@
     "# define numerical standard scaler.\n",
     "num_transf = StandardScaler()\n",
     "\n",
-    "# define categorical one-hot encoder.\n",
+    "# define categorical one-hot encoder with first columns dropped.\n",
     "cat_transf = OneHotEncoder(\n",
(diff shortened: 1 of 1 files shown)


RobertSamoilescu:
@ascillitoe, it is more like a standard procedure and not really a problem (I believe it might be for some linear estimators). Just wanted to be consistent with some the best practices. For more context on how it came about, check this [issue](https://github.com/SeldonIO/alibi/issues/718) from alibi.

ascillitoe (approved):
LGTM! Taking your word for it that the outlier examples don't need updating 🙂

doc/source/examples/cd_clf_adult.ipynb
```diff
@@ -209,7 +209,7 @@
     "# define categorical one-hot encoder.\n",
```
ascillitoe, line 209: Sorry @RobertSamoilescu, one last thought before we merge, is it worth adding a note as to why we do `drop="first"` here? i.e.

```python
# Define a categorical one-hot encoder with 1st variable dropped to avoid multicollinearity
```

PR #558, opened 2022-07-18
[WIP] Kernel function extension

Srceh:
This PR aims to increase the number of supported kernels and associated operations. 

The objectives of the PR can be summarized as follows:

- [ ] Allow different kernels to be built upon a base kernel class (i.e. periodic kernel).
- [ ] Allow each kernel to select a subset of the feature dimensions, enabling different feature dimensions to be treated differently (i.e. use periodic kernel on time-related features and RBF on other numeric features).
- [ ] Implement combined kernels by adding or product two or more different kernels.
- [ ] Provide generic methods to initialise and train different kernel parameters as a whole set. 

The work can therefore be divided into the following stages:

- [ ] Design and change the current kernel implementation, adding base classes for kernels and combined kernels.
- [ ] Change the parameter initialisation and training behaviour among existing kernel-based detectors.
- [ ] Adding docs and tutorial notebooks on the new kernels and usages.
- [ ] Make additional testing on the introduced functions and procedures.

### DIFF
--- a/alibi_detect/utils/pytorch/kernels.py
+++ b/alibi_detect/utils/pytorch/kernels.py
@@ -29,12 +29,66 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
     return sigma
 
 
-class GaussianRBF(nn.Module):
+class BaseKernel(nn.Module):
+    """
+    The base class for all kernels.
+    Args:
+        nn (_type_): _description_
+    """
+    def __init__(self) -> None:
+        super().__init__()
+        self.parameter_dict: dict = {}
+        self.active_dims: Optional[list] = None
+
+    def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+        raise NotImplementedError
+
+
+class SumKernel(nn.Module):
+    """
+    Construct a kernel by summing two kernels.
+    Args:
+        nn (_type_): _description_
+    """
     def __init__(
         self,
-        sigma: Optional[torch.Tensor] = None,
-        init_sigma_fn: Callable = sigma_median,
-        trainable: bool = False
+        kernel_a: BaseKernel,
+        kernel_b: BaseKernel
+    ) -> None:
+        super().__init__()
+        self.kernel_a = kernel_a
+        self.kernel_b = kernel_b
+
+    def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+        return self.kernel_a(x, y) + self.kernel_b(x, y)
+
+
+class ProductKernel(nn.Module):
+    """
+    Construct a kernel by multiplying two kernels.
+    Args:
+        nn (_type_): _description_
+    """
+    def __init__(
+        self,
+        kernel_a: BaseKernel,
+        kernel_b: BaseKernel
+    ) -> None:
+        super().__init__()
+        self.kernel_a = kernel_a
+        self.kernel_b = kernel_b
+
+    def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+        return self.kernel_a(x, y) * self.kernel_b(x, y)
+
+
+class GaussianRBF(BaseKernel):
+    def __init__(
+       self,
+       sigma: Optional[torch.Tensor] = None,
+       init_fn_sigma: Callable = sigma_median,
+       trainable: bool = False,
+       active_dims: Optional[list] = None
     ) -> None:
         """
         Gaussian RBF kernel: k(x,y) = exp(-(1/(2*sigma^2)||x-y||^2). A forward pass takes
@@ -54,28 +108,28 @@ class GaussianRBF(nn.Module):
             Whether or not to track gradients w.r.t. `sigma` to allow it to be trained.
         """
         super().__init__()
+        self.parameter_dict['sigma'] = 'bandwidth'
         if sigma is None:
             self.log_sigma = nn.Parameter(torch.empty(1), requires_grad=trainable)
             self.init_required = True
         else:
-            sigma = sigma.reshape(-1)  # [Ns,]
             self.log_sigma = nn.Parameter(sigma.log(), requires_grad=trainable)
             self.init_required = False
-        self.init_sigma_fn = init_sigma_fn
-        self.trainable = trainable
+        self.init_fn_sigma = init_fn_sigma
+        self.active_dims = active_dims
 
     @property
     def sigma(self) -> torch.Tensor:
         return self.log_sigma.exp()
 
     def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch.Tensor],
-                infer_sigma: bool = False) -> torch.Tensor:
+                infer_parameter: bool = False) -> torch.Tensor:
 
         x, y = torch.as_tensor(x), torch.as_tensor(y)
         dist = distance.squared_pairwise_distance(x.flatten(1), y.flatten(1))  # [Nx, Ny]
 
-        if infer_sigma or self.init_required:
-            if self.trainable and infer_sigma:
+        if infer_parameter or self.init_required:
+            if self.trainable and infer_parameter:
                 raise ValueError("Gradients cannot be computed w.r.t. an inferred sigma value")
             sigma = self.init_sigma_fn(x, y, dist)
             with torch.no_grad():
@@ -86,6 +140,173 @@ class GaussianRBF(nn.Module):
         # TODO: do matrix multiplication after all?
         kernel_mat = torch.exp(- torch.cat([(g * dist)[None, :, :] for g in gamma], dim=0))  # [Ns, Nx, Ny]
         return kernel_mat.mean(dim=0)  # [Nx, Ny]
+
+
+class RationalQuadratic(nn.Module):
+    def __init__(
+        self,
+        alpha: torch.Tensor = None,
+        init_fn_alpha: Callable = None,
+        sigma: torch.Tensor = None,
+        init_fn_sigma: Callable = None,
+        trainable: bool = False,
+        active_dims: Optional[list] = None
+    ) -> None:
+        """
+        Rational Quadratic kernel: k(x,y) = (1 + ||x-y||^2 / (2*sigma^2))^(-alpha).
+        A forward pass takesa batch of instances x [Nx, features] and y [Ny, features]
+        and returns the kernel matrix [Nx, Ny].
+
+        Parameters
+        ----------
+        alpha
+            Exponent parameter of the kernel.
+        sigma
+            Bandwidth used for the kernel.
+        """
+        super().__init__()
+        self.parameter_dict['alpha'] = 'exponent'
+        self.parameter_dict['sigma'] = 'bandwidth'
+        if alpha is None:
+            self.alpha = nn.Parameter(torch.empty(1), requires_grad=trainable)
+            self.init_required = True
+        else:
+            self.alpha = alpha
+            self.init_required = False
+        if sigma is None:
+            self.log_sigma = nn.Parameter(torch.empty(1), requires_grad=trainable)
+            self.init_required = True
+        else:
+            self.log_sigma = nn.Parameter(sigma.log(), requires_grad=trainable)
+            self.init_required = False
+        self.init_fn_alpha = init_fn_alpha
+        self.init_fn_sigma = init_fn_sigma
+        self.active_dims = active_dims
+
+    @property
+    def sigma(self) -> torch.Tensor:
+        return self.log_sigma.exp()
+
+    def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch.Tensor]) -> torch.Tensor:
+        x, y = torch.as_tensor(x), torch.as_tensor(y)
+        dist = distance.squared_pairwise_distance(x.flatten(1), y.flatten(1))
+        kernel_mat = (1 + torch.square(dist) / (2 * self.alpha * (self.sigma ** 2))) ** (-self.alpha)
+        return kernel_mat
+
+
+class Periodic(nn.Module):
+    def __init__(
+        self,
+        tau: torch.Tensor = None,
+        init_fn_tau: Callable = None,
+        sigma: torch.Tensor = None,
+        init_fn_sigma: Callable = None,
+        trainable: bool = False,
+        active_dims: Optional[list] = None
+    ) -> None:
+        """
+        Periodic kernel: k(x,y) = .
+        A forward pass takesa batch of instances x [Nx, features] and y [Ny, features]
+        and returns the kernel matrix [Nx, Ny].
+
+        Parameters
+        ----------
+        tau
+            Period of the periodic kernel.
+        sigma
+            Bandwidth used for the kernel.
+        """
+        super().__init__()
+        self.parameter_dict['tau'] = 'period'
+        self.parameter_dict['sigma'] = 'bandwidth'
+        if tau is None:
+            self.log_tau = nn.Parameter(torch.empty(1), requires_grad=trainable)
+            self.init_required = True
+        else:
+            self.log_tau = nn.Parameter(tau.log(), requires_grad=trainable)
+            self.init_required = False
+        if sigma is None:
+            self.log_sigma = nn.Parameter(torch.empty(1), requires_grad=trainable)
+            self.init_required = True
+        else:
+            self.log_sigma = nn.Parameter(sigma.log(), requires_grad=trainable)
+            self.init_required = False
+        self.init_fn_tau = init_fn_tau
+        self.init_fn_sigma = init_fn_sigma
+        self.active_dims = active_dims
+
+    @property
+    def tau(self) -> torch.Tensor:
+        return self.log_tau.exp()
+
+    @property
+    def sigma(self) -> torch.Tensor:
+        return self.log_sigma.exp()
+
+    def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch.Tensor]) -> torch.Tensor:
+        x, y = torch.as_tensor(x), torch.as_tensor(y)
+        dist = torch.sqrt(distance.squared_pairwise_distance(x.flatten(1), y.flatten(1)))
+        kernel_mat = torch.exp(-2 * torch.square(
+            torch.sin(torch.as_tensor(np.pi) * dist / self.tau)) / (self.sigma ** 2))
+        return kernel_mat
+
+
+class LocalPeriodic(nn.Module):
+    def __init__(
+        self,
+        tau: torch.Tensor = None,
+        init_fn_tau: Callable = None,
+        sigma: torch.Tensor = None,
+        init_fn_sigma: Callable = None,
+        trainable: bool = False,
+        active_dims: Optional[list] = None
+    ) -> None:
+        """
+        Local periodic kernel: k(x,y) = .
+        A forward pass takesa batch of instances x [Nx, features] and y [Ny, features]
+        and returns the kernel matrix [Nx, Ny].
+
+        Parameters
+        ----------
+        tau
+            Period of the periodic kernel.
+        sigma
+            Bandwidth used for the kernel.
+        """
+        super().__init__()
+        self.parameter_dict['tau'] = 'period'
+        self.parameter_dict['sigma'] = 'bandwidth'
+        if tau is None:
+            self.log_tau = nn.Parameter(torch.empty(1), requires_grad=trainable)
+            self.init_required = True
+        else:
+            self.log_tau = nn.Parameter(tau.log(), requires_grad=trainable)
+            self.init_required = False
+        if sigma is None:
+            self.log_sigma = nn.Parameter(torch.empty(1), requires_grad=trainable)
+            self.init_required = True
+        else:
+            self.log_sigma = nn.Parameter(sigma.log(), requires_grad=trainable)
+            self.init_required = False
+        self.init_fn_tau = init_fn_tau
+        self.init_fn_sigma = init_fn_sigma
+        self.active_dims = active_dims
+
+    @property
+    def tau(self) -> torch.Tensor:
+        return self.log_tau.exp()
+
+    @property
+    def sigma(self) -> torch.Tensor:
+        return self.log_sigma.exp()
+
+    def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch.Tensor]) -> torch.Tensor:
+        x, y = torch.as_tensor(x), torch.as_tensor(y)
+        dist = distance.squared_pairwise_distance(x.flatten(1), y.flatten(1))
+        kernel_mat = torch.exp(-2 * torch.square(
+            torch.sin(torch.as_tensor(np.pi) * dist / self.tau)) / (self.sigma ** 2)) * \
+            torch.exp(-0.5 * torch.square(dist / self.tau))
+        return kernel_mat
 
 
 class DeepKernel(nn.Module):


alibi_detect/cd/pytorch/lsdd.py
```diff
@@ -15,7 +15,8 @@ def __init__(
             preprocess_x_ref: bool = True,
             update_x_ref: Optional[Dict[str, int]] = None,
             preprocess_fn: Optional[Callable] = None,
-            sigma: Optional[np.ndarray] = None,
+            # sigma: Optional[np.ndarray] = None,
+            kernel: BaseKernel = GaussianRBF(),
```
ojcobb, line 19: Perhaps confusingly, LSDD isn't a kernel method for which one can plug in their own kernel. The method is specifically tied to the GaussianRBF kernel. Therefore we should go back to not offering a `kernel` argument.

ojcobb: (Obviously this applied across all the affected LSDD files. I won't duplicate the comment.)

alibi_detect/cd/pytorch/context_aware.py
```diff
@@ -4,13 +4,36 @@
...
 from alibi_detect.utils.pytorch import get_device
-from alibi_detect.utils.pytorch.kernels import GaussianRBF
+from alibi_detect.utils.pytorch.kernels import BaseKernel, GaussianRBF
 from alibi_detect.cd._domain_clf import _SVCDomainClf
 from tqdm import tqdm
 
 logger = logging.getLogger(__name__)
 
 
+def _sigma_median_diag(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.Tensor:
```
ojcobb, line 14: Not a huge fan of having this at the top of the script. Really I guess it belongs alongside the `sigma_median` function in utils anyway (perhaps even as an option within that function). Could be a good time to move (and then import) it.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -29,12 +36,76 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
-class GaussianRBF(nn.Module):
+class BaseKernel(nn.Module):
+    """
+    The base class for all kernels.
+    """
+    def __init__(self) -> None:
+        super().__init__()
+        self.parameter_dict: dict = {}
+        self.active_dims: Optional[list] = None
+        self.feature_axis: int = -1
```
ojcobb, line 47: As discussed on call, I would expose this behaviour via a wrapper that handles the subsetting and feeding of inputs into a desired underlying kernel. Otherwise we'll have to duplicate this logic within every future kernel, as will users wishing to define their own.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -29,12 +36,76 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
-        trainable: bool = False
+        kernel_a: BaseKernel,
+        kernel_b: BaseKernel
+    ) -> None:
+        super().__init__()
+        self.kernel_a = kernel_a
+        self.kernel_b = kernel_b
+
+    def forward(self, x: torch.Tensor, y: torch.Tensor,  infer_parameter: bool = False) -> torch.Tensor:
+        return (self.kernel_a(x, y, infer_parameter) + self.kernel_b(x, y, infer_parameter)) / 2
```
ojcobb, line 74: Nit pick but the `/2` is unexpected behaviour under the name `SumKernel`. I like having the `/2` as it ensures upper bounds remain known as kernels are added, but perhaps we name as `MeanKernel` or `AverageKernel` instead?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -29,12 +36,76 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+
+    def forward(self, x: torch.Tensor, y: torch.Tensor,  infer_parameter: bool = False) -> torch.Tensor:
+        return self.kernel_a(x, y, infer_parameter) * self.kernel_b(x, y, infer_parameter)
+
+
+class GaussianRBF(BaseKernel):
+    def __init__(
+       self,
+       sigma: Optional[torch.Tensor] = None,
+       init_fn_sigma: Callable = sigma_median,
```
ojcobb, line 105: Maybe we should have a standard kwarg name for this across all kernels. Perhaps now's a good time to change to something more generic like `param_initialiser`?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -54,30 +125,39 @@ def __init__(
             Whether or not to track gradients w.r.t. `sigma` to allow it to be trained.
         """
         super().__init__()
+        self.parameter_dict['sigma'] = 'bandwidth'
```
ojcobb, line 128: Perhaps I'm being stupid but I don't see what's going on here?

alibi_detect/utils/pytorch/kernels.py
ojcobb: Probs also a good time to achieve consistency by choosing only one of `infer` or `init`

alibi_detect/utils/pytorch/kernels.py
Srceh: No, you are right for questioning this. 

The original design purpose for this `parameter_dict` is to hold all relevant kernel parameter symbols and names in one place, so we can later implement a generical loop to initialise/train all parameters (i.e. `for params in list(parameter_dict)`)   

However, the current implementation wrote the initialisation for every parameter manually, so this dictionary bit might no longer be necessary.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -88,6 +168,271 @@ def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch
...
+        if active_dims is not None:
+            self.active_dims = torch.tensor(active_dims)
+        else:
+            self.active_dims = None
+        self.feature_axis = feature_axis
+        self.trainable = trainable
+
+    @property
+    def alpha(self) -> torch.Tensor:
+        return self.raw_alpha
```
ojcobb, line 220: Not sure why we need to define `raw_alpha` and this property?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -88,6 +168,271 @@ def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch
...
+            Exponent parameter of the kernel.
+        sigma
+            Bandwidth used for the kernel.
+        """
+        super().__init__()
+        self.parameter_dict['alpha'] = 'exponent'
+        self.parameter_dict['sigma'] = 'bandwidth'
+        if alpha is None:
+            self.raw_alpha = nn.Parameter(torch.empty(1), requires_grad=trainable)
+            self.init_required = True
```
ojcobb, line 199: At the moment it seems we would have the unexpected behaviour whereby if the user specified alpha but not sigma (or vice versa) then both get inferred from scratch?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -88,6 +168,271 @@ def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch
         return kernel_mat.mean(dim=0)  # [Nx, Ny]
 
 
+class RationalQuadratic(BaseKernel):
+    def __init__(
+        self,
+        alpha: torch.Tensor = None,
+        init_fn_alpha: Callable = pseudo_init_fn,
+        sigma: torch.Tensor = None,
```
ojcobb, line 176: Is this a sensible initialisation function to have as default? The name `pseudo` makes me think perhaps not? If so would be worth elaborating in the docstring of `pseudo_init_fn`. If not we should raise a warning or some other safeguard.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -88,6 +168,271 @@ def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch
...
+                raise ValueError("Length of sigma and alpha must be equal")
+        else:
+            kernel_mat = (1 + torch.square(dist) / (2 * self.alpha * (self.sigma ** 2))) ** (-self.alpha)
+        return kernel_mat
+
+
+class Periodic(BaseKernel):
+    def __init__(
+        self,
+        tau: torch.Tensor = None,
```
ojcobb, line 261: Would advocate for just having `period` as a more interpretable argument name here.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -88,6 +168,271 @@ def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch
...
+
+        if len(self.sigma) > 1:
+            if len(self.sigma) == len(self.tau):
+                kernel_mat = []
+                for i in range(len(self.sigma)):
+                    kernel_mat.append(torch.exp(-2 * torch.square(
+                        torch.sin(torch.as_tensor(np.pi) * dist / self.tau[i])) / (self.sigma[i] ** 2)))
+                kernel_mat = torch.stack(kernel_mat, dim=0).mean(dim=0)
+            else:
+                raise ValueError("Length of sigma and alpha must be equal")
```
ojcobb, line 339: Typo alpha -> tau

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -88,6 +168,271 @@ def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch
...
+            Bandwidth used for the kernel.
+        """
+        super().__init__()
+        self.parameter_dict['tau'] = 'period'
+        self.parameter_dict['sigma'] = 'bandwidth'
+        if tau is None:
+            self.log_tau = nn.Parameter(torch.empty(1), requires_grad=trainable)
+            self.init_required = True
+        else:
+            self.log_tau = nn.Parameter(tau.log(), requires_grad=trainable)
```
ojcobb, line 288: If we are going to the effort to make parameters trainable we should probably make it possible to train one and not the other, particularly given the period will (almost always?) be fixed according to domain specific knowledge whereas one may wish to train the bandwidth.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -88,6 +168,271 @@ def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch
...
+                kernel_mat = torch.stack(kernel_mat, dim=0).mean(dim=0)
+            else:
+                raise ValueError("Length of sigma and alpha must be equal")
+        else:
+            kernel_mat = torch.exp(-2 * torch.square(
+                torch.sin(torch.as_tensor(np.pi) * dist / self.tau)) / (self.sigma ** 2))
+        return kernel_mat
+
+
+class LocalPeriodic(BaseKernel):
```
ojcobb, line 346: Could this be defined as a product kernel of `Periodic` and `GaussianRBF`?

ojcobb:
Nice work! Indeed the previous coupling to `GaussianRBF` got a bit nasty in places and is nice to have got rid of all that and extended with some new possibilities!

A few things that may be worth further consideration:

1. At the moment specifying parameters, their initialisation functions and trainability is a bit tricky. It seems currently they must be all trainable or all fixed.  Moreover if some are specified and some aren't then they all get the initialisation function applied regardless. Is there something we can do to make this a bit cleaner?
2. Following on from the brief discussion we had on call -- I realise that using dunder methods to define the addition and multiplication of kernels syntactically raises the issue of making associated weights trainable. However given that the vast majority of use cases won't require trainable combinations it seems that perhaps a favourable approach is to proceed with the syntactic approach and then deal with trainability separately when desired. For example the deep kernel would be a lightweight class that defines a `composite_kernel = w_a * kernel_a + w_b * kernel_b` where `w_a` and `w_b` have been defined as trainable parameters.
3. I know we mentioned this before (and I commented above) but unless there's a compelling reason I'm not aware of we should try to remove the necessity to duplicate `active_dims` and `feature_axis` logic within each kernel.

ascillitoe:
Self-requesting a review for this since I expect this PR will require moderate changes to the save/load functionality.

alibi_detect/cd/pytorch/lsdd.py
Srceh: Fixed, now it just specifies RBF by default.

alibi_detect/cd/pytorch/context_aware.py
Srceh: The new parameter implementation assigns the default value of 1 during __init__, so this function is no longer needed.

alibi_detect/utils/pytorch/kernels.py
Srceh: This is now done with a `DimensionSelectKernel`. So no selection is required for each kernel.

alibi_detect/utils/pytorch/kernels.py
Srceh: Average kernel sounds better.

alibi_detect/utils/pytorch/kernels.py
Srceh: There is now an `infer_kernel_parameter` that does a loop over `parameter_dict`, so the overall behaviour is no longer dependent on the kernel.

alibi_detect/utils/pytorch/kernels.py
Srceh: There is now an `infer_kernel_parameter` that does a loop over `parameter_dict`, so the overall behaviour is no longer dependent on the kernel.

alibi_detect/utils/pytorch/kernels.py
Srceh: Was doing that for some data type consistency, removed as now there is a new parameter implementation.

alibi_detect/utils/pytorch/kernels.py
Srceh: With the new kernel implementation, each parameter can individually specify whether initialisation is required. The pre-set values can then be kept.

alibi_detect/utils/pytorch/kernels.py
Srceh: Now removed given the new parameter implementation.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -88,6 +168,271 @@ def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch
...
+            if self.trainable and infer_parameter:
+                raise ValueError("Gradients cannot be computed w.r.t. an inferred sigma value")
+            sigma = self.init_fn_sigma(x, y, dist)
+            alpha = self.init_fn_alpha(x, y, dist)
+            with torch.no_grad():
+                self.log_sigma.copy_(sigma.log().clone())
+                self.raw_alpha.copy_(alpha.clone())
+            self.init_required = False
+
+        if len(self.sigma) > 1:
```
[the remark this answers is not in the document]
Srceh: Fixed now, all follow the RBF implementation.

doc/source/examples/cd_mmd_cifar10.ipynb
```diff
@@ -765,11 +765,8 @@
   }
  ],
  "metadata": {
-  "interpreter": {
-   "hash": "ffba93b5284319fb7a107c8eacae647f441487dcc7e0323a4c0d3feb66ea8c5e"
-  },
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "Python 3.8.13 ('detect_cpu_py38')",
```
jklaise, line 769: Please revert changes to this notebook as no content has changed (note that in it's current form it will throw a warning for any user that doesn't have an environment named `detect_cpu_py38`.

alibi_detect/utils/pytorch/kernels.py
Srceh: Given we call 'bandwidth' as 'sigma', I was wondering if it is better to keep all parameters as greek letters for consistency.

alibi_detect/utils/pytorch/kernels.py
Srceh: The new parameter implementation should allow such behaviour. But at the moment for existing kernels, these are still specified by the global `trainable` argument. Guess the better solution here is to add a `trainable` option for each parameter.

alibi_detect/utils/pytorch/kernels.py
Srceh: Yes indeed, was writing it there just to make myself familiar with the kernel implementation logistic, will remove it later.

doc/source/examples/cd_mmd_cifar10.ipynb
Srceh: Apologies, staged this one by mistake, now fixed.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -5,6 +5,34 @@
 from typing import Optional, Union, Callable
 
 
+def infer_kernel_parameter(kernel, x, y, dist, infer_parameter):
```
jklaise, line 8: General remark - please add type hints.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +48,122 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+class AveragedKernel(nn.Module):
+    """
+    Construct a kernel by averaging two kernels.
+
+    Parameters:
+    ----------
+        kernel_a
+            the first kernel to be averaged.
+        kernel_b
+            the second kernel to be averaged.
```
jklaise, line 120: General remark - there should be no indentation in the numpy docstring format. If you're using PyCharm you can configure the docstring format so that it is automatically generated in the correct format when an empty dosctring is first created.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +48,122 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
     n = min(x.shape[0], y.shape[0])
     n = n if (x[:n] == y[:n]).all() and x.shape == y.shape else 0
     n_median = n + (np.prod(dist.shape) - n) // 2 - 1
     sigma = (.5 * dist.flatten().sort().values[n_median].unsqueeze(dim=-1)) ** .5
-    return sigma
+    return sigma.log()
 
 
-class GaussianRBF(nn.Module):
+class KernelParameter(object):
```
jklaise, line 60: Can be just `class KernelParameter:` as don't need specific inheritance from `object` in Python 3.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +48,122 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+        requires_init: bool = False
+    ) -> None:
+        super().__init__()
+        self.value = nn.Parameter(value if value is not None else torch.ones(1),
+                                  requires_grad=requires_grad)
+        self.init_fn = init_fn
+        self.requires_init = requires_init
+
+
+class BaseKernel(nn.Module):
```
jklaise, line 79: Not sure if we should make this an abstract base class (and the `forward` method an abstract method) or is it too much faff to get the multiple inheritance order right etc.? Same with the `tensorflow` equivalent. @ascillitoe

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +48,122 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
     sigma = (.5 * dist.flatten().sort().values[n_median].unsqueeze(dim=-1)) ** .5
-    return sigma
+    return sigma.log()
 
 
-class GaussianRBF(nn.Module):
+class KernelParameter(object):
+    """
+    Parameter class for kernels.
+    """
```
jklaise, line 63: General remark - missing constructor docstring for parameters here and in other classes.

@ascillitoe did we have a convention documenting class constructors either at the class level or at the `__init__` level? Does it matter for docs rendering?

alibi_detect/utils/pytorch/kernels.py
jklaise: I would rather be in favour of renaming `sigma` to `bandwidth` (is this PR backwards incompatible without renaming?) and `tau` to `period`.

alibi_detect/cd/base.py
```diff
@@ -462,7 +462,7 @@ def __init__(
             preprocess_x_ref: bool = True,
             update_x_ref: Optional[Dict[str, int]] = None,
             preprocess_fn: Optional[Callable] = None,
-            sigma: Optional[np.ndarray] = None,
+            # sigma: Optional[np.ndarray] = None,
```
jklaise, line 465: General remark - commented out code should be removed before the PR is final. Is there a particular reason to have these lines commented out during review rather than removed?

alibi_detect/utils/pytorch/kernels.py
jklaise: On the one hand I like the flexibility of the new `DimensionSelectKernel`, but on the other it leads to quite a bit of verbosity: 
image

Not sure if worth investigating other options here? E.g. the selection logic could be implemented as a `torch/tf` specific method in e.g. `TorchBaseKernel` and then it's application could be controlled via a parameter (could be the same `active_dims`) to `forward/call` - then custom kernels would be expected both to subclass but also if they want the selection capability to call `super().forward(x, y, active_dims)` to get the selected dimensions before applying custom logic.  This is just one idea, perhaps there are others. @arnaudvl @ojcobb

jklaise:
@Srceh is support for combining more than 2 kernels out of scope for this PR? Similarly for using the Python data model with dunder methods for adding/multiplying etc. kernels?

I actually prefer building up slow but if we have these extensions in mind in the future then would be good to clarify and take into account in the design.

arnaudvl:
> @Srceh is support for combining more than 2 kernels out of scope for this PR? Similarly for using the Python data model with dunder methods for adding/multiplying etc. kernels?
> 
> I actually prefer building up slow but if we have these extensions in mind in the future then would be good to clarify and take into account in the design.

I mentioned this earlier to @Srceh that we would want arbitrary amounts of kernels to be able to be combined, e.g. via passing `nn.ModuleList` with any amount of kernels (possibly incl. weights for weighted combinations) instead of 2 kernels.

alibi_detect/cd/context_aware.py
```diff
@@ -22,8 +23,10 @@ def __init__(
             preprocess_x_ref: bool = True,
             update_ref: Optional[Dict[str, int]] = None,
             preprocess_fn: Optional[Callable] = None,
-            x_kernel: Callable = None,
-            c_kernel: Callable = None,
+            # x_kernel: Callable = None,
+            x_kernel: BaseKernel = None,
+            # c_kernel: Callable = None,
+            c_kernel: BaseKernel = None,
```
jklaise, line 29: Have we fully discussed the implications of relying on the `BaseKernel` interface here? I suppose `Callable` is the most flexible thing for users as they can literally do anything, but it also doesn't specify the interface enough (i.e. need to take two parameters of a certain type and return something of a certain type). As far as I can see this change basically asks the following of a user wanting a custom kernel:
1. Subclass from `BaseKernel`
2. Implement `forward` according to the interface

This approach is ok, but perhaps we could even be less restrictive and only ask that a user specified kernel implements a `forward` method with the right types (this could be implemented using `typing.Protocol`). Of course this would not be enough if internally we rely on certain kernel attributes in which case we need to stick with the subclassing approach. @mauicv @ascillitoe

jklaise: As a follow-up, I see we rely on things like `kernel.trainable` checks which would rule out the less restrictive approach, just want to make sure we're fully happy with that.

alibi_detect/cd/mmd_online.py
```diff
@@ -17,8 +19,8 @@ def __init__(
             window_size: int,
             backend: str = 'tensorflow',
             preprocess_fn: Optional[Callable] = None,
-            kernel: Callable = None,
-            sigma: Optional[np.ndarray] = None,
+            kernel: Union[BaseKernelTorch, BaseKernelTF] = None,
```
jklaise, line 22: @mauicv @ascillitoe I feel like this is confusing for typing purposes and somehow unnecessarily indicates coupling between kernel interfaces and backends. Feel like there should be a top-level class that is backend-agnostic for typing purposes, perhaps a `Protocol` is valid, although not if we have to rely on inheriting from `BaseKernel`.

alibi_detect/cd/context_aware.py
jklaise: In fact the restriction to subclassing is even stronger than I thought because we're implicitly saying that kernels can only be either `nn.Module` or `tf.keras.Model` types (and coupled the kernel interface to the backend fairly strongly). 

Understand it stems mostly from the use of trainable kernels and using these restrictive interfaces makes it easier to deal with in the code (e.g. can always assume certain attributes exist), but could we be getting too restrictive? @arnaudvl @ojcobb

mauicv: Why does relying on kernel attributes restrict us here? I think protocols also work for [attributes](https://mypy.readthedocs.io/en/stable/protocols.html#using-isinstance-with-protocols)? The user would have to specify the attributes to match the protocol though.

jklaise: Didn't know that protocols support attributes.

Even so, would we have to be careful about the capabilities? E.g. device management via kernel attributes may not be relevant for user defined custom kernels?

alibi_detect/utils/pytorch/kernels.py
ascillitoe: [PEP 257](https://peps.python.org/pep-0257/) says:

> The class constructor should be documented in the docstring for its __init__ method.

But then the [NumPy style guide](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) says:


> The __init__ method may be documented in either the class level
> docstring, or as a docstring on the __init__ method itself.
> 
>  Either form is acceptable, but the two should not be mixed. Choose one
>  convention to document the __init__ method and be consistent with it.

We usually put the docstring in `__init__` so I guess you could say that is our convention...

ascillitoe: P.s. I shall add a note to `CONTRIBUTING.md`

alibi_detect/cd/context_aware.py
ascillitoe: Stating the obvious perhaps, but I think we need to think quite carefully about how a user might create their own custom subclassed kernel. For example, having abstract methods etc to make sure important methods are implemented (and error messages are clear), and thinking about what attributes the user needs to consider when building a custom kernel. 

Part of this might just be making sure we have good documentation around custom kernels, also `ABC`'s or `Protocol`'s could help...

On a side note, if we have a wider range of kernels and also support elementary operations to combine kernels, are we certain that we even need the capability for fully customised kernels?

alibi_detect/cd/mmd_online.py
ascillitoe: My initial idea for simplifying serialisation and perhaps the optional deps stuff was to look at backend-agnostic parent classes for kernels and even preprocessing (i.e. replace the combination of `preprocess_drift` and `partial` with a preprocessing class). These backend-agnostic classes would have `get_config` and `from_config` methods, and also do the dispatching to the relevant backend-specific classes. That being said I didn't get around to doing any sort of POC so it might be a non-starter... Doing something similar but only for typing, via a `Protocol`, sounds like an interesting one...

alibi_detect/utils/pytorch/kernels.py
ascillitoe: Agree it would be worth thinking about, assuming we're actually expecting users to sub-class `BaseKernel` to create their own custom kernels...

jklaise (commented):
Only some minor comments for now.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -80,90 +81,215 @@ class BaseKernel(nn.Module):
...
+        raise ValueError('Kernels can not be used as divisor.')
+
+    def __sub__(self, other):
+        raise ValueError('Kernels do not support substraction.')
+
+    def __rsub__(self, other):
+        raise ValueError('Kernels do not support substraction.')
+
+
+class SumKernel(nn.Module):
```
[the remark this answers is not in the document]
Srceh: The composite kernels behave a little different from the base class, as there is a defined `forward` method (the `BaseKernel `forward method requires an abstract method of `kernel_function`).

alibi_detect/cd/context_aware.py
arnaudvl: The code here still needs cleaning up (see commented out sections). I believe we are fine though on being fairly strict and imposing the `BaseKernel`?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -1,8 +1,44 @@
...
 import torch
 from torch import nn
 from . import distance
-from typing import Optional, Union, Callable
+from typing import Optional, Union, Callable, List
+from copy import deepcopy
+
+
+def infer_kernel_parameter(
+    kernel: 'BaseKernel',
```
arnaudvl, line 11: Type can just be changed to `BaseKernel`? @jklaise can you confirm?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -1,8 +1,44 @@
...
+    x
+        Tensor of instances with dimension [Nx, features].
+    y
+        Tensor of instances with dimension [Ny, features].
+    dist
+        Tensor with dimensions [Nx, Ny], containing the pairwise distances between `x` and `y`.
+    infer_parameter
+        Whether to infer the kernel parameter.
+    """
+    if kernel.trainable and infer_parameter:
```
arnaudvl, line 33: This relates back to @ojcobb 's comment on the possible requirement to set `trainable` on a parameter by parameter basis. It is possible that we do not allow for this just yet in this PR, but then a GitHub issue should be opened to follow up.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -1,8 +1,44 @@
...
+        raise ValueError("Gradients cannot be computed w.r.t. an inferred sigma value")
+    for parameter in kernel.parameter_dict.values():
```
arnaudvl, line 35: Is there a specific reason why we can't adhere to the usual way of initialising torch parameters, with custom init functions?

alibi_detect/utils/pytorch/tests/test_kernels_pt.py
```diff
@@ -3,7 +3,8 @@
 import pytest
 import torch
 from torch import nn
-from alibi_detect.utils.pytorch import GaussianRBF, DeepKernel
+from typing import Union
+from alibi_detect.utils.pytorch import GaussianRBF, DeepKernel, BaseKernel, RationalQuadratic, Periodic
```
arnaudvl, line 8: Missing explicit test for `infer_kernel_parameter`

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +56,307 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
     -------
-    The computed bandwidth, `sigma`.
+    The logrithm of the computed bandwidth, `log-sigma`.
     """
     n = min(x.shape[0], y.shape[0])
     n = n if (x[:n] == y[:n]).all() and x.shape == y.shape else 0
     n_median = n + (np.prod(dist.shape) - n) // 2 - 1
     sigma = (.5 * dist.flatten().sort().values[n_median].unsqueeze(dim=-1)) ** .5
-    return sigma
+    return sigma.log()
```
arnaudvl, line 65: What is the reason why this is changed within the function? Can't this just be done afterwards? Because it is used in various places, so would need very careful checking, and the name of the function should become `log_sigma_median` or something similar for clarity instead? The behaviour would also deviate from `alibi_detec.cd.pytorch.context_aware._sigma_median_diag`, which leads to confusion.

alibi_detect/utils/pytorch/kernels.py
arnaudvl: Any conclusion on this?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +56,307 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+            The pre-specified value of the parameter.
+        init_fn
+            The function used to initialize the parameter.
+        requires_grad
+            Whether the parameter requires gradient.
+        requires_init
+            Whether the parameter requires initialization.
+        """
+        super().__init__()
+        self.value = nn.Parameter(value if value is not None else torch.ones(1),
```
arnaudvl, line 91: Why is the default value of any parameter if not specified 1?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +56,307 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+        active_dims
+            Indices of the dimensions of the feature to be used for the kernel. If None, all dimensions are used.
+        feature_axis
+            Axis of the feature dimension.
+        """
+        super().__init__()
+        self.parameter_dict: dict = {}
+        if active_dims is not None:
+            self.active_dims = torch.as_tensor(active_dims)
+        else:
```
arnaudvl, line 113: Redundant `else`

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +56,307 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+        """
+        super().__init__()
+        self.value = nn.Parameter(value if value is not None else torch.ones(1),
+                                  requires_grad=requires_grad)
+        self.init_fn = init_fn
+        self.requires_init = requires_init
+
+
+class BaseKernel(nn.Module):
+    def __init__(self, active_dims: list = None, feature_axis: int = -1) -> None:
```
arnaudvl, line 98: As mentioned by @ojcobb [here](https://github.com/SeldonIO/alibi-detect/pull/558#issuecomment-1208162385), it is unclear why both `active_dims` and `feature_axis` are present. The parameter description also doesn't clarify this and should be more descriptive.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +56,307 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+        self.parameter_dict: dict = {}
+        if active_dims is not None:
+            self.active_dims = torch.as_tensor(active_dims)
+        else:
+            self.active_dims = None
+        self.feature_axis = feature_axis
+        self.init_required = False
+
+    @abstractmethod
+    def kernel_function(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch.Tensor],
```
arnaudvl, line 119: As we are in the `torch` specific submodule, I would actually fairly strongly argue that `np.ndarray` inputs should not be allowed. Note that the current implementation (e.g. line 125) would lead to device placement issues?

alibi_detect/utils/pytorch/kernels.py
arnaudvl: What would be a concrete example of where `feature_axis != -1`?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +56,307 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+            x = torch.index_select(x, self.feature_axis, self.active_dims)
+            y = torch.index_select(y, self.feature_axis, self.active_dims)
+        if len(self.parameter_dict) > 0:
+            return self.kernel_function(x, y, infer_parameter)
+        else:
+            return self.kernel_function(x, y)
+
+    def __add__(
+        self,
+        other: Union['BaseKernel', 'SumKernel', 'ProductKernel', torch.Tensor]
```
arnaudvl, line 136: Same comment as at the top of this file: any reason why `"BaseKernel"` and not just `BaseKernel` is used as type? Silly question maybe, so just want to know :) @jklaise ?

arnaudvl: Personally don't like `other` as a argument name, especially given absence of docstrings it is not easy to understand the logic straight away. It would be helpful to have a more clear name.

alibi_detect/cd/context_aware.py
arnaudvl: Actually it seems like my above comment is not correct and `nn.Module` is the required restriction since e.g. SumKernel or ProductKernel do not inherit from BaseKernel?

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +56,307 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+    ) -> 'SumKernel':
+        if isinstance(other, SumKernel):
+            other.kernel_list.append(self)
+            return other
+        elif (isinstance(other, BaseKernel) or
```
arnaudvl, line 141: Can shorten to `isinstance(other, (BaseKernel, ProductKernel, torch.Tensor))`

alibi_detect/utils/pytorch/kernels.py
jklaise: Yea I can't see why not (perhaps an ordering issue as `BaseKernel` is defined further down?).

alibi_detect/utils/pytorch/kernels.py
jklaise: I think `other` is a standard name for implementing `__add__` (and other (!) Python data model methods), though if a different name would make the code clearer then I don't see a reason not to do it (perhaps the code needs some comments anyway, explaining the handling of sum kernels?).

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +56,307 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+        if isinstance(other, torch.Tensor):
+            return self.__mul__(1. / other)
+        else:
+            raise ValueError('Kernels can only be divided by a constant.')
+
+    def __rtruediv__(self, other):
+        raise ValueError('Kernels can not be used as divisor.')
+
+    def __sub__(self, other):
+        raise ValueError('Kernels do not support substraction.')
```
arnaudvl, line 188: Repeated typo: subtraction

alibi_detect/utils/pytorch/kernels.py
Srceh: Yes indeed, as it is defined below, mypy would suggest an error without the quote.

alibi_detect/utils/pytorch/kernels.py
arnaudvl: Ok, got it, didn't know that. Clarifying comments would be helpful in that case.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +56,307 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+            Whether the parameter requires initialization.
+        """
+        super().__init__()
+        self.value = nn.Parameter(value if value is not None else torch.ones(1),
+                                  requires_grad=requires_grad)
+        self.init_fn = init_fn
+        self.requires_init = requires_init
+
+
+class BaseKernel(nn.Module):
```
arnaudvl, line 97: For sum/product/... kernels, can we abstract the operation away roughly as follows (with an extension to `torch.Tensor` + `__add__` etc)? Or is it not worth abstracting this away?

```python
class CombinedKernel(nn.Module):
   def __init__(self, kernels: nn.ModuleList, operation: str = 'sum') -> None:
      super().__init__()
      self.kernels = kernels
      self.operation = get_operation(operation)

   def forward(self, x: torch.Tensor, y: torch.Tensor, infer_parameters: bool = False) -> torch.Tensor:
      return torch.stack([k(x, y, infer_parameters) for k in self.kernels]).self.operation(dim=0)

   def __add__(self, other):
    
   ...

class SumKernel(CombinedKernel):
   def __init__(self, kernels: nn.ModuleList):
      super().__init__(kernels, 'sum')
```

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -20,21 +56,307 @@ def sigma_median(x: torch.Tensor, y: torch.Tensor, dist: torch.Tensor) -> torch.
...
+        for k in self.kernel_list:
+            if isinstance(k, BaseKernel) or isinstance(k, SumKernel) or isinstance(k, ProductKernel):
+                value_list.append(k(x, y, infer_parameter))
+            elif isinstance(k, torch.Tensor):
+                value_list.append(k * torch.ones((x.shape[0], y.shape[0])))
+            else:
+                raise ValueError(type(k) + 'is not supported by SumKernel.')
+        return torch.sum(torch.stack(value_list), dim=0)
+
+    def __add__(
```
arnaudvl, line 214: From here onwards we need lots of comments/docstrings for maintainability. Otherwise this might turn out to be a mess quite quickly.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -52,43 +374,217 @@ def __init__(
...
     def sigma(self) -> torch.Tensor:
-        return self.log_sigma.exp()
+        return self.parameter_dict['log-sigma'].value.exp()
 
-    def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch.Tensor],
-                infer_sigma: bool = False) -> torch.Tensor:
+    def kernel_function(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch.Tensor],
+                        infer_parameter: bool = False) -> torch.Tensor:
 
         x, y = torch.as_tensor(x), torch.as_tensor(y)
```
arnaudvl, line 399: SImilar comment as before: this could lead to device placement issues.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -52,43 +374,217 @@ def __init__(
...
-        if sigma is None:
-            self.log_sigma = nn.Parameter(torch.empty(1), requires_grad=trainable)
-            self.init_required = True
-        else:
-            sigma = sigma.reshape(-1)  # [Ns,]
-            self.log_sigma = nn.Parameter(sigma.log(), requires_grad=trainable)
-            self.init_required = False
-        self.init_sigma_fn = init_sigma_fn
+        super().__init__(active_dims, feature_axis)
+        self.parameter_dict['log-sigma'] = KernelParameter(
```
arnaudvl, line 383: General question: is it good practice to have hyphenated key values as parameters? @jklaise

alibi_detect/utils/pytorch/kernels.py
arnaudvl: A key part would also be to clarify how `[m,n]` kernel matrices are spliced for mixed-type data with `active_dims` on subsets of the features (e.g. 3 numerical and 4 categorical) and where we use combined kernels (e.g. a product kernel).

alibi_detect/utils/pytorch/kernels.py
jklaise: I'd stick to underscores as it's a standard.

arnaudvl:
Are none of the other examples affected at all? They still work with the updated kernels producing the same results? Would be great if true, but it sounds almost too good to be true...

alibi_detect/utils/pytorch/kernels.py
Srceh: Currently, we allow each parameter to be configured separately on `trainable` for a torch and `requires_grad` for tf.  This `kernel.trainable` was used to provide an accessble attribute for other methods to determine the training requirements. Guess it would be better here that I just do a quick implementation that checks all the parameters and returns if at least one parameter requires training.

alibi_detect/utils/pytorch/kernels.py
Srceh: Here I just followed the previous error message, which doesn't allow both `trainable` and `infer_parameter`. Should be easily changed if allowing both is the desired behaviour.

alibi_detect/utils/pytorch/tests/test_kernels_pt.py
Srceh: Will add.

alibi_detect/utils/pytorch/kernels.py
Srceh: I aimed to declare all sigma parameters directly as `log-sigma` for the non-negative constraint of sigma. But we can go back to the direct sigma to avoid confusion.

alibi_detect/utils/pytorch/kernels.py
Srceh: More or less treating it as a placeholder and assuming the user would go for `infer` or `trainable`, guess I should make this behaviour clear.

alibi_detect/utils/pytorch/kernels.py
Srceh: I thought if we use an image batch with shape`[n_batch, n_pixel, n_colour]`, if we want to select a sub-patch of every image across all images, we might want `feature_axis=1` and 'active_dims=[...]'.

alibi_detect/utils/pytorch/kernels.py
Srceh: Good pick. I initially followed the previous implementation and used extra device manipulation to bypass the issue (now removed). Will rewrite it to torch tensor only and double-check things.

Srceh:
> Are none of the other examples affected at all? They still work with the updated kernels producing the same results? Would be great if true, but it sounds almost too good to be true...

Should be fine on the pytest level, but might be worth some further investigation.

alibi_detect/cd/lsdd.py
```diff
@@ -18,7 +18,8 @@ def __init__(
             preprocess_x_ref: bool = True,
             update_x_ref: Optional[Dict[str, int]] = None,
             preprocess_fn: Optional[Callable] = None,
-            sigma: Optional[np.ndarray] = None,
+            # sigma: Optional[np.ndarray] = None,
+            # kernel: BaseKernel = None,
```
ascillitoe, line 22: Is the `kernel` here erroneous?

alibi_detect/cd/pytorch/context_aware.py
```diff
@@ -22,8 +45,8 @@ def __init__(
             preprocess_x_ref: bool = True,
             update_ref: Optional[Dict[str, int]] = None,
             preprocess_fn: Optional[Callable] = None,
-            x_kernel: Callable = GaussianRBF,
-            c_kernel: Callable = GaussianRBF,
+            x_kernel: BaseKernel = GaussianRBF(init_fn_sigma=_sigma_median_diag),
```
ascillitoe, line 48: Init'ing `GaussianRBF` with kwargs via the defaults here does make deserialization a little more cumbersome...

alibi_detect/cd/tensorflow/context_aware.py
```diff
@@ -4,13 +4,36 @@
...
 from alibi_detect.cd.base import BaseContextMMDDrift
-from alibi_detect.utils.tensorflow.kernels import GaussianRBF
+from alibi_detect.utils.tensorflow.kernels import GaussianRBF, BaseKernel
 from alibi_detect.cd._domain_clf import _SVCDomainClf
 from tqdm import tqdm
 
 logger = logging.getLogger(__name__)
 
 
+def _sigma_median_diag(x: tf.Tensor, y: tf.Tensor, dist: tf.Tensor) -> tf.Tensor:
```
ascillitoe, line 14: Why is this moved?

alibi_detect/cd/lsdd.py
Srceh: yes, thanks for noticing, will remove all these redundant bits. (and probably use some unique character for these comment-outs next time for easy search...)

alibi_detect/cd/tensorflow/context_aware.py
Srceh: Probably due to a linter error of an undefined function if my memory is correct.

alibi_detect/utils/pytorch/kernels.py
ascillitoe: IMO it would be more consistent with the rest of the library to do `class BaseKernel(nn.Module, ABC):` and:

```
    @abstractmethod
    def kernel_function(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch.Tensor],
                        infer_parameter: Optional[bool] = False) -> torch.Tensor:
        pass
```

rather than raise `NotImplementedError`. I believe `ABC` behaves quite nicely as a mix-in class. 

I'm not quite so sure whether we need to make `forward` an abstract method, since it is already defined in `BaseKernel`. Either way, the `cd_create_customised_kernel.ipynb` example is really nice for this, but we will need additional documentation on how to make a custom kernel (and possibly how `feature_axis` etc fit into this).

alibi_detect/cd/context_aware.py
jklaise: Looks like the above needs more thought, allowing `nn.Module` basically removes all advantages of having `BaseKernel`. Not sure if product/sum kernels should inherit also from `BaseKernel` (or perhaps they should be of some type `CombinedKernel` and both base+combined are allowed in the detectors?).

alibi_detect/cd/pytorch/context_aware.py
arnaudvl: Yes ideally we can do the same across all detectors.

alibi_detect/utils/pytorch/kernels.py
ascillitoe: Comment based on offline discussion:

I believe with the current design, performing elementary operations on `BaseKernel`'s, for example the following:

```python
Kernel_avg = (Kernel_0 + Kernel_1) / torch.tensor(2.0)
```

will return something that is no longer ` BaseKernel` (`Kernel_avg` will be a `SumKernel` I think?). This means lots of the code in places such as `MMDDrift` is no longer quite accurate (type hints, `isinstance` checks etc), since we are expecting a `BaseKernel`. It might be worth rethinking the top-level kernel classes (e.g. maybe `SingleKernel` and `CombinedKernel`, which both inherit from a very generic `BaseKernel`). This might also make serialization easier.

ascillitoe: Regarding serialization, we need to decide how much configurability we want in the `config.toml`. Assuming we have:

```python
Kernel_avg = (Kernel_0 + Kernel_1) / torch.tensor(2.0)
```

the first option is to attempt to serialize the individual kernels separately, and save the expression, so the toml would look like:

```toml
kernel_0 = '/kernels/kernel_0'  # filepath to 1st serialized kernel
kernel_1 = '/kernels/kernel_1'  # filepath to 2nd serialized kernel
kernel = '(kernel_0 + kernel_1)/2'  # expression defining combined kernel
```

The second option is to simply serialize the entire combined kernel. Hopefully this will be simple if `SumKernel` etc are `nn.Module` and `tf.keras.Model`. 

My preference is for the latter, since the former will likely get rather complex!

PR #564, opened 2022-07-22
Add detector type to metadata

ascillitoe:
This PR updates the detector `meta` attribute to specify whether it is a drift, adversarial or outlier detector. The information is saved in `detector_type`. Previously, `detector_type` was used to specify 'offline' vs 'online' detectors. This is now done via the `online` field.

The detector `schemas` and `_set_config`/`get_config` methods are also updated so that the full detector metadata is saved in the `config.toml` when `save_detector` is called.

**Edit**: Due to the duplication of fields resulting from writing `self.meta` to config (i.e. `name`, `backend`, `data_type` and possibly other kwarg's), we will avoid writing it to config for now. We should rethink what is included in the config `meta` at a later date. See https://github.com/SeldonIO/alibi-detect/issues/567.


alibi_detect/base.py
```diff
@@ -126,6 +127,9 @@ def get_config(self) -> dict:  # TODO - move to BaseDetector once config save/lo
             # Set x_ref_preprocessed flag
             preprocess_at_init = getattr(detector, 'preprocess_at_init', True)  # If no preprocess_at_init, always true!
             cfg['x_ref_preprocessed'] = preprocess_at_init and detector.preprocess_fn is not None
+            # Add in detector metadata
+            cfg['meta'] = self.meta  # type: ignore[attr-defined]
+            cfg['meta']['config_spec'] = __config_spec__
```
jklaise, line 132: Was this really missing previously..?

alibi_detect/saving/loading.py
```diff
@@ -527,7 +527,7 @@ def _replace(cfg: dict, orig: Optional[str], new: Optional[str]) -> dict:
     The updated dictionary.
     """
     for k, v in cfg.items():
-        if v == orig:
+        if isinstance(v == orig, bool) and v == orig:
```
jklaise, line 530: why this change?

alibi_detect/saving/schemas.py
```diff
@@ -73,10 +72,15 @@ class Config:
...
 
-class MetaData(CustomBaseModel):
-    version: str = __version__
-    config_spec: str = __config_spec__
-    version_warning: Optional[bool] = None  # None instead of False as same effect in logicals but hides from config
+class MetaData(CustomBaseModelWithKwargs):
+    name: str
+    online: bool
+    data_type: Optional[str]
+    detector_type: str
```
jklaise, line 79: should we add the `Literal` type here?

alibi_detect/saving/schemas.py
```diff
@@ -87,7 +91,7 @@ class DetectorConfig(CustomBaseModel):
     "Name of the detector e.g. `MMDDrift`."
     backend: Literal['tensorflow', 'pytorch', 'sklearn'] = 'tensorflow'
     "The detector backend."
-    meta: Optional[MetaData] = MetaData()
+    meta: Optional[MetaData]
```
jklaise, line 94: Why change from setting to `MetaData()` ?

jklaise (commented):
LGTM bar a few comments. Note this would be a breaking change for `detector_type` metadata entry, will need @arnaudvl blessing!

alibi_detect/saving/loading.py
ascillitoe: `_replace` loops through the entire `cfg`, which is a `dict` potentially containing numerous nested `dict`'s. For each value (`v`) in the `cfg` it compares it to `orig` and swaps it for `new` if it `v==orig`. 

The change to include the entire `self.meta` attribute in the config introduced an unexpected edge case for the above. `SpotTheDiffDrift` adds a number of non-standard items to `self.meta`:
https://github.com/SeldonIO/alibi-detect/blob/28aee232e8e121a30f1860814c6ea6f71419fb80/alibi_detect/cd/tensorflow/spot_the_diff.py#L161-L165

Some of these are `ndarray`'s, which broke the `if v == orig:` logic, since the `ndarray` `__eq__` returns an array. I've replaced it with the above so that we only check `v == orig` if `v == orig` returns a `bool`. This shouldn't affect the operation of `_replace` since it is a private function and we only use to for simple operations such as replacing `None` with `'None'`.

alibi_detect/saving/schemas.py
ascillitoe: @jklaise I actually just realised, I agree we should update to `Literal`. Only issue is the docstrings for `data_type` are out of date throughout the library i.e. we say `Optionally specify the data type (tabular, image or time-series)`. 

We should probably agree on all the permitted options and update the docstrings/docs at the same time. I can do as part of this PR if you think its worth it?

Options I can think of are `tabular`, `image`, `time-series`, `text`, `graph`. `graph` isn't included in our [algo overview](https://docs.seldon.io/projects/alibi-detect/en/stable/overview/algorithms.html) so don't know if we should include...

alibi_detect/saving/schemas.py
ascillitoe: The original was sort of a hack to get `pydantic` to set the `self.config['meta']` field. This is now set in its entirety by `_set_config` and `get_config`. IMO `meta: Optional[MetaData]` is more natural, and this is how we do all the other pydantic nested models like `ModelConfig` etc.

ascillitoe:
> LGTM bar a few comments. Note this would be a breaking change for `detector_type` metadata entry, will need @arnaudvl blessing!

@arnaudvl has just given his blessing offline 🙌🏻

ascillitoe:
**Note**: We should wait until #537 is merged and rerun CI before merging this.

alibi_detect/saving/loading.py
jklaise: It sounds like adding `self.meta` to config is potentially a bigger thing than the scope of the original issue? Perhaps we should reduce the scope here and re-think later (although if things appear ok then maybe it's good to press forward)?

What are the main benefits from adding `self.meta` to the config? Conscious that some of the information would be duplicated as `self.meta` may contain user-passed parameters which would be in the config anyway.

alibi_detect/saving/schemas.py
jklaise: Would prefer a different PR so as to keep things separate. I'm not sure how the docstrings currently relate to the functionality of passing `data_type`, i.e. what does a user specify exactly if we haven't even defined what they can do?

alibi_detect/saving/schemas.py
```diff
@@ -74,9 +73,9 @@ class Config:
 
 
 class MetaData(CustomBaseModel):
-    version: str = __version__
-    config_spec: str = __config_spec__
-    version_warning: Optional[bool] = None  # None instead of False as same effect in logicals but hides from config
+    version: str
+    config_spec: str
+    version_warning: bool = False
```
ascillitoe, line 78: `version_warning` changed because setting to `None` no longer hides from config (since we now convert `None` to `"None"` when writing to `config.toml`, instead of it being removed).

ascillitoe:
@jklaise I have removed the changes to `alibi_detect/base.py` that resulted in `self.meta` being written to the config. The `meta` section in the `config.toml` now just looks like this again:

```toml
[meta]
version = "0.10.0rc1"
config_spec = "0.1"
version_warning = false
```

alibi_detect/base.py
```diff
@@ -155,26 +152,16 @@ def from_config(cls, config: dict):
         return detector
 
     def _set_config(self, inputs):  # TODO - move to BaseDetector once config save/load implemented for non-drift
-        """
```
jklaise: Should we keep some part of this docstring?

alibi_detect/saving/loading.py
ascillitoe: The intention behind this was so that fields like `detector_type` could be viewed without needing to instantiate the detector. Based on offline discussion @jklaise and I have decided to remove `self.meta` from config for now, and will review at a later date (https://github.com/SeldonIO/alibi-detect/issues/567).

alibi_detect/saving/schemas.py
ascillitoe: We should also review the need and use case behind the `data_type` kwarg. I've included this in https://github.com/SeldonIO/alibi-detect/issues/567.

alibi_detect/base.py
ascillitoe: Good point, overly aggressive pruning on my part. Will add back in!

jklaise:
LGTM, remember to update PR title and will need a **breaking change** entry in the changelog for the release.

ascillitoe:
@jklaise Since we're releasing imminently I've updated the CHANGELOG as part of this PR.

PR #565, opened 2022-07-25
Pin sphinx <5.1.0

mauicv:
[CI failing due to `sphinx==5.1.0`](https://github.com/sphinx-doc/sphinx/issues/10701). Pin to less than.


ascillitoe (approved):
LGTM. We should check that dependabot monitors minor `sphinx` releases going foward.

ascillitoe:
@mauicv FYI dependabot is now tracking minor releases like we want: https://github.com/SeldonIO/alibi-detect/pull/571

PR #591, opened 2022-08-16
Fix to allow config.toml to be loaded with [meta] not present

ascillitoe:
https://github.com/SeldonIO/alibi-detect/pull/564 introduced a bug whereby `config.toml` files were not loadable via `load_detector` if the `[meta]` field was not present. The `[meta]` field is only present when `config.toml` files are generated by `save_detector`, hence why this was not picked up during previous CI.

This PR fixes this issue, and adds tests to check that `validate_config` and `load_detector` both work when `[meta]` is not present.

This will be released in a `v0.10.3` patch once merged.

### DIFF
--- a/alibi_detect/base.py
+++ b/alibi_detect/base.py
@@ -146,7 +146,7 @@ class DriftConfigMixin:
         """
         # Check for existing version_warning. meta is pop'd as don't want to pass as arg/kwarg
         meta = config.pop('meta', None)
-        meta = {} if meta is None else meta
+        meta = {} if meta is None else meta  # Needed because pydantic sets meta=None if it is missing from the config
         version_warning = meta.pop('version_warning', False)
         # Init detector
         detector = cls(**config)
--- a/alibi_detect/base.py
+++ b/alibi_detect/base.py
@@ -144,8 +144,10 @@ class DriftConfigMixin:
         config
             A config dictionary matching the schema's in :class:`~alibi_detect.saving.schemas`.
         """
-        # Check for exisiting version_warning. meta is pop'd as don't want to pass as arg/kwarg
-        version_warning = config.pop('meta', {}).pop('version_warning', False)
+        # Check for existing version_warning. meta is pop'd as don't want to pass as arg/kwarg
+        meta = config.pop('meta', None)
+        meta = {} if meta is None else meta
+        version_warning = meta.pop('version_warning', False)
         # Init detector
         detector = cls(**config)
         # Add version_warning
--- a/alibi_detect/saving/tests/test_saving.py
+++ b/alibi_detect/saving/tests/test_saving.py
@@ -270,12 +270,13 @@ def preprocess_hiddenoutput(classifier, backend):
 
 @parametrize('cfg', CFGS)
 def test_load_simple_config(cfg, tmp_path):
-    x_ref_path = tmp_path.joinpath('x_ref.npy')
-    cfg_path = tmp_path.joinpath('config.toml')
+    save_dir = tmp_path
+    x_ref_path = str(save_dir.joinpath('x_ref.npy'))
+    cfg_path = save_dir.joinpath('config.toml')
     # Save x_ref in config.toml
     x_ref = cfg['x_ref']
     np.save(x_ref_path, x_ref)
-    cfg['x_ref'] = str(x_ref_path)
+    cfg['x_ref'] = 'x_ref.npy'
     # Save config.toml then load it
     with open(cfg_path, 'w') as f:
         toml.dump(cfg, f)
--- a/alibi_detect/saving/tests/test_saving.py
+++ b/alibi_detect/saving/tests/test_saving.py
@@ -10,6 +10,7 @@ from functools import partial
 from pathlib import Path
 from typing import Callable
 
+import toml
 import dill
 import numpy as np
 import pytest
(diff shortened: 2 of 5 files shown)


ascillitoe:
@jklaise @mauicv would one of you be able to review this, please?

The MacOS test has hung after completion, but all tests have passed.

alibi_detect/saving/tests/test_saving.py
```diff
@@ -259,6 +268,22 @@ def preprocess_hiddenoutput(classifier, backend):
     return preprocess_fn
 
 
+@parametrize('cfg', CFGS)
+def test_load_simple_config(cfg, tmp_path):
```
jklaise, line 272: Worth adding a docstring saying exactly what this is testing, i.e. not using the `save_detector` functionality here as that adds `[meta]` field automatically?

alibi_detect/saving/tests/test_saving.py
```diff
@@ -259,6 +268,22 @@ def preprocess_hiddenoutput(classifier, backend):
...
+    x_ref_path = str(save_dir.joinpath('x_ref.npy'))
+    cfg_path = save_dir.joinpath('config.toml')
+    # Save x_ref in config.toml
+    x_ref = cfg['x_ref']
+    np.save(x_ref_path, x_ref)
+    cfg['x_ref'] = 'x_ref.npy'
+    # Save config.toml then load it
+    with open(cfg_path, 'w') as f:
+        toml.dump(cfg, f)
+    cd = load_detector(cfg_path)
```
jklaise, line 283: Is it necessary to dump to `.toml` first instead of calling `load_detector` on the dictionary? I suppose if we're strictly testing loading of `.toml` then it's ok.

alibi_detect/saving/tests/test_saving.py
```diff
@@ -259,6 +268,22 @@ def preprocess_hiddenoutput(classifier, backend):
...
+    assert cd.__class__.__name__ == cfg['name']
```
jklaise, line 284: Is there a way to assert "more", e.g. get the config dict?

alibi_detect/saving/validate.py
```diff
@@ -38,6 +38,7 @@ def validate_config(cfg: dict, resolved: bool = False) -> dict:
 
     # Get meta data
     meta = cfg.get('meta')
+    meta = {} if meta is None else meta
```
jklaise, line 41: Why not `meta = cfg.get('meta', {})`? Ease of reading?

alibi_detect/base.py
```diff
@@ -144,8 +144,10 @@ def from_config(cls, config: dict):
         config
             A config dictionary matching the schema's in :class:`~alibi_detect.saving.schemas`.
         """
-        # Check for exisiting version_warning. meta is pop'd as don't want to pass as arg/kwarg
-        version_warning = config.pop('meta', {}).pop('version_warning', False)
+        # Check for existing version_warning. meta is pop'd as don't want to pass as arg/kwarg
+        meta = config.pop('meta', None)
+        meta = {} if meta is None else meta
```
jklaise, line 149: Why not `meta = config.pop('meta', {})`? Ease of reading?

alibi_detect/saving/tests/test_validate.py
```diff
@@ -16,19 +16,14 @@
     'x_ref': np.array([[-0.30074928], [1.50240758], [0.43135768], [2.11295779], [0.79684913]]),
     'p_val': 0.05
 }
-cfgs = [mmd_cfg]
-n_tests = len(cfgs)
 
+# Define a detector config dict without meta (as simple as it gets!)
+mmd_cfg_nometa = mmd_cfg.copy()
```
jklaise, line 21: Do we need to worry about shallow copies and shared objects here?

jklaise (approved):
Just a few questions from me.

alibi_detect/saving/tests/test_saving.py
ascillitoe: I agree with you on this one, being able to pass config dicts to `load_detector` is something I think is nice. However, I can't recall why now but sometime back we decided to remove this capability (I think just to reduce complexity). The function signature for `load_detector` is currently ` load_detector(filepath: Union[str, os.PathLike], **kwargs)`.

I'd quite like to revisit this at some point...

alibi_detect/saving/validate.py
ascillitoe: This bit was the root of the bug. The pydantic schema for `DetectorConfig` sets `meta` to `None` if it doesn't exist in the `config.toml`. Since `meta` does exist in `cfg` (but equals `None`), `cfg.get('meta', {})` will still return `None`, and then line 42 fails.

alibi_detect/base.py
ascillitoe: See previous comment.

alibi_detect/saving/validate.py
jklaise: Is this something worth leaving a comment in the code for (to prevent overzealous code "clean-up")?

ascillitoe: Good idea. Added a comment to this and the line in `base.py`.

alibi_detect/saving/tests/test_validate.py
ascillitoe: I did check and it doesn't seem to be an issue. Which is confusing since I think it would be for a list i.e. the pop-ing some from `mmd_cfg_nometa` would have removed it from `mmd_cfg` too 🤷🏻‍♂️ 

I've changed it to `deepcopy` anyway just to be on the safe side...

jklaise (approved):
LGTM

PR #595, opened 2022-08-22
Update CONTRIBUTING.md

ascillitoe:
Update `CONTRIBUTING.md` with a note on where to add class docstrings. 

Will do the same for `alibi` and delete the `contributing-update` branch upon merge.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,7 +37,8 @@ options are defined in `setup.cfg`.
 We adhere to the `numpy` style docstrings (https://numpydoc.readthedocs.io/en/stable/format.html)
 with the exception of ommiting argument types in docstrings in favour of type hints in function
 and class signatures. If you're using a `PyCharm`, you can configure this under
-`File -> Settings -> Tools -> Python Integrated Tools -> Docstrings`.
+`File -> Settings -> Tools -> Python Integrated Tools -> Docstrings`. For classes, we adhere to the convention
+of including docstrings in their `__init__` method, rather than as a class level docstring.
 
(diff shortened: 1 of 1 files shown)


mauicv:
Shall we add a note for those of us who now us VS-code? :laughing: I'm using [this](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring) extension

mauicv:
Otherwise LGTM!

mauicv (approved):

mauicv:
Actually another question: How do we document static classes or classes that don't implement an `__init__` method? The similarity method backends do [this](https://github.com/SeldonIO/alibi/blob/master/alibi/explainers/similarity/backends/pytorch/base.py) for example.

ascillitoe:
> Shall we add a note for those of us who now us VS-code on how to set up auto docstrings? laughing I'm using [this](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring) extension

Yuck... but yes good idea.



> Otherwise LGTM! Although why was the docs CI cancelled?

I don't know... perhaps because I authored the PR via the github UI and it doesn't touch the codebase?



> Actually another question: How do we document static classes or classes that don't implement an `__init__` method? The similarity method backends do [this](https://github.com/SeldonIO/alibi/blob/master/alibi/explainers/similarity/backends/pytorch/base.py) for example.

Good question. Maybe the convention should be "add docstring in __init__ if it exists, otherwise add to the class-level docstring"?

ascillitoe:
@mauicv just updated following your comments.

PR #596, opened 2022-08-22
Outlier Detector Revamp

mauicv:
### What is this:

This is an experimental branch for reviewing potential implementations of the outlier detector revamp work. This should not be merged into master.

See following [notion docs](https://www.notion.so/Outlier-Revamp-Experimental-Work-a16269c0b42c4dfe9d52613bf8661dba) for more details.


alibi_detect/od/base.py
```diff
@@ -0,0 +1,71 @@
...
+        if getattr(self, 'aggregator') and self.aggregator.fitted:
+            scores = self.aggregator.transform(scores)
+            output['aggregate_scores'] = scores
+
+        if self.threshold_inferred:
+            p_vals = (1 + (scores[:, None] < self.val_scores).sum(-1))/len(self.val_scores)
+            preds = scores > self.threshold
+            output.update(scores=scores, preds=preds, p_vals=p_vals)
+
+        return output
```
arnaudvl, line 71: What is the reasoning to keep `raw_scores`, `normalised_scores` as well as `aggregate_scores`?

alibi_detect/od/knn.py
```diff
@@ -0,0 +1,48 @@
...
+            construct_name=self.__class__.__name__
+        ).verify_backend(backend)
+
+        self.k = k
+        self.kernel = kernel
+        self.ensemble = isinstance(self.k, np.ndarray)
+        self.normaliser = normaliser
+        self.aggregator = aggregator
+        self.fitted = False
+        self.backend = backends[backend]
```
arnaudvl, line 39: It looks like `self.backend` here is the equivalent of `self._detector` for drift detectors. Any specific reason for the slightly different approach here?

alibi_detect/od/knn.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+    def fit(self, X: np.ndarray) -> None:
+        self.x_ref = self.backend.fit(X)
+        val_scores = self.score(X)
+        if getattr(self, 'normaliser'): self.normaliser.fit(val_scores)
+
+    def score(self, X: np.ndarray) -> np.ndarray:
+        return self.backend.score(X, self.x_ref, self.k, kernel=self.kernel)
```
arnaudvl, line 48: What is the reason to pass all the attributes here explicitly as input args?

alibi_detect/od/backends/torch/knn.py
```diff
@@ -0,0 +1,17 @@
+import numpy as np
+import torch
+
+
+class KNNTorch:
+    def score(X, x_ref, k, kernel=None):
```
arnaudvl, line 6: Personally prefer all lowercase as for the drift detectors. Also see comment on why the input args would be necessary since they are all already attributes of the method.

alibi_detect/od/backends/torch/knn.py
```diff
@@ -0,0 +1,17 @@
...
+        ensemble = isinstance(k, (np.ndarray, list, tuple))
```
arnaudvl, line 7: Would like to see `list` as option for `k` but currently the top level method only allows types `Union[int, np.ndarray]`.

alibi_detect/od/knn.py
```diff
@@ -0,0 +1,48 @@
...
+from alibi_detect.od.backends import KNNTorch, KNNKeops
+from alibi_detect.utils.frameworks import BackendValidator
+
+backends = {
+    'pytorch': KNNTorch,
+    'keops': KNNKeops
+}
+
+
+class KNN(OutlierDetector):
```
arnaudvl, line 16: Requires device management, possibly also for the kernel.

alibi_detect/od/backends/keops/knn.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+import torch
+from pykeops.torch import LazyTensor
+
+
+def cdist(X, Y):
```
arnaudvl, line 6: Similar remarks about lowercasing throughout the script as in the `torch` version.

alibi_detect/od/backends/torch/knn.py
arnaudvl: Also already have a `self.ensemble` attribute so no need to recompute.

alibi_detect/od/backends/keops/knn.py
```diff
@@ -0,0 +1,25 @@
...
+class KNNKeops:
+    def score(X, x_ref, k, kernel=None):
+        ensemble = isinstance(k, (np.ndarray, list, tuple))
+        X = torch.as_tensor(X)
+        X_keops = LazyTensor(X[:, None, :])
+        x_ref_keops = LazyTensor(x_ref[None, :, :])
+        K = -kernel(X_keops, x_ref_keops) if kernel else cdist(X_keops, x_ref_keops)
+        ks = np.array(k) if ensemble else np.array([k])
+        bot_k_inds = K.argKmin(np.max(ks), dim=1)
+        all_knn_dists = (X[:, None, :] - x_ref[bot_k_inds][:, ks-1, :]).norm(dim=2)
```
arnaudvl, line 19: This defeats the purpose of using `keops`. You can leverage `keops`'s argKmin functionality. See [here](https://github.com/getkeops/keops/blob/main/pykeops/pykeops/tutorials/knn/plot_knn_torch.py) for more detail. Clear benchmarking is required in a simple example script.

arnaudvl, line 19: Doesn't this defeat the `keops` speed up because you are basically doing the same as the distance computation here? You might be able to use `Kmin_argKmin` instead of `argKmin` in the line above and compute both the indices and distances within `keops` at once. Once you have those you can index into it for the ensemble.

mauicv: Ah, I see what you mean. I think I'd assumed that `len(bot_k_inds)` is much less than `len(x_ref)` and `len(X)` and hence `(X[:, None, :] - x_ref[bot_k_inds][:, ks-1, :]).norm(dim=2)` wouldn't need to use `keops`?

arnaudvl: I think (not 100% sure due to multiple possible k's) you can do this within `keops` using `Kmin_argKmin` which would be drastically faster (expanding dims and doing broadcasted operations is slow and memory intensive). If `Kmin_argKmin` does not suffice for multiple k's, then you could implement this as a custom formula using a generic reduction. Check [this page](https://www.kernel-operations.io/keops/python/api/pytorch/Genred_torch.html) for more info.

alibi_detect/od/knn.py
mauicv: Essentially it is but `backend` here is just the score functionality and doesn't inherit from `BaseDetector`. I've added more notes here on [my thinking](https://www.notion.so/Hierarchy-055ebe62262c4ee78265bbae820addc7).

alibi_detect/od/processor/base.py
```diff
@@ -0,0 +1,13 @@
+import numpy as np
+
+
+class BaseProcessor():
+    def __init__(self, detectors):
+        self.detectors = detectors
+
+    def __call__(self, X):
+        B, _ = X.shape
```
arnaudvl, line 9: This implies that the input `x` has shape `(n, d)`, but would fail for e.g. `x = (n, d1, d2)`.

alibi_detect/od/ensemble.py
```diff
@@ -0,0 +1,26 @@
...
+from alibi_detect.od.base import OutlierDetector
+from alibi_detect.od.aggregation import BaseTransform, AverageAggregator, PValNormaliser
+from alibi_detect.od.processor import BaseProcessor
+
+
+class Ensemble(OutlierDetector):
+    def __init__(
+            self,
+            detectors: List[OutlierDetector], 
+            aggregator: Optional[BaseTransform] = AverageAggregator(),
```
arnaudvl, line 12: A good default aggregator is likely quite important for the ensemble. Simply averaging has its drawbacks.

alibi_detect/od/processor/base.py
```diff
@@ -0,0 +1,13 @@
...
+        results = np.empty((B, len(self.detectors)))
+        for ind, detector in enumerate(self.detectors):
+            results[:, ind] = detector.score(X)
```
arnaudvl, line 12: Does this break when you have an ensemble of ensembles (e.g. an ensemble incl. a KNN detector with `k=[1,2,3,4,5]` and some other detector)?

alibi_detect/od/base.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        Scores the instances and then compares to pre-inferred threshold.
+        For ensemble models the scores from each constituent is added to the output.
+        p-values are also returned by comparison to validation scores (of inliers)
+        """
+        output = {}
+        scores = self.score(X)
+        output['raw_scores'] = scores
+
+        if getattr(self, 'normaliser') and self.normaliser.fitted:
```
arnaudvl, line 58: Is updating the scores possibly multiple times within the predict but outside of the actual scoring function the best way to do it? Because if I just call `detector.score(x)` then I possibly get different scores which determine my p-value.

alibi_detect/od/processor/parallel.py
```diff
@@ -0,0 +1,26 @@
+import ray
+import numpy as np
+
+from alibi_detect.od.processor.base import BaseProcessor
+
+
+class ParallelProcessor(BaseProcessor):
```
arnaudvl, line 7: @ascillitoe can be of help here.

alibi_detect/od/ensemble.py
```diff
@@ -0,0 +1,26 @@
...
+            processor=BaseProcessor):
+
+        self.detectors = detectors
+        self.normaliser = normaliser
+        self.aggregator = aggregator
+        self.processor = processor(detectors=detectors)
+
+    def fit(self, X):
+        for detector in self.detectors:
+            detector.fit(X)
```
jklaise, line 23: Even if this is done in parallel, `fit` times vary wildly from basically zero (non-parametric e.g. `knn`) and lengthy model fitting phases. Wonder if worth investigating ensembling fitted detectors so the fitting gets outsourced to the user in whatever way they want to do it?

ascillitoe: Minor tangential thought; if we are not going to have an `Ensemble.fit` method, it might still be worth looping through all detectors and checking for a `fitted` attribute or something? Rather than leaving it until `score` is called to find out...

alibi_detect/od/processor/parallel.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, detectors):
+        self.detectors = detectors
+    
+    def __call__(self, X):
+        """This might not be the best way to do this. Need to checK:
+        1. Does ray serlize each detector as we pass it to pfunc.
+        2. Is this an issue?
```
jklaise, line 14: Yea, iirc `ray` uses `pickle` to serialize arguments so it won't work for all detectors out--of-the-box (e.g. unpicklable `tensorflow` detectors).

ascillitoe:
@mauicv can we close this now?

PR #602, opened 2022-08-26
Learned kernel MMD with KeOps backend

arnaudvl:
Following #548 , also extending the learned kernel MMD detector with the KeOps backend to further scale and speed up the detector.

- [x] Add tests.
- [x] Check if additional test for `LearnedKernelDrift` is required given decreased coverage.
- [x] Fix typing.
- [x] Add docs, clarify that to optimally benefit from the speed ups, an initial bandwidth for `kernel_a` and optionally `kernel_b` needs to be set.
- [x] Extend MMD benchmarking example.
- [x] Agree on API of `utils.keops.kernels.DeepKernel`. The main issue here is that `DeepKernel.proj` is not used within the DeepKernel's forward pass, but explicitly called by the learned kernel drift detector. The reason is that we first need to do regular torch tensor computations using the projection, and then with those projected features compute the kernel matrix. KeOps is only used for the latter, not when computing the projection itself. So technically you could pass a separate projection (i.e. `DeepKernel.proj`) to the drift detector and apply a weighted sum kernel later on the original and projected data. But this would break the consistency of the API's/input kwargs with the other backends and realistically make the detector harder to understand. As a result, I chose to keep the same DeepKernel format as the PyTorch and TensorFlow backends and deal with this difference directly in the drift detector. This also means there are explicit checks in place (`self.has_proj` and `self.has_kernel_b`) to check if the DeepKernel format is used and we can do the projection separately.
- [x] Would like to add `num_workers` for both KeOps and PyTorch backends since it can make a significant difference for the dataloader speed for higher number of instances. #611 
- [x] There is an additional `batch_size_predict` kwarg which I would also like to add to the PyTorch backend. The reason is that the optimal batch size for training can be wildly different than that for prediction (where we just care about being as fast as possible within our compute budget). So if we e.g. pick `batch_size=32` for training we might want to change this to `batch_size_predict=1000` for prediction using the trained kernel. There is also another reason why they can be very different: during training of the detector the whole training batch (all tensors incl. the projection etc) needs to fit on the GPU. But for predictions across all permutations at once we can first compute all the projections separately, and then lazily evaluate both the projections and original instances for all permutations. This means we can likely get away with much higher batch sizes for the projection predictions. #612 
- [x] Follow up PR separating out the PyTorch and KeOps trainers. #613 

The smaller potential PyTorch changes (`num_workers` and `batch_size_predict`) can be done in a quick follow up PR.


alibi_detect/utils/keops/kernels.py
```diff
@@ -1,10 +1,10 @@
 from pykeops.torch import LazyTensor
 import torch
 import torch.nn as nn
-from typing import Callable, Optional
+from typing import Callable, Optional, Union
 
 
-def sigma_mean(x: LazyTensor, y: LazyTensor, dist: LazyTensor, n_min: int = None) -> torch.Tensor:
+def sigma_mean(x: LazyTensor, y: LazyTensor, dist: LazyTensor, n_min: int = 100) -> torch.Tensor:
```
jklaise, line 7: What's the reason for changing the default here?

arnaudvl: Inferring the bandwidth requires `[n_min, n_min]` matrix to be evaluated just to check whether the passed `x == y`. Initially it defaulted to `nx`. This wasn't as big a deal for the normal MMD as by default you do this once at init, so it doesn't impact the prediction time. However for the learned kernel MMD you would have to do this every single prediction call as you learn new kernels, and this becomes a bottleneck, even though it is just a simple check in case you haven't provided an initial bandwidth (which will be updated during learning anyway afterwards). As a result, I reduced it to a default value of `min(nx, n_min)` with `n_min=100` as default setting. I believe this is a reasonable check to see whether the instances in x and y are the same (so you basically check the first `min(nx, n_min)` instances to see if they are the same).

arnaudvl: Just thinking that as an alternative I could infer the bandwidth by default in the same way as for MMD, and provide that bandwidth as initial one for every new kernel. This might actually be better as it aligns closer with normal MMD.

arnaudvl: Actually ignore the above comment. The deep kernel is defined outside of the detector so we can't assume access to heuristic setting.

alibi_detect/utils/keops/kernels.py
```diff
@@ -114,3 +114,65 @@ def forward(self, x: LazyTensor, y: LazyTensor, infer_sigma: bool = False) -> La
...
+
+class DeepKernel(nn.Module):
+    def __init__(
+        self,
+        proj: nn.Module,
+        kernel_a: nn.Module = GaussianRBF(trainable=True),
+        kernel_b: Optional[nn.Module] = GaussianRBF(trainable=True),
+        eps: Union[float, str] = 'trainable'
+    ) -> None:
+        """
```
jklaise, line 127: One for a follow-up - we need consistency for where we put docstrings, e.g. the `torch` one has the docstring in the class body not in the constructor which makes it hard to do side-by-side comparison. Please open an issue and delegate someone.

alibi_detect/cd/keops/learned_kernel.py
```diff
@@ -0,0 +1,341 @@
...
+            n_permutations: int = 100,
+            batch_size_permutations: int = 1000000,
+            var_reg: float = 1e-5,
+            reg_loss_fn: Callable = (lambda kernel: 0),
+            train_size: Optional[float] = .75,
+            retrain_from_scratch: bool = True,
+            optimizer: torch.optim.Optimizer = torch.optim.Adam,  # type: ignore
+            learning_rate: float = 1e-3,
+            batch_size: int = 32,
+            batch_size_predict: int = 1000000,
```
jklaise, line 34: Tip: for numbers with at least 4 digits use underscores to visually be able to tell between e.g. a million and hundred thousand, i.e. `int = 1_000_000`.

alibi_detect/cd/keops/learned_kernel.py
```diff
@@ -0,0 +1,341 @@
...
+            k_yy = self.kernel(y_proj, y2_proj, y, y2)
+            h_mat = k_xx + k_yy - k_xy - k_xy.t()
+
+            h_i = h_mat.sum(1).squeeze(-1)
+            h = h_i.sum()
+            mmd2_est = (h - n) / (n * (n - 1))
+            var_est = 4 * h_i.square().sum() / (n ** 3) - 4 * h.square() / (n ** 4)
+            reg_var_est = var_est + self.var_reg
+
+            return mmd2_est/reg_var_est.sqrt()
```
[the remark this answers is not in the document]
arnaudvl: There are good reasons for these small differences:
1. We can't straightforwardly take a trace of a symbolic matrix (can't easily/efficiently access entries of the matrix besides matrix reduction operations), hence the difference vs. `h_mat.trace()` in the PyTorch implementation.
2. In this implementation we're only taking the sum of the symbolic tensor `h_mat` 1x vs. 3x for the PyTorch implementation.

alibi_detect/cd/keops/learned_kernel.py
```diff
@@ -0,0 +1,341 @@
...
+                k_xx.append(self.kernel(x_proj, x2_proj, x, x2).sum(1).sum(1).squeeze(-1))
+                k_yy.append(self.kernel(y_proj, y2_proj, y, y2).sum(1).sum(1).squeeze(-1))
+
+        c_xx, c_yy, c_xy = 1 / (m * (m - 1)), 1 / (n * (n - 1)), 2. / (m * n)
+        # Note that the MMD^2 estimates assume that the diagonal of the kernel matrix consists of 1's
+        stats = c_xx * (torch.cat(k_xx) - m) + c_yy * (torch.cat(k_yy) - n) - c_xy * torch.cat(k_xy)
+        return stats[0], stats[1:]
+
+    @staticmethod
+    def trainer(
```
jklaise, line 308: Noting that this method is identical to the `torch` one. Since it's a static method anyway, perhaps it makes sense to factor out as a utility function in some `torch`-based module and use for both `torch` and `keops`?

arnaudvl: Makes sense. Preferred as part of this PR or a quick follow up one (there are a few things as listed above which impact the `torch` version as well and therefore I wanted to do separate in small PRs).

jklaise: Happy to have everything impacting the `torch` backend as a follow-up.

alibi_detect/cd/keops/learned_kernel.py
```diff
@@ -0,0 +1,341 @@
...
+        def __init__(self, kernel: nn.Module, var_reg: float, has_proj: bool, has_kernel_b: bool):
+            super().__init__()
+            self.kernel = kernel
+            self.has_proj = has_proj
+            self.has_kernel_b = has_kernel_b
+            self.var_reg = var_reg
+
+        def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+            n = len(x)
+            if self.has_proj:
```
ascillitoe, line 176: It might be worth adding a comment on why we have to have this `type: ignore`? Since we can probably remove it if we make the kernel definitions more strict later on...

alibi_detect/utils/keops/tests/test_kernels_keops.py
```diff
@@ -64,3 +63,59 @@ def test_gaussian_kernel(gaussian_kernel_params):
...
+
+
+@pytest.fixture
+def deep_kernel_params(request):
+    return tests_dk[request.param]
+
+
+@pytest.mark.skipif(not has_keops, reason='Skipping since pykeops is not installed.')
+@pytest.mark.parametrize('deep_kernel_params', list(range(n_tests_dk)), indirect=True)
+def test_deep_kernel(deep_kernel_params):
```
jklaise, line 93: Not a huge fan of using this pattern to parametrizing tests, is there a reason not to parametrize each parameter directly? @ascillitoe

alibi_detect/utils/keops/tests/test_kernels_keops.py
```diff
@@ -64,3 +63,59 @@ def test_gaussian_kernel(gaussian_kernel_params):
...
         assert (k_xx_min >= 0.).all() and (k_xy_min >= 0.).all()
+
+
+if has_keops:
+    class MyKernel(nn.Module):
+        def __init__(self):
+            super().__init__()
+
+        def forward(self, x: LazyTensor, y: LazyTensor) -> LazyTensor:
+            return (- ((x - y) ** 2).sum(-1)).exp()
```
jklaise, line 74: Is typing the only reason why this needs to be inside `if has_keops` block? If so we could just use forward-references to `'LazyTensor'`? @ascillitoe

alibi_detect/cd/keops/tests/test_learned_kernel_keops.py
```diff
@@ -0,0 +1,130 @@
...
+from typing import Callable, Optional, Union
+from alibi_detect.utils.frameworks import has_keops
+from alibi_detect.utils.pytorch import GaussianRBF as GaussianRBFTorch
+from alibi_detect.utils.pytorch import mmd2_from_kernel_matrix
+if has_keops:
+    from alibi_detect.cd.keops.learned_kernel import LearnedKernelDriftKeops
+    from alibi_detect.utils.keops import GaussianRBF
+    from pykeops.torch import LazyTensor
+
+n = 50
```
jklaise, line 15: Should add at least a comment (or a more descriptive variable name) to see at a glance what this is used for.

jklaise (changes requested):
Overall LGTM. Ok with proposed departure for the `DeepKernel` API for `keops` assuming `DeepKernel` is never really meant to be used by user directly?

arnaudvl:
> Overall LGTM. Ok with proposed departure for the `DeepKernel` API for `keops` assuming `DeepKernel` is never really meant to be used by user directly?

No, its intended usage is always within a learned detector. I will wait then until @ascillitoe leaves comments before making possibly some final changes and merging.

alibi_detect/cd/learned_kernel.py
```diff
@@ -27,13 +30,15 @@ def __init__(
             update_x_ref: Optional[Dict[str, int]] = None,
             preprocess_fn: Optional[Callable] = None,
             n_permutations: int = 100,
+            batch_size_permutations: int = 1000000,
```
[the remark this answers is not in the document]
ascillitoe: (or are you leaving for a follow-up PR?)

ascillitoe:
@arnaudvl the proposed api for the keops `DeepKernel` API looks sensible to me. Certainly preferable to having to make the actual detector more backend-specific. Plus, as you say, I can't see a good reason for someone to call `DeepKernel` directly.

alibi_detect/cd/learned_kernel.py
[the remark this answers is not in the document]
arnaudvl: As mentioned in the TODO's, would like to incorporate `num_workers` with the torch backend and top level detector in a quick follow up PR.

alibi_detect/utils/keops/tests/test_kernels_keops.py
ascillitoe: Yeh the more conventional way to do it would be to parametrize each parameter separately. e.g. see `test_save_model` in `test_saving.py`:

```python
@parametrize('model', [encoder_model])
@parametrize('layer', [None, -1])
def test_save_model(data, model, layer, backend, tmp_path):
```

This approach has the advantage of giving much more descriptive test names which is useful when things go wrong.

We keep writing tests with the current pattern for consistency with existing tests. But, unless we are going to refactor existing tests very soon maybe we should prioritise adopting a new pattern so that we have less refactoring to do later...

alibi_detect/utils/keops/tests/test_kernels_keops.py
ascillitoe: Good point. Looks like that might be the case...

alibi_detect/utils/keops/tests/test_kernels_keops.py
arnaudvl: Ok I was not aware here tbh and followed existing patterns.

alibi_detect/cd/tests/test_learned_kernel.py
```diff
@@ -37,7 +41,16 @@ def forward(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
         return torch.einsum('ji,ki->jk', self.dense(x), self.dense(y))
 
 
-tests_lkdrift = ['tensorflow', 'pytorch', 'PyToRcH', 'mxnet']
+if has_keops:
+    class MyKernelKeops(nn.Module):
```
ascillitoe, line 45: Same comment as [this one](https://github.com/SeldonIO/alibi-detect/pull/602/files/1f6d5e02e6c89bec080874cd9fb3b02a3f9fdf28#r964683557)?

alibi_detect/utils/keops/tests/test_kernels_keops.py
arnaudvl: #610

alibi_detect/cd/keops/learned_kernel.py
arnaudvl: Should be fixed now. `self.proj` is actually a `nn.Module` so the updated if statement simply checks for that.

alibi_detect/utils/keops/tests/test_kernels_keops.py
ascillitoe: Did the foward-ref not work?

arnaudvl: @jklaise and I were chatting earlier and decided (correct me if I'm wrong) that it's not necessarily a better solution, just a bit different, so I kept it as is.

alibi_detect/cd/keops/learned_kernel.py
ascillitoe: Assuming you've decided this is best left to do on a library-wide level later?

alibi_detect/utils/keops/tests/test_kernels_keops.py
ascillitoe: OK fair enough!

doc/source/cd/methods/learnedkerneldrift.ipynb
```diff
@@ -2,14 +2,22 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {},
+   "metadata": {
+    "pycharm": {
```
ascillitoe, line 6: Super nitpicky: It is preferable to strip out this unnecessary metadata...

ascillitoe (approved):
LGTM!

jklaise (approved):

PR #604, opened 2022-09-02
Serialisation of (online) state for online detectors

ascillitoe:
This PR implements the functionality to save and load state for online detectors. At a given time step, the `save_state` method can be called to create a "checkpoint". This can later be loaded via the `load_state` method. At any time, the ~`reset_state`~ `reset` method can be used to reset the state back to the `t=0` timestep. 

## Scope

This PR deals with _online state_ only. See https://github.com/SeldonIO/alibi-detect/pull/604#issuecomment-1352896731 for a discussion on _online_ versus _offline_ state.

### Saving and loading state

The state of online detectors can now be saved and loaded via `save_state` and `load_state`:

The detector's state may be saved with the `save_state` method:

```python
cd = CVMDriftOnline(x_ref, ert, window_sizes)  # Instantiate detector at t=0
cd.predict(x_1)  # t=1
cd.save_state('checkpoint_t1')  # Save state at t=1
cd.predict(x_2)  # t=2
```

The previously saved state may then be loaded via the `load_state` method:

```python
# Load state at t=1
cd.load_state('checkpoint_t1')
```

At any point, the state may be reset with the `reset` method. Also see [colab notebook](https://gist.github.com/ascillitoe/ddc65fb65c1a479554c9249e70897689).

### Saving and loading detector with state

Calling `save_detector` with `save_state=True` will save an online detectors state to `state/` within the detector save directory. `load_detector` will simply attempt to load state if a `state/` directory exists. 

```python
from alibi_detect.cd import LSDDDriftOnline
from alibi_detect.saving import save_detector, load_detector

# Init detector (self.t = 0)
dd = LSDDDriftOnline(x_ref, ert, window_size)

# Perform predict call to update state (e.g. self.t = self.t + 1)
dd.predict(x)

# Save detector with its state included
save_detector(dd, filepath, save_state=True)

# Load stateful detector (i.e. self.t = 1)
dd_new = load_detector(filepath)
```

## TODO's:

- [x] Implement for remaining detectors and backends.
- [x] Integrate with the `save_detector` and `load_detector` functions, to allow state to be saved and loaded when the detector itself is serialized/unserialized.
- [x] Tests.
- [x] Docs. 
- [x] Changelog
- [x] Public facing example. - _examples added to [Saving and loading](https://seldon--604.org.readthedocs.build/projects/alibi-detect/en/604/overview/saving.html#saving-state) page and detector methods pages._
- [x] Test all detectors and backends in `test_saving.py` state test.

~## Outstanding considerations (specific to LSDD for now but maybe more widely applicible)~

~There _might be_ an open question to resolve regarding what we define "state" to be. This PR currently considers it to be only the attributes that are updated in `_update_state` (`self.t`, `self.test_window` and `self.k_xtc`). In other words, "state" is defined as any attribute that is dependent on time (updated when a new instance `x_t` is given via `score` or `predict`).~

~However, there is already a notion of "state" introduced when we `initialise` a detector (or reinitialise it via the `reset` method). Here, in addition to the attributes already mentioned, we set `self.ref_inds`, `self.c2s`, and `self.init_test_inds`. This leads to considerations:~

~1. Will there be confusion between the `reset` and `reset_state` methods, and do we need to change the docstrings or names? 
2. There is randomness involved in the initialisation of `LSDDDrift` (in `_configure_ref_subset`). It is likely that if the detector is instantiated later on, and `load_state` is used to restart from a checkpoint, predictions will still be different compared to those that were observed after `save_state` was called with the original detector. This would only be avoided if random seeds were set both times. With this in mind, do we want to change our definition of "state" to include `self.ref_inds`, `self.c2s`, and `self.init_test_inds`?~


arnaudvl:
1. Stateful here relates to whatever state (attributes) changes between prediction calls, not what state is set in the init of the detector, which would make all detectors stateful of course. To avoid possible confusion 1 suggestion would be to name the methods `reset_detector` and `reset_state`?
2. I believe that in general randomness should be handled outside of the detector/library, similar to e.g. PyTorch models. Isn't that randomness already eliminated though when just loading a saved detector? What am I missing here?

ascillitoe:
> * Stateful here relates to whatever state (attributes) changes between prediction calls, not what state is set in the init of the detector, which would make all detectors stateful of course. To avoid possible confusion 1 suggestion would be to name the methods `reset_detector` and `reset_state`?
> * I believe that in general randomness should be handled outside of the detector/library, similar to e.g. PyTorch models. Isn't that randomness already eliminated though when just loading a saved detector? What am I missing here?

Nice idea with the name change, and yes I agree, "state" does not refer to any attributes set in init, as that would be "config" (with our definitions).

My only concern is that users might expect a detector to give the same predictions as the original when loaded from a "checkpoint" via `save/load_state`. That is currently not the case (unless seeds are set manually when the original and re-loaded detectors are instantiated). `save/load_detector` do not help with this.

Maybe the answer is just to make it clear in the docstrings though... as in any case statistically the detectors behaviour should be the same after the checkpoint even if the exact predictions are not the same?

ascillitoe:
Will be conflicts until #618 is merged.

ascillitoe:
Regarding the codecov report, the `4.32%` decrease is not accurate. This is based on an old `master` commit where we were still counting tests in coverage.

ascillitoe:
@arnaudvl @ojcobb (and @jklaise/@mauicv) could do with your thoughts on this. 

In the latest implementation, I have removed the new `reset_state` method. State can still be reset with the existing `reset` method. **`reset` calls `_initialise` internally, which involves a random operation** (for the LSDD detector at least). Because of this, we have a lack of determinism even if random seeds are manually set (explanation below). 

**Question**: Shall we keep a `reset_state` method separate to `reset`, or is this just confusing for users? @arnaudvl [previously suggested](https://github.com/SeldonIO/alibi-detect/pull/604#issuecomment-1235658280) we could rename `reset` to `reset_detector` to avoid confusion... but it might be worth discussing what we think the use case of each method actually is anyway. If we don't care about determinism in the case of Example 1, we can remove `reset_state` I think...

As Example 2 shows, determinism in the case of saving/loading of a detector is not affected by this decision anyway...

## Difference between `reset` and `reset_state`

(all examples are for `LSDDDriftOnline`)

### `reset`

`reset` is an existing method, which calls `_initialise`:

```python
    def reset(self) -> None:
        "Resets the detector but does not reconfigure thresholds."
        self._initialise()
```

`_initialise` typically sets some attributes to zero, and calls `_configure_ref_subset` (this method contains random ops!):

```python
    def _initialise(self) -> None:
        self.t = 0  # corresponds to a test set of ref data
        self.test_stats = np.array([])  # type: ignore[var-annotated]
        self.drift_preds = np.array([])  # type: ignore[var-annotated]
        self._configure_ref_subset()
```

### `reset_state`

`reset_state` was/is a new method that specifically only resets the core "stateful" attributes (those updated by _`update_state`):

```python
    def reset_state(self):
        """
        Reset the detectors state.
        """
        self.t = 0
        self.test_window = self.x_ref_eff[self.init_test_inds]
        self.k_xtc = self.kernel(self.test_window, self.kernel_centers)
```

This requires `_initialise` to have been run (so that `self.test_window` has been set), however it doesn't re-run `_initialise`, therefore doesn't involve random ops. 

## Examples
### Example 1: Determinism when resetting

When resetting an instantiated detector and repeating predictions, test stats will be repeatable if `reset_state` used. Example from `test_lsdd_online_pt.py`:

```python
    # Run for 50 time steps
    test_stats_1 = []
    for t, x_t in enumerate(x):
        preds = dd.predict(x_t)
        test_stats_1.append(preds['data']['test_stat'])
        if t == 20:
            dd.save_state(tmp_path)

    # Clear state and repeat, check that same test_stats both times
    dd.reset_state()
    test_stats_2 = []
    for t, x_t in enumerate(x):
        preds = dd.predict(x_t)
        test_stats_2.append(preds['data']['test_stat'])
    np.testing.assert_array_equal(test_stats_1, test_stats_2)  # passes!
```    

This fails if `reset` is used, even if `torch.manual_seed()` is run before instantiating the detector and before `reset`. Setting seeds externally does not help here because the number of `torch.random` operations run prior to reaching `_initialise` is different in both cases (fresh instantiation also involves random operations in `_configure_kernel_centers` and `_configure_thresholds`).

### Example 2: Determinism when saving/loading

When saving and loading a stateful detector via `save_detector(..., save_state=True)` and `load_detector`, predictions following `save_detector` and `load_detector` will consistent as long as seeds are set manually. Example from `test_saving.py`:

```python
     with fixed_seed(seed):
         dd = detector(X_ref, ert=100, window_size=10, backend='pytorch')
    # Run for 50 time-steps
    test_stats = []
    for t, x_t in enumerate(X_h0[:50]):
        test_stats.append(dd.predict(x_t)['data']['test_stat'])
        if t == 20:
            # Save detector (with state)
            save_detector(dd, tmp_path, save_state=True)

    # Check state/ dir created
    state_path = dd.state_path if detector == CVMDriftOnline else dd._detector.state_path
    assert state_path == tmp_path.joinpath('state')
    assert state_path.is_dir()

    # Load
    with fixed_seed(seed):
        dd_new = load_detector(tmp_path)
    # Check attributes and compare predictions at t=21
    assert dd_new.t == 21
    np.testing.assert_array_equal(dd_new.predict(X_h0[21])['data']['test_stat'], test_stats[21])
```

This use case does not depend on design of `reset`/`reset_state` etc.

ascillitoe:
Another consideration: if we were to include `reset_state`, we should also decide whether we wish to reset the results attributes `self.test_stats` and `self.drift_preds` in this (like `reset` does). Or is there any argument to reset a detectors state but retain the results attributes? I can't think of an argument but throwing it out there anyway...

ascillitoe:
Additional side-note, this issue with setting random seeds not giving deterministic behaviour for a given operation (in this case the `torch.randperm` called by `reset` -> `_initialise` -> `_configure_ref_subset`) is something we've run into a few times before. The issue is that even if a given random state is set externally (i.e. `torch.manual_seed`), the random state will be different by the time we get to the `torch.randperm` if a different number of random operations are called before we get there, since each random op cycles the random state. This is made even more difficult for something like `LSDDDriftOnline`, since we have random ops in a `while` loop, so cannot do not know how many random ops will be called.

The only solution I can think of for this is a scikit-learn style approach, where we accept `random_state` as a kwarg, and then use `self.random_state` in random ops we want to be deterministic. We would have to be careful with ops like the `torch.randperm` in `_configure_ref_subset`, since we do need this to possess a degree of randomness over each iteration in the `while` loop...

ascillitoe:
@arnaudvl @ojcobb a possible alternative strategy to make `reset` deterministic is to avoid reworking `configure_ref_subset` so that we don't do the `while` loop to find `self.init_test_inds` if it already exists. e.g. for `LSDDDriftOnline`:

```python
    def _configure_ref_subset(self):
        etw_size = 2 * self.window_size - 1  # etw = extended test window
        nkc_size = self.n - self.n_kernel_centers  # nkc = non-kernel-centers
        rw_size = nkc_size - etw_size  # rw = ref-window
        # Check if already configured, re-initialise stateful attributes w/o reconfiguring if so
        if self.init_test_inds is not None:
            self.test_window = self.x_ref_eff[self.init_test_inds]
            self.k_xtc = self.kernel(self.test_window, self.kernel_centers)
            return
        # Make split and ensure it doesn't cause an initial detection
        lsdd_init = None
        while lsdd_init is None or lsdd_init >= self.get_threshold(0):
            # Make split
            perm = torch.randperm(nkc_size)
            self.ref_inds, self.init_test_inds = perm[:rw_size], perm[-self.window_size:]
            self.test_window = self.x_ref_eff[self.init_test_inds]
            # Compute initial lsdd to check for initial detection
            self.c2s = self.k_xc[self.ref_inds].mean(0)  # (below Eqn 21)
            self.k_xtc = self.kernel(self.test_window, self.kernel_centers)
            h_init = self.c2s - self.k_xtc.mean(0)  # (Eqn 21)
            lsdd_init = h_init[None, :] @ self.H_lam_inv @ h_init[:, None]  # (Eqn 11)
```

Would be interested if I have missed any obvious reasons not to do this (as I probably have...)

ascillitoe:
## Offline vs online state

Based on offline, our current view is that we have a distinction between offline state (stateful attributes such as `self.thresholds` computed at instantiation) and online state (stateful attributes updated by `_update_state` when `score`/`predict` are called). In the future we wish to save/load offline state within `save_detector`/`load_detector`, to avoid repeating expensive instantiation procedures such as configuring thresholds. **This PR only deals with online state**.

The previous issue was that, for the multivariate detectors (LSDD and MMD), online state was initialised in a non-deterministic way within `configure_ref_subset`, which was called in `_initialise`, itself called in `reset`. This meant that, unlike univariate detectors, the `reset` method of multivariate detectors was not deterministic. Also, the `reset` method didn't just reset online state, but also reset offline state such as `self.test_window` etc.

The latest commits (09c4305 onwards) move the setting of online state into `_initialise_state`, and `configure_ref_subset` is taken out of `reset`. There is now a clear distinction; `load_state`, `save_state`, `_initialise_state` and `reset` deal with online state. Resetting offline state can be achieved by re-instantiating a fresh detector with a different random seed. 

@ojcobb could I get your thoughts on these changes, please?

### Another question...

Thinking ahead to when we handle offline and online state, do we need two reset methods? The latest commits change `reset` so that it resets _online state_ only. Do we instead want to reserve `reset` for resetting _offline state_, and rename the current `reset` to `reset_state` (since the other methods for handling _online state_ are all named `_state`)? I am wondering if this is actually neccesary since offline state can be reset by simply reinstantiating the detector (caveat that this depends on our definition of _offline state__ i.e. all attributes, or just results of `configure_thresholds` etc).

ojcobb:
I will take a closer look at the code next week but the overall approach you've described sounds sensible to me. I agree that there's no need for a method to reset offline state when this would just correspond to reinitialising the detector.

ascillitoe:
> I will take a closer look at the code next week but the overall approach you've described sounds sensible to me. I agree that there's no need for a method to reset offline state when this would just correspond to reinitialising the detector.

Cheers! and yes no rush.

alibi_detect/saving/loading.py
```diff
@@ -140,6 +139,14 @@ def _load_detector_config(filepath: Union[str, os.PathLike]) -> ConfigurableDete
     logger.info('Instantiating detector.')
     detector = _init_detector(cfg)
     logger.info('Finished loading detector.')
+
+    # Load state if it exists (and detector supports it)
+    # TODO - this will be removed in follow-up offline state PR, as loading to be moved to __init__ (w/ state_dir kwarg)
+    if isinstance(detector, StatefulDetector):
```
ojcobb, line 145: Nitpick but perhaps the above logging info could reflect whether it has finished loading a stateless or stateful detector, perhaps mentioning the timestep in the latter case?

ascillitoe: Nice idea!

ascillitoe: Done in 9e844105587d850ce9e3325073b60b3c5dcf95d5

ojcobb:
Beyond what we've discussed around the preemptive renaming of `reset` to `reset_state` it all looks good to me! Good work on the additional docstrings too!

ascillitoe:
Offline, @ojcobb asked why we don't have a separate `reset` and `reset_state` method, which is something mentioned in the notion doc. 

This is because the second reset method will be added in the follow-up PR for handling "offline state". Since the other online state methods are all named `..._reset`, the reset method for "online state" (`reset` in this PR) is named `reset_state` in the follow-up PR, and a new `reset` method is added to reset the "offline state". We need to decide if we want to make this change now or leave the follow-up. Making the change now would mean renaming `reset` to `reset_state` (in examples too), and perhaps fowarding `reset` to `reset_state` perhaps with a deprecation warning that its purpose will soon change...

Alternatively, we could leave `reset` as is, and add a new `reset_all` (or `reinitialize`?) or something similar for the offline state.  

@jklaise @mauicv

ascillitoe:
> Beyond what we've discussed around the preemptive renaming of `reset` to `reset_state` it all looks good to me! Good work on the additional docstrings too!

Thanks @ojcobb !

alibi_detect/base.py
```diff
@@ -229,6 +225,17 @@ def from_config(self): ...
     def _set_config(self): ...
 
 
+@runtime_checkable
+class StatefulDetector(Detector, Protocol):
```
mauicv, line 229: 😎 Proto-cool!

mauicv: We should probably rename these `DetectorProtocol`, `ConfigurableDetectorProtocol` and `StatefulDetectorProtocol`. Otherwise we risk people thinking `Detector` is the base detector object and for some reason these protocols inherit from it!

alibi_detect/saving/saving.py
```diff
@@ -110,7 +117,9 @@ def _cleanup_filepath(orig_files: set, filepath: Path):
 
 
 # TODO - eventually this will become save_detector (once outlier and adversarial updated to save via config.toml)
-def _save_detector_config(detector: ConfigurableDetector, filepath: Union[str, os.PathLike]):
+def _save_detector_config(detector: ConfigurableDetector,
```
mauicv, line 120: do we not want to include StatefulDetector here in type hints?

alibi_detect/base.py
mauicv: Also should `StatefulDetector` inherit from `ConfigureableDetector`? Are all stateful detectors configurable? If not are they intended to be at some point?

alibi_detect/cd/base_online.py
```diff
@@ -106,6 +112,66 @@ def _configure_ref_subset(self):
...
+        Parameters
+        ----------
+        filepath
+            The directory to load state from.
+        """
+        self._set_state_dir(filepath)
+        self._load_state()
+        logger.info('State loaded for t={} from {}'.format(self.t, self.state_dir))
+
+    def _save_state(self):
```
mauicv, line 153: Why do we have both public and private `load_state` and `save_state` methods?

alibi_detect/cd/base_online.py
```diff
@@ -290,7 +377,92 @@ def _configure_ref(self):
     def _update_state(self, x_t: np.ndarray):
         pass
 
+    def _set_state_path(self, dirpath: Union[str, os.PathLike]):
```
mauicv, line 380: What's the difference between `BaseUniDriftOnline` and `BaseMultiDriftOnline` that means one has `_set_state_path` and the other doesn't?

mauicv:
Perhaps a silly question about the general functionality? Is the intention that the user is saving state to the detector save directory or elsewhere? 

```py
from alibi_detect.cd import LSDDDriftOnline
from alibi_detect.saving import save_detector, load_detector

dd_new = load_detector(filepath)

dd.predict(x)
cd.save_state('check_pt') 
```

if the user then late runs:

```py
dd = load_detector(filepath)
```

That will not load the saved state right?

ascillitoe:
The 3.7 build failure is occurring before we even reach the state handling functionality. It appears to be a not-seen-before issue with the `MMDDriftOnlineTF` detector itself. See https://github.com/SeldonIO/alibi-detect/issues/706.

alibi_detect/cd/base_online.py
ascillitoe: Good question. This pattern was chosen to minimise duplication once we add offline state in the follow-up PR (POC: https://github.com/ascillitoe/alibi-detect/pull/44). `_save_state` then has additional logic to save either an online or offline state file:

```python
    def save_state(self, filepath: Union[str, os.PathLike]):
        """
        Save a detector's state to disk in order to generate a checkpoint.
        Parameters
        ----------
        filepath
            The directory to save state to.
        """
        self._set_state_dir(filepath)
        self._save_state()

    def _save_state(self, offline: bool = False):
        """
        Private method to save a detector's offline or online state to disk.
        "Offline state" is state created when a detector is instantiated. For example, thresholds and kernel centers.
        "Online state" is the state updated by `score` (and `predict`) methods.
        Parameters
        ----------
        offline
            Whether to save offline state. Otherwise, online state is saved.
        """
        suffix = '.pt' if self.backend == Framework.PYTORCH else '.npz'
        filename = 'offline_state' if offline else 'state'
        keys = self.offline_state_keys if offline else self.online_state_keys
        save_state_dict(self, keys, self.state_dir.joinpath(filename + suffix))
```

We could instead have `save_state` and `save_offline_state`, however the above avoids duplicating the suffix handling bit etc, plus we want to keep everything to do with offline state private.

Another option would be to keep everything in `save_state` for now and change to the above later if you think that's better...

alibi_detect/cd/base_online.py
ascillitoe: Thanks, that shouldn't exist. I renamed it `_set_state_dir` but somehow didn't delete the old one.

ascillitoe: Removed in [5dea156](https://github.com/SeldonIO/alibi-detect/pull/604/commits/5dea1566bff351cdeb5865f5c612d60817f5cd40).

ascillitoe:
> Perhaps a silly question about the general functionality? Is the intention that the user is saving state to the detector save directory or elsewhere?
> 
> ```python
> from alibi_detect.cd import LSDDDriftOnline
> from alibi_detect.saving import save_detector, load_detector
> 
> dd_new = load_detector(filepath)
> 
> dd.predict(x)
> cd.save_state('check_pt') 
> ```
> 
> if the user then late runs:
> 
> ```python
> dd = load_detector(filepath)
> ```
> 
> That will not load the saved state right?

Not silly at all. I've gone back and forth over this functionality a fair bit. The current behaviour supports two distinct use cases:

1. **"Runtime checkpointing"**: Users can give whatever filepath they like to `save_state` and `load_state` to save/load checkpoints within their runtime. For example:
   ```python
   cd = CVMDriftOnline(x_ref, ert, window_sizes)  # Instantiate detector at t=0
   cd.predict(x_1)  # t=1
   cd.save_state('checkpoint_t1')  # Save state at t=1
   cd.predict(x_2)  # t=2
   cd.predict(x_3)  # t=3
   cd.predict(x_4)  # t=4
   cd.save_state('checkpoint_t4')  # Save state at t=4

   # Go back to t=1 for whatever reason...
   cd.load_state('checkpoint_t1')
   ```

2. **Serialising detectors**: Calling `save_detector(cd, filepath, save_state=True) will save the state in `filepath/state/`, along with the serialised detector itself. `load_detector(filepath) will then load state if a `state/` dir exists inside `filepath`.

alibi_detect/base.py
ascillitoe: Ha ha you have converted me into a  `Protocol` fan!

Good point about the renaming and inheritance. I'll do the inheritance once now, and have added an issue to rename the `Protocol`'s since renaming `Detector` and `ConfigurableDetector` isn't really related to online state.

alibi_detect/saving/saving.py
ascillitoe: Good idea!

alibi_detect/base.py
ascillitoe: Issue: https://github.com/SeldonIO/alibi-detect/issues/707

alibi_detect/base.py
```diff
@@ -229,6 +225,17 @@ def from_config(self): ...
     def _set_config(self): ...
 
 
+@runtime_checkable
+class StatefulDetector(ConfigurableDetector, Protocol):
```
mauicv, line 229: Are you planning to add an `OfflineStatefulDetector` in this [PR](https://github.com/ascillitoe/alibi-detect/pull/44/files) out of interest?

alibi_detect/utils/_state.py
```diff
@@ -0,0 +1,71 @@
...
+from pathlib import Path
+from alibi_detect.utils.pytorch import _save_state_dict as _save_state_dict_pt, \
+    _load_state_dict as _load_state_dict_pt
+
+from alibi_detect.base import BaseDetector
+
+logger = logging.getLogger(__name__)
+
+
+def save_state_dict(detector: BaseDetector, keys: tuple, filepath: Path):
```
mauicv, line 20: should this be `StatefulDetectorProtocol`?

alibi_detect/saving/loading.py
```diff
@@ -115,19 +117,16 @@ def _load_detector_config(filepath: Union[str, os.PathLike]) -> ConfigurableDete
...
     -------
     The instantiated detector.
     """
-    # Load toml if needed
-    if isinstance(filepath, (str, os.PathLike)):
-        config_file = Path(filepath)
-        config_dir = config_file.parent
-        cfg = read_config(config_file)
-    else:
-        raise ValueError("`filepath` should point to a directory containing a 'config.toml' file.")
```
mauicv, line 130: why are you removing this error? Do we check somewhere else?

alibi_detect/saving/saving.py
```diff
@@ -110,7 +117,9 @@ def _cleanup_filepath(orig_files: set, filepath: Path):
...
+                          filepath: Union[str, os.PathLike],
+                          save_state: bool = False):
```
mauicv, line 122: Should `save_state` default to False for stateful detectors? In my mind, I'd assume when saving it would save everything by default. But perhaps that's not the intended use case?

alibi_detect/base.py
ascillitoe: Good question. I didn't think too much about this. I think whether we need `OfflineStatefulDetector` or not in the future will depend on whether all detectors end up having "offline state" (in which case we won't need the extra protocol to distinguish those that don't) or not. Maybe to be safe `StatefulDetector` should be renamed `OnlineStatefulDetector` for now?

alibi_detect/base.py
```diff
@@ -229,6 +225,17 @@ def from_config(self): ...
...
+    """Type Protocol for detectors that have support for save/loading of state.
+
+    Used for typing save and load functionality in `alibi_detect.saving.saving`.
+    """
```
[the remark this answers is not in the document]
ascillitoe: Nice idea

alibi_detect/saving/loading.py
ascillitoe: I removed this since we already make these checks in `load_detector`. However, thinking again, since `_load_detector_config` will eventually become `load_detector` (when we remove legacy save/load), it makes more sense to leave this check in. I'll revert this now.

alibi_detect/saving/saving.py
mauicv: I notice also that the default behaviour for `load_detector` is to load the detector's state if it exists in the save directory. I guess if we have the above saving functionality perhaps we should have symmetric load functionality. So: `load_detector(dir, load_state=False)`.

mauicv: I think in general the best thing is to have the defaults be the typical user expectations. So save would default to saving everything in the same way load defaults to loading everything but have the extra functionality that allows them to ignore state while saving if they choose. I'm not sure thought, but perhaps there's a good reason to expect that save not include state...

mauicv (commented):
Some minor suggestions

jklaise:
> Offline, @ojcobb asked why we don't have a separate `reset` and `reset_state` method, which is something mentioned in the notion doc.
> 
> This is because the second reset method will be added in the follow-up PR for handling "offline state". Since the other online state methods are all named `..._reset`, the reset method for "online state" (`reset` in this PR) is named `reset_state` in the follow-up PR, and a new `reset` method is added to reset the "offline state". We need to decide if we want to make this change now or leave the follow-up. Making the change now would mean renaming `reset` to `reset_state` (in examples too), and perhaps fowarding `reset` to `reset_state` perhaps with a deprecation warning that its purpose will soon change...
> 
> Alternatively, we could leave `reset` as is, and add a new `reset_all` (or `reinitialize`?) or something similar for the offline state.
> 
> @jklaise @mauicv

Is there a need to have a method to reset "offline" state at all? Should we just tell people to re-initialize the detector? Or would this be a convenience to avoid expensive re-initialization in case the constructor runs a lengthy computation?

alibi_detect/base.py
mauicv: It probably makes sense to do so in the future PR so that it makes sense contextually.I'm happy

alibi_detect/utils/_state.py
ascillitoe: I struggled to make this change. `mypy` gives:

```
alibi_detect/cd/base_online.py:436: error: Argument 1 to "load_state_dict" has incompatible type "BaseUniDriftOnline"; expected "StatefulDetectorOnline"  [arg-type]
```
`save_state_dict` is called in `BaseUniDriftOnline.save_state`, with the `detector` arg being a `BaseUniDriftOnline` object. For some reason mypy doesn't recognise `BaseUniDriftOnline` as being a `StatefulDetectorOnline` despite it having `save_state` and `load_state` methods...

I'm tempted to leave this as `BaseDetector` since technically `save_state_dict` and `load_state_dict` will actually work for any detector i.e. they just save/load attributes to/from disk, regardless of whether the detector has `save_state` and `load_state` methods... Open to suggestions on how to fix the typing though.

ascillitoe:
> > Offline, @ojcobb asked why we don't have a separate `reset` and `reset_state` method, which is something mentioned in the notion doc.
> > This is because the second reset method will be added in the follow-up PR for handling "offline state". Since the other online state methods are all named `..._reset`, the reset method for "online state" (`reset` in this PR) is named `reset_state` in the follow-up PR, and a new `reset` method is added to reset the "offline state". We need to decide if we want to make this change now or leave the follow-up. Making the change now would mean renaming `reset` to `reset_state` (in examples too), and perhaps fowarding `reset` to `reset_state` perhaps with a deprecation warning that its purpose will soon change...
> > Alternatively, we could leave `reset` as is, and add a new `reset_all` (or `reinitialize`?) or something similar for the offline state.
> > @jklaise @mauicv
> 
> Is there a need to have a method to reset "offline" state at all? Should we just tell people to re-initialize the detector? 

Tagging @ojcobb for this discussion too. I'm not strongly opinionated either way. I can see a `reset` (offline state) method being useful in the case where a detector has been instantiated with lots of arguments. i.e. to save the user needing to pack all their args/kwargs into a dict so they can easily reinstantate the detector later.

> Or would this be a convenience to avoid expensive re-initialization in case the constructor runs a lengthy computation?

I don't think the `reset` (offline state) method would be for this. Our definition of "offline state" is essentially expensive-to-compute attributes that we want to save in order to avoid expensive computation at load time. In the 2nd PR these expensive operations are encapsulated in `_initialise_offline_state`. The offline `reset` method would pretty much just call `_initialise_offline_state` i.e. it would run the lengthy computations!

alibi_detect/utils/_state.py
mauicv: urgh, mypy!!!! Yeah lets just leave it for now!

jklaise:
@ascillitoe in that case I don't see a particular reason to introduce another "reset" method, especially since having 2 would very likely lead to confusion!

Regarding possible renaming of `reset` to `reset_state`, I don't have strong opinions and can see pros for both approaches.

alibi_detect/saving/saving.py
ascillitoe: > I notice also that the default behaviour for `load_detector` is to load the detector's state if it exists in the save directory. I guess if we have the above saving functionality perhaps we should have symmetric load functionality. So: `load_detector(dir, load_state=False)`.

I'm not so sure about complicating `load_detector` with addition options. It seems quite natural to me that `load_detector` will load whatever you saved in `filepath` with `save_detector`. In other words, deciding what to save is done only at save time. The user could always through away the state later by running `dd = load_detector()` and then `dd.reset()`.

ascillitoe: > I think in general the best thing is to have the defaults be the typical user expectations. So save would default to saving everything in the same way load defaults to loading everything but have the extra functionality that allows them to ignore state while saving if they choose. I'm not sure thought, but perhaps there's a good reason to expect that save not include state...

I am leaning toward agreeing with you wrt to by default saving online state. Offline state will need more thought in the future I think. Saving offline state might limit the degree to which "config" can be later edited (e.g. offline state would include saving `self.thresholds`, which would prevent the `window_size` being changed at a later date). In other words, IMO adopting a rule that we save/and load EVERYTHING might be a step too far. Referring back to my last comment, I think its more about saving "whatever makes sense" to `filepath`, and then `load_detector` loads EVERYTHING that is in `filepath`.

alibi_detect/utils/_state.py
ascillitoe: 🙈

alibi_detect/base.py
ascillitoe: I've renamed it to `OnlineStatefulDetector`. Hope that is OK! :)

ascillitoe:
> e.g. `reset` identifying an online detector that has some state - this means we can't repurpose `reset` for anything else in the future

This is a really important point I think. Since we are using `_state` to identify "online state" functionality, I'm now thinking it makes a lot of sense to rename `reset` to `reset_state`. The final think to think about with this would be how to deprecated `reset`. For example, point `reset` to `reset_state` for now and give a deprecation warning saying its functionality might change in future? Or simply just rename and list as a breaking change?

jklaise:
Maybe I'm missing something but I thought `reset` didn't exist at all until this PR? Agree that using `reset_state` likely makes most sense.

ascillitoe:
> Maybe I'm missing something but I thought `reset` didn't exist at all until this PR? Agree that using `reset_state` likely makes most sense.

It did exist:

```python
    def reset(self) -> None:
        "Resets the detector but does not reconfigure thresholds."
        self._initialise()
        """
        Resets the detector to its initial state (`t=0`). This does not include reconfiguring thresholds.
        """
        self._initialise_state()
```

But its behavior has now changed. Previously, `self._initialise` called in `reset` did more than just initialising online state, so `reset` essentially reset all "online state" as well as some "offline state" (essentially everything except thresholds). This meant there was some potentially undesirable behavior with `reset` not being deterministic (see https://github.com/SeldonIO/alibi-detect/pull/604#issuecomment-1352896731). With the new behavior, `reset` exclusively only resets "online state" (takes us back to `t=0`).

ascillitoe:
p.s. my latest follow-up changes to @mauicv's comments have broken the CI. Working on a fix now!

jklaise:
> > Maybe I'm missing something but I thought `reset` didn't exist at all until this PR? Agree that using `reset_state` likely makes most sense.
> 
> It did exist:
> 
> ```python
>     def reset(self) -> None:
>         "Resets the detector but does not reconfigure thresholds."
>         self._initialise()
>         """
>         Resets the detector to its initial state (`t=0`). This does not include reconfiguring thresholds.
>         """
>         self._initialise_state()
> ```
> 
> But its behavior has now changed. Previously, `self._initialise` called in `reset` did more than just initialising online state, so `reset` essentially reset all "online state" as well as some "offline state" (essentially everything except thresholds). This meant there was some potentially undesirable behavior with `reset` not being deterministic (see [#604 (comment)](https://github.com/SeldonIO/alibi-detect/pull/604#issuecomment-1352896731)). With the new behavior, `reset` exclusively only resets "online state" (takes us back to `t=0`).

Ah ok. Shall we introduce `reset_state` and deprecate the currently existing `reset` then?

alibi_detect/base.py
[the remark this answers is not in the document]
mauicv: ok, makes sense

ascillitoe:
> Ah ok. Shall we introduce reset_state and deprecate the currently existing reset then? 

I've done this in [aa6b0d3](https://github.com/SeldonIO/alibi-detect/pull/604/commits/aa6b0d382dced113678f2642982f38f52e00b7f8), but kept `reset` as a link to `reset_state` with a `DeprecationWarning`. Can remove if you don't think this is necessary...

jklaise:
@ascillitoe Is this file move ok? I thought `conftest.py` should always be inside a folder named `tests`? [alibi_detect/saving/tests/conftest.py → alibi_detect/conftest.py](https://github.com/SeldonIO/alibi-detect/pull/604/files#diff-e1cda4eb9074d4d0004843812944b7e3d382617320a0d7a6c2b4315363f91a9a)

alibi_detect/base.py
```diff
@@ -229,6 +225,17 @@ def from_config(self): ...
...
 
+@runtime_checkable
+class StatefulDetectorOnline(ConfigurableDetector, Protocol):
+    """Type Protocol for detectors that have support for save/loading of online state.
+
+    Used for typing save and load functionality in `alibi_detect.saving.saving`.
+    """
+    def save_state(self, filepath): ...
+
+    def load_state(self, filepath): ...
```
jklaise, line 236: Should parameters have type hints?

ascillitoe:
> @ascillitoe Is this file move ok? I thought `conftest.py` should always be inside a folder named `tests`? [alibi_detect/saving/tests/conftest.py → alibi_detect/conftest.py](https://github.com/SeldonIO/alibi-detect/pull/604/files#diff-e1cda4eb9074d4d0004843812944b7e3d382617320a0d7a6c2b4315363f91a9a) Doesn't this make `conftest` a public module of `alibi-detect` which we wouldn't want?

Mmn good point, thinking again it doesn't seem ideal to have it outside of `tests/`. I moved it so that we didn't have to duplicate the `seed` fixture in multiple places. I couldn't think of a better way to have a global `conftest` that is shared across all tests. Do you know of a way?

Re it becoming a public module I suspect you're mostly right. We do `__all__ = ["ad", "cd", "models", "od", "utils", "saving"]` in `alibi_detect.__init__` so it won't be exposed by `dir(alibi_detect)` at least. However, `seed` could technically be imported by:

```

```

However, with our current `alibi-detect` installed, the following also works! (think I need to undo the file move, but maybe we also need to double check this separately)

```
from alibi_detect.saving.tests.conftest import seed
seed(0)
```

alibi_detect/cd/base_online.py
```diff
@@ -106,6 +113,66 @@ def _configure_ref_subset(self):
...
+        """
+        Load the detector's state from disk, in order to restart from a checkpoint previously generated with
+        `save_state`.
+
+        Parameters
+        ----------
+        filepath
+            The directory to load state from.
+        """
+        self._set_state_dir(filepath)
```
jklaise, line 150: Why is this necessary when loading?

alibi_detect/cd/base_online.py
```diff
@@ -106,6 +113,66 @@ def _configure_ref_subset(self):
...
+    def _set_state_dir(self, dirpath: Union[str, os.PathLike]):
+        """
+        Set the directory path to store state in, and create an empty directory if it doesn't already exist.
+
+        Parameters
+        ----------
+        dirpath
+            The directory to save state file inside.
+        """
+        self.state_dir = Path(dirpath)
```
jklaise, line 125: Should this be a private attribute?

alibi_detect/cd/base_online.py
```diff
@@ -290,7 +387,79 @@ def _configure_ref(self):
...
+
+    def _load_state(self, offline: bool = False):
+        """
+        Private method to load a detector's state from disk.
+
+        TODO - Method slightly verbose as designed to facilitate loading of "offline" state in follow-up PR.
+        """
+        filename = 'state'
+        load_state_dict(self, self.state_dir.joinpath(filename + '.npz'), raise_error=True)
+
```
jklaise, line 447: Seems like a lot of duplicated code that's exactly the same as for `BaseMultiDriftOnline` which suggests we may want to refactor using functions instead of methods or a mixin class? Or perhaps the class hierarchy needs to be updated.

alibi_detect/saving/loading.py
```diff
@@ -139,7 +141,16 @@ def _load_detector_config(filepath: Union[str, os.PathLike]) -> ConfigurableDete
     # Init detector from config
     logger.info('Instantiating detector.')
     detector = _init_detector(cfg)
+
+    # Load state if it exists (and detector supports it)
+    # TODO - this will be removed in follow-up offline state PR, as loading to be moved to __init__ (w/ state_dir kwarg)
+    if isinstance(detector, StatefulDetectorOnline):
+        state_dir = config_dir.joinpath((STATE_PATH))
```
jklaise, line 148: Nitpick: my IDE says the extra parentheses are unnecessary.

alibi_detect/cd/base_online.py
ascillitoe: Just so that `self.state_dir` is set (and converted from `str` to `pathlib.Path) when `load_state` is called as well as when `save_state` is called. 

I thought it might be helpful to have `state_dir` as a public attribute so that a user could see interrogate the detector to see where state was loaded from. Although thinking about it more, for the backend detectors one would have to do `detector._detector.state_dir` (access a private attribute) anyway. I guess we'd probably want to define a `@property` if we actually want to support this functionality properly...

Happy to just make it private if you think its better though...

alibi_detect/cd/base_online.py
ascillitoe: See https://github.com/SeldonIO/alibi-detect/pull/604#discussion_r1069655715

alibi_detect/saving/saving.py
jklaise: This is pretty important to decide on (ofc we can change default behaviour later), what is the more common use case after developing an online detector and deciding to deploy it - not sure the answer is so clear? @ojcobb @arnaudvl 

Regardless of behaviour, this needs to be explicit in the docs (not just docstrings imo).

alibi_detect/cd/base_online.py
ascillitoe: Fair point. Having a  `BaseDriftOnline` class for generic methods, or a mix-in both sound much nicer than this. I'll have a rethink 👍🏻

alibi_detect/saving/tests/test_saving.py
```diff
@@ -808,6 +808,136 @@ def test_save_onlinefetdrift(data, tmp_path, seed):
...
+@parametrize("detector", [MMDDriftOnline, LSDDDriftOnline])
+@parametrize_with_cases("data", cases=ContinuousData, prefix='data_')
+def test_save_multivariate_online_state(detector, data, backend, seed, tmp_path):
+    """
+    Test the saving (and loading) of multivariate online detectors' state when `save_state=True` is passed to
+    `save_detector`.
+    """
+    # Skip if backend not `tensorflow` or `pytorch`
+    if backend not in ('tensorflow', 'pytorch'):
+        pytest.skip("Detector doesn't have this backend")
```
jklaise, line 820: Is this due to some `keops` behaviour? Basically asking why skip here.

alibi_detect/saving/tests/test_saving.py
```diff
@@ -808,6 +808,136 @@ def test_save_onlinefetdrift(data, tmp_path, seed):
...
+    for child in tmp_path.joinpath('state').glob('*'):
+        if child.is_file():
+            child.unlink()
+    with pytest.raises(FileNotFoundError):
+        load_detector(tmp_path)
+
+
+@parametrize("detector", [FETDriftOnline])
+@parametrize_with_cases("data", cases=BinData, prefix='data_')
+def test_save_fet_online_state(detector, data, seed, tmp_path):
```
jklaise, line 894: `seed` not used

alibi_detect/cd/tests/test_cvm_online.py
```diff
@@ -12,29 +12,22 @@
...
 n_features = [1, 3]
 
-tests_cvmdriftonline = list(product(window_sizes, batch_size, n_features))
-n_tests = len(tests_cvmdriftonline)
 
+@pytest.mark.parametrize('window_sizes', window_sizes)
+@pytest.mark.parametrize('batch_size', batch_size)
+@pytest.mark.parametrize('n_feat', n_features)
+def test_cvmdriftonline(window_sizes, batch_size, n_feat, seed):
+    with fixed_seed(seed):
```
jklaise, line 20: Noting that the previous version of tests didn't have a fixed seed, presumably it wasn't needed in this setting as test suite has been passing. Is there a need to introduce a fixed seed here as it seems detrimental to the testing for this particular set of tests?

jklaise: P.S. same comment applies to tests below.

jklaise (changes requested):
Overall LGTM, main question about default behaviour wrt state saving when `save_detector` is called on online detectors. Regardless of choice, I believe this should be prominent in saving and method docs.

alibi_detect/base.py
ascillitoe: Added in 15d1dfaf49ab2a34eb82d78b39449992ee210d14. Note: I also updated the pre-existing `get_config` and `set_config` methods here.

alibi_detect/saving/loading.py
ascillitoe: Corrected in da5671403b2ec500a577935db4e25618746974a7.

alibi_detect/saving/tests/test_saving.py
ascillitoe: The `test_saving` file cycles through all possible backends:

```python
backends = ['tensorflow', 'pytorch', 'sklearn']
if has_keops:  # pykeops only installed in Linux CI
    backends.append('keops')
backend = param_fixture("backend", backends)
```

We have to skip tests if the associated detector doesn't have that backend. In this case, online detectors do not have a `keops` backend.

alibi_detect/cd/tests/test_cvm_online.py
ascillitoe: Mmn yeh I figured it would be good to add since there is randomness in the initialization of these detectors (in `_configure_thresholds`, and in the generation of `x_ref`/`x_h0`/`x_h1`). Although the tests do currently pass without fixing the seed, this doesn't actually mean they pass for any random seed. I seem to recall that when I looked into this before, `np.random.seed`'s set in one test leaked into others. Presumably, this means we have been implicitly fixing the seed in these tests anyway. 

Ideally (IMO), we'd get to a point where any random operations in tests are done inside `with fixed_seed(seed)`'s, then if a new bug is introduced, we can go back and reproduce it with the same random seed.

alibi_detect/saving/tests/test_saving.py
```diff
@@ -808,6 +808,136 @@ def test_save_onlinefetdrift(data, tmp_path, seed):
...
+    for child in tmp_path.joinpath('state').glob('*'):
+        if child.is_file():
+            child.unlink()
+    with pytest.raises(FileNotFoundError):
+        load_detector(tmp_path)
+
+
+@parametrize("detector", [CVMDriftOnline])
+@parametrize_with_cases("data", cases=ContinuousData, prefix='data_')
+def test_save_cvm_online_state(detector, data, seed, tmp_path):
```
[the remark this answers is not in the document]
ascillitoe: Removed in 6240fdfc3161015e6ce90e85800761bbf7bfdfa6

alibi_detect/saving/tests/test_saving.py
ascillitoe: Removed in 6240fdfc3161015e6ce90e85800761bbf7bfdfa6

alibi_detect/cd/base_online.py
ascillitoe: To reduce duplication, 5daf1b1 adds a `BaseDriftOnline` class. @jklaise @mauicv could I get your thoughts on the design of `BaseDriftOnline` please? I've gone with a parent class rather than mix-in since it seems strange to define a mix-in in `alibi_detect/base.py` when it is only to be used in two classes (`BaseMultiDriftOnline` and `BaseUniDriftOnline`). I also decided to put it in `alibi_detect/cd/base_online.py` rather than `alibi_detect/base.py` since at the moment the concept of "online" detectors is specific to drift (this may change if we decide stateful outlier detectors are in fact "online").

alibi_detect/cd/tests/test_cvm_online.py
jklaise: I agree that in case a bug happens then it's valuable to be able to reproduce with the same seed. But, on the other hand, "any random operations in tests done inside `with fixed_seed(seed)`" sounds like the opposite to what we want to do (unless for tests where we compare outputs of the same seed) - as stuff should pass most tests with any seed?

alibi_detect/cd/base_online.py
```diff
@@ -8,24 +8,98 @@
...
     online_state_keys: Tuple[str, ...]
+    backend: Optional[Literal['pytorch', 'tensorflow']] = None
+
+    @abstractmethod
+    def _configure_thresholds(self):
+        pass
+
+    @abstractmethod
+    def _update_state(self, x_t):
+        pass
```
jklaise, line 33: Type hints of parameters and return types required.

alibi_detect/cd/base_online.py
jklaise: LGTM however noting that there's quite a few abstract methods, some of which (not all?) are implemented in the `Multi/Uni` abstract methods, which come with their own set of abstract methods... Worried that this may become a bit tricky to keep track of. As a minimum, would group all abstract methods to come after each other and add docstrings on expected implementation and also, where valid, which of the `Multi/Uni` classes implement these methods (+ type hints as always).

alibi_detect/cd/base_online.py
ascillitoe: This is unfortunately necessary, since univariate online detectors have `def _update_state(self, x_t: np.ndarray):`  whilst multivariate have `def _update_state(self, x_t: torch.Tensor):` (or `tf.Tensor`). We violate Liskov's substitution principle slightly.

I sort of think this is OK to not add type hints in the abstract method since we only have it there to signal that sub-classes must have an `_update_method` which takes an instance and updates online state, but we don't specify the exact type. However, we could also do `def _update_state(self, x_t: Union[np.ndarray, 'torch.Tensor', 'tf.Tensor'):` and then add `# type: ignore[override]` in the sub-class? This is actually what we did previously...

P.s. its a similar story for the `get_threshold` method...

ascillitoe:
[d190589](https://github.com/SeldonIO/alibi-detect/pull/604/commits/d1905893a31ae8b663479087295b078de5b4da82) removes `save_state` from `save_detector`. The new logic is to always save state if `self.t > 0`. If this is not desired, `.reset()` can be called prior to using `save_detector`.

alibi_detect/cd/tests/test_cvm_online.py
ascillitoe: Yeh fair point, tests should generally pass with any seed, especially if they are unit type tests. The problem at the moment is we have lots of functional tests where we are testing a detector's predictions and checking things like Expected Runtime (ERT) for online detectors. We probably want more granular unit tests in lots of places...

alibi_detect/cd/base_online.py
ascillitoe: @jklaise @mauicv I have got rid of the new `BaseDriftOnline` class. There are stil some existing issues like the ones above, but this PR now shouldn't introduce new ones at least! 

The `save_state`/`load_state` methods are now added via `StateMixin`.

alibi_detect/cd/base_online.py
ascillitoe: Tidied and moved to `StateMixin`. See https://github.com/SeldonIO/alibi-detect/pull/604#discussion_r1072154613.

jklaise (approved):
LGTM! Should we add some documentation somewhere that `save` will save the state by default, and if that's not desired one should call `reset_state` first? Or is it going to be too confusing for now?

ascillitoe:
> LGTM! Should we add some documentation somewhere that save will save the state by default, and if that's not desired one should call reset_state first? Or is it going to be too confusing for now?

Thanks! Will add this documentation now, just realised I added it in https://github.com/SeldonIO/alibi-detect/pull/628 instead of here. Doh!

ascillitoe:
**Final note**: Still need to consider moving `conftest.py` back...

doc/source/overview/saving.md
```diff
@@ -143,26 +143,34 @@ the scikit-learn API.
 ```
 ````
 
-## Saving state
+## Online detectors
 
-[Online drift detectors](../cd/methods.md#online) are stateful, with their state updated upon each `predict` call. 
-When saving an online detector, the `save_state` option controls whether to include the detector's state:
+[Online drift detectors](../cd/methods.md#online) are stateful, with their state updated each timestep `t` (each time
+`.predict()` or `.state()` is called). {func}`~alibi_detect.saving.save_detector` will save the state of online
```
jklaise, line 149: Do you mean `score()` instead of `state()`? On that note, do we even document the usage/use cases of `score()`? If not, perhaps should leave it out.

ascillitoe: Good spot thanks. Also fair point about not really documenting it. I'll remove.

ascillitoe: Removed `.state()` in [d888276](https://github.com/SeldonIO/alibi-detect/pull/604/commits/d88827651e6078e065140ab6c601e4cd0199e224)

mauicv (approved):
LGTM!

PR #605, opened 2022-09-06
Enforce utf-8 when reading README.md in setup.py

### LINKED ISSUE #600: Enforce utf8 decoding to open `README.md` in `setup.py`
In our `setup.py` we open our `READ.md` to fill in the `setuptools` `long_description` metadata (used on the PyPI page). If users have a non-default codec used for PYTHONIOENCODING this can cause `pip install`'s to fail as the `README.md` cannot be read (e.g. see https://github.com/SeldonIO/alibi-detect/issues/597#issuecomment-1225544213). 

We should change the following in the `setup.py`:

```python
def readme():
    with open("README.md") as f:
        return f.read()
```

to:

```python
def readme():
    with open("README.md", encoding="utf-8") as f:
        return f.read()
```

**Note**: The same applies to Alibi.

ascillitoe:
Resolves #600.

Alibi equivalent to be opened after merge.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import find_packages, setup
 
 
 def readme():
-    with open("README.md") as f:
+    with open("README.md", encoding="utf-8") as f:
         return f.read()
 
 


mauicv (approved):
LGTM!

jklaise (approved):

PR #606, opened 2022-09-06
Feature/tabular warning

### LINKED ISSUE #588: Raise a warning if `TabularDrift` used with `categories_per_feature=None`
The `TabularDrift` detector applies `KSDrift` to continuous features and `ChiSquareDrift` to categorical features.

However, if `categories_per_feature` is set to `None`, `KSDrift` is applied to all features, as this implies all features are continuous. We should still allow this behavior, but should/could add a warning to remind users, as well as potentially suggest they simply use `KSDrift` in this case.

ascillitoe:
Resolves #588.

The following warning is raised when a `categories_per_feature` dict is not provided:

```python
warnings.warn('No `categories_per_feature` dict provided so all features are assumed to be categorical. '
              '`KSDrift` will be applied to all features.')
```

### DIFF
--- a/alibi_detect/cd/tabular.py
+++ b/alibi_detect/cd/tabular.py
@@ -3,6 +3,7 @@ from scipy.stats import chi2_contingency, ks_2samp
 from typing import Callable, Dict, List, Optional, Tuple, Union
 from alibi_detect.cd.base import BaseUnivariateDrift
 from alibi_detect.utils.warnings import deprecated_alias
+import warnings
 
 
 class TabularDrift(BaseUnivariateDrift):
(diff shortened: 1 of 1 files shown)


alibi_detect/cd/tabular.py
```diff
@@ -106,6 +107,11 @@ def __init__(
                                  'Dict[int, NoneType], Dict[int, int], Dict[int, List[int]]')
             self.x_ref_categories = categories_per_feature
             self.cat_vars = list(self.x_ref_categories.keys())
+        # No categories_per_feature dict so assume no categorical features present
+        else:
+            self.x_ref_categories, self.cat_vars = {}, []
+            warnings.warn('No `categories_per_feature` dict provided so all features are assumed to be categorical. '
```
arnaudvl, line 113: "... all features are assumed to be numerical." instead of categorical.

arnaudvl (changes requested):

alibi_detect/cd/tabular.py
ascillitoe: Ooops, good spot, thanks!

arnaudvl (approved):

alibi_detect/cd/tabular.py
ascillitoe: Corrected now! 👍🏻

Issue #572, opened 2022-07-26
Handle errors when fetching models/datasets in save/load tests


ascillitoe:
Any errors such as timeouts that occur when fetching huggingface models (and possibly datasets) in the new `saving/tests/test_saving.py` tests will cause the CI to fail:

https://github.com/SeldonIO/alibi-detect/runs/7521461525

We should handle this more gracefully. 

Related:
- https://github.com/SeldonIO/alibi-detect/issues/315

PR #607, opened 2022-09-06
Add some missing logic for failed URI's in datasets and test_saving

ascillitoe:
This PR resolves two issues:

1. https://github.com/SeldonIO/alibi-detect/issues/315#issuecomment-1141822714:  The `try/except RequestException` pattern is added to `corruption_types_cifar10c` in `alibi_detect.datasets`.

2. #572:
   - Skip tests in `test_saving.py` when URI's for huggingface models (e.g. `tokenizer`'s or `embedding`'s) are down.
   - Skip tests when the URI's in `get_movie_sentiment_data` are down.

### DIFF
--- a/alibi_detect/datasets.py
+++ b/alibi_detect/datasets.py
@@ -11,6 +11,7 @@ import requests
 from alibi_detect.utils.data import Bunch
 from alibi_detect.utils.url import _join_url
 from requests import RequestException
+from urllib.error import URLError
 from scipy.io import arff
 from sklearn.datasets import fetch_kddcup99
 
@@ -59,7 +60,11 @@ def fetch_kdd(target: list = ['dos', 'r2l', 'u2r', 'probe'],
     """
 
     # fetch raw data
-    data_raw = fetch_kddcup99(subset=None, data_home=None, percent10=percent10)
+    try:
+        data_raw = fetch_kddcup99(subset=None, data_home=None, percent10=percent10)
+    except URLError:
+        logger.exception("Could not connect, URL may be out of service")
+        raise
 
     # specify columns
     cols = ['duration', 'protocol_type', 'service', 'flag', 'src_bytes', 'dst_bytes',
--- a/alibi_detect/datasets.py
+++ b/alibi_detect/datasets.py
@@ -290,7 +290,11 @@ def corruption_types_cifar10c() -> List[str]:
     url = 'https://storage.googleapis.com/seldon-datasets/'
     folder = 'cifar10c'
     filetype = 'npy'
-    corruption_types = google_bucket_list(url, folder, filetype)
+    try:
+        corruption_types = google_bucket_list(url, folder, filetype)
+    except RequestException:
+        logger.exception("Could not connect, URL may be out of service")
+        raise
     corruption_types.remove('labels')
     return corruption_types
 
--- a/alibi_detect/datasets.py
+++ b/alibi_detect/datasets.py
@@ -290,11 +290,7 @@ def corruption_types_cifar10c() -> List[str]:
     url = 'https://storage.googleapis.com/seldon-datasets/'
     folder = 'cifar10c'
     filetype = 'npy'
-    try:
-        corruption_types = google_bucket_list(url, folder, filetype)
-    except RequestException:
-        logger.exception("Could not connect, URL may be out of service")
-        raise
+    corruption_types = google_bucket_list(url, folder, filetype)
     corruption_types.remove('labels')
     return corruption_types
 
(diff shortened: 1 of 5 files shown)


alibi_detect/datasets.py
```diff
@@ -290,7 +290,11 @@ def corruption_types_cifar10c() -> List[str]:
     url = 'https://storage.googleapis.com/seldon-datasets/'
     folder = 'cifar10c'
     filetype = 'npy'
-    corruption_types = google_bucket_list(url, folder, filetype)
+    try:
+        corruption_types = google_bucket_list(url, folder, filetype)
+    except RequestException:
+        logger.exception("Could not connect, URL may be out of service")
+        raise
```
jklaise, line 297: This exact error with the message is already caught and raised in `google_bucket_list`, not sure we need to do this again here?

alibi_detect/saving/tests/datasets.py
```diff
@@ -63,4 +65,7 @@ def data_synthetic_nd(data_shape):
 class TextData:
     @staticmethod
     def movie_sentiment_data():
-        return get_movie_sentiment_data()
+        try:
+            return get_movie_sentiment_data()
+        except RequestException:
+            pytest.skip('Movie sentiment dataset URL down')
```
jklaise, line 71: How does this actually work? The test within which this data is used is magically skipped?

alibi_detect/saving/tests/test_saving.py
```diff
@@ -202,7 +202,10 @@ def nlp_embedding_and_tokenizer(model_name, max_len, uae, backend):
     backend = 'tf' if backend == 'tensorflow' else 'pt'
 
     # Load tokenizer
-    tokenizer = AutoTokenizer.from_pretrained(model_name)
+    try:
+        tokenizer = AutoTokenizer.from_pretrained(model_name)
+    except OSError:
```
jklaise, line 207: Is `OSError` the only one that can happen? Seems a bit weird here, is it because the network call was done somewhere else and failed silently and the loading is attempted from the locally downloaded file (which doesn't exist) or is `from_pretrained` performing the network call but `hugginface` isn't raising network errors up the stack?

jklaise (approved):
LGTM, left some questions.

alibi_detect/datasets.py
ascillitoe: Thanks, you're totally correct. I have removed the above. 

Think I got confused as was trying to address this issue https://github.com/SeldonIO/alibi-detect/issues/315#issuecomment-1141822714. The actual correction needs to be to add a `pytest.skip` if the above error is raised when `corruption_types_cifar10c` is used within `test_datasets.py`. I'll do this now.

alibi_detect/saving/tests/datasets.py
ascillitoe: Embarrassingly I actually don't quite understand why this works. But I tested it by changing `MOVIESENTIMENT_URLS` to contain incorrect url's, and running `pytest alibi_detect/saving/tests/test_saving.py -k test_save_preprocess_nlp -sv`, and `SKIPPED (Movie sentiment dataset URL down)` is returned. Magic indeed! 

I think this is a special feature of the "cases" from `pytest-cases` that we use to represent datasets in the new saving tests. They're like fancy fixtures, and have clever capabilities like being able to use [pytest marks](https://smarie.github.io/python-pytest-cases/#pytest-marks-skip-xfail) (i.e. `skip`).

alibi_detect/saving/tests/test_saving.py
ascillitoe: This is the full error traceback:

https://github.com/SeldonIO/alibi-detect/runs/7521461525

I think the `huggingface` strategy must be what you described i.e. if a `HTTPError` then they try to load from cache or a local directory. I tried passing an incorrect `model_name` to `from_pretrained` and using `HTTPError` or `ValueError` in `except`, but then I get the following:

`OSError: We couldn't connect to 'https://huggingface.co/' to load this model, couldn't find it in the cached files and it looks like bert-base-cased is not the path to a directory containing a config.json file.`

Seems to be like the other errors are not raised up the stack as you say?

ascillitoe: Maybe the solution might just be to do `except (OSError, HTTPError)`?

jklaise: Yea probably, in case they decide to change the error type at some point.

ascillitoe: Done!

PR #614, opened 2022-09-08
Ignore tests in codecov

ascillitoe:
Update codecov setting to ignore `test` directories.

Alibi equivalent to be opened post-approval.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -23,3 +23,6 @@ comment:
   layout: "reach,diff,flags,tree"
   behavior: default
   require_changes: no
+
+ignore:
+  - "**/*test*"  # ignore tests themselves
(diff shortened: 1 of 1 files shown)


ascillitoe:
Revised coverage below. Looks like the desired files are being ignored...

image

jklaise:
Not sure why the folders/files still appear on the PR view:
image

Otherwise looks good (do we need to ignore `conftest.py` too or are they covered by ignoring `tests` folders?). Care to open the same in `alibi`? :)

jklaise (approved):

ascillitoe:
> Not sure why the folders/files still appear on the PR view:

I'm hoping it is because `codecov.yml` needs to be in `master` before it takes effect. My screenshot above was from me manually running `codecov` on my `master` fork and uploading to my personal codecov account. Not 100% sure how else to verify without just merging and seeing what happens...

PR #624, opened 2022-09-20
Revert "Tabular classifier drift example - dummy trap (#557)"

RobertSamoilescu:
This reverts commit 4e2419877f89cb7e72721644a947f8b390f123fb.
Fixes incorrect suggestion to drop first column for one-hot encoding for data preprocessing. 

Check this [gist](https://gist.github.com/RobertSamoilescu/c609ded901c6c50388b6648a73066fed) to see how dropping first column can affect the model performance.


jklaise (approved):
Thanks Robert, same comments as here: https://github.com/SeldonIO/alibi/pull/773#issuecomment-1252098035

PR #628, opened 2022-09-21
Change tensorflow model format to `SavedModel` to support sub-classed models

ascillitoe:
This PR changes the format we use to serialize TensorFlow models from the old HDF5 to the newer [SavedModel](https://www.tensorflow.org/guide/saved_model) format.

## Motivation

As well as being the default (and recommended) TensorFlow model format, the `SavedModel` format has the advantage of supporting serialisation of sub-classed tensorflow models. That is, models constructed by subclassing `tf.keras.Model`, rather than by using `tf.keras.Sequential`, `tf.keras.models.Model` etc.

## Limitations 

- The new format improves the handling of custom tensorflow objects (such as layers and models) slightly. If these are not passed to `tf.keras.models.load_model` via `custom_objects` (or registered with `@tf.keras.utils.register_keras_serializable()`) the model will be loaded as a `keras.saving.saved_model.load.<model_class>`*.  This is a _rough copy_ of the original serialized model, that behaves the same wrt inference, but cannot be cloned (which is done in a number of learned detectors such as `ClassifierDrift`). To load the fully-functional model, all custom objects must be supplied at load time.
- The [HiddenOutput](https://docs.seldon.io/projects/alibi-detect/en/stable/api/alibi_detect.cd.tensorflow.html?highlight=hiddenoutput#alibi_detect.cd.tensorflow.HiddenOutput) class does not work for subclassed models. Therefore, subclassed models cannot be saved/loaded when `layer` is specified in the [ModelConfig](https://docs.seldon.io/projects/alibi-detect/en/stable/api/alibi_detect.saving.schemas.html?highlight=modelconfig#alibi_detect.saving.schemas.ModelConfig).

*in `tensorflow>=2.9`. In older versions, loading of the model will fail completely if the custom objects are not provided.

## Main changes
- Changed `save_format` from `'h5'` to `'tf'` in `save_model` and `load_model`, although stuck with `h5` for the legacy save/load functions.
- Removed support for passing a `custom_objects` dictionary via config since support for this was very flaky. Custom objects in the dictionary could only realistically be specified as registered object strings (`'@mymodel` etc). However, this is confusing as tensorflow already has its own `@tf.keras.utils.register_keras_serializable()` decorator.
- `load_detector` now allows arbitrary `kwargs`, which are passed to `tf.keras.models.load_model` (or `torch.load`). This is to be used to provide the `custom_objects` at load time (see example below).
- Added subclassed models to CI.

## Example

[Example notebook](https://gist.github.com/ascillitoe/87ee6464378dc95e5129564352516b3a) demonstrating serialisation of a detector with a subclassed tensorflow model. Observe how the custom objects must be passed to `load_detector` in order to avoid the error `KeyError: 'layers'`.

## Backwards compatibility

`tf.keras.models.load_model` automatically detects whether a given model path represents a `h5` model or `SavedModel`. This means we should be backwards compatible, in that we can simply move to saving `SavedModel`'s, but still support loading of legacy `h5` models.

## TODO's

- [x] Better document limitations of `SavedModel` format in docs.
~~- [ ] Add a more involved example of passing custom objects to `load_detector`.~~ - _More challenging than first envisaged; I wanted to demonstrate on the amazon example, where a subclassed `ClassifierTF` model is used, but saving here is not supported due to the use of `tokenize_transformer`. This is not related to subclassed models so would like to leave for a follow-up PR._
- [ ] CHANGELOG.md - see https://github.com/SeldonIO/alibi-detect/pull/628#pullrequestreview-1272835501

## Old notes etc

[This notebook](https://gist.github.com/ascillitoe/bab98843256d99be14ed6fa4a85ff640) contains some experiments run to explore limitations wrt to the `SavedModel` format.


alibi_detect/saving/tensorflow/_loading.py
```diff
@@ -274,7 +274,7 @@ def load_detector_legacy(filepath: Union[str, os.PathLike], suffix: str, **kwarg
             emb, tokenizer = load_text_embed(filepath)
         try:  # legacy load_model behaviour was to return None if not found. Now it raises error, hence need try-except.
             model = load_model(filepath, load_dir='encoder')
-        except FileNotFoundError:
+        except OSError:
```
ascillitoe, line 277: Changed to `OSError` because we are now relying on `tf.keras.model.load_model` to raise an error when loading fails, and this is what it raises...

alibi_detect/saving/tensorflow/_saving.py
```diff
@@ -150,6 +150,9 @@ def save_embedding_config(embed: TransformerEmbedding,
     return cfg_embed
 
 
+#######################################################################################################
+# TODO: Everything below here is legacy saving code, and will be removed in the future
```
ascillitoe, line 154: This comment was in the wrong place...

alibi_detect/saving/tensorflow/_saving.py
```diff
@@ -237,24 +237,24 @@ def save_detector_legacy(detector, filepath):
         save_tf_vae(detector, filepath)
     elif isinstance(detector, (ChiSquareDrift, ClassifierDrift, KSDrift, MMDDrift, TabularDrift)):
         if model is not None:
-            save_model(model, filepath, save_dir='encoder')
+            save_model(model, filepath, save_dir='encoder', save_format='h5')
```
ascillitoe, line 240: Stick to saving in `h5` format for legacy saves...

alibi_detect/saving/tests/test_saving.py
```diff
@@ -204,7 +204,7 @@ def nlp_embedding_and_tokenizer(model_name, max_len, uae, backend):
 
     # Load tokenizer
     try:
-        tokenizer = AutoTokenizer.from_pretrained(model_name + 'TODO')
+        tokenizer = AutoTokenizer.from_pretrained(model_name)
```
ascillitoe, line 207: This had been accidentally introduced in a previous PR. It meant any tests using `nlp_embedding_and_tokenizer` were being skipped 😬

alibi_detect/cd/utils.py
```diff
@@ -118,5 +118,7 @@ def get_input_shape(shape: Optional[Tuple], x_ref: Union[np.ndarray, list]) -> O
     else:
         logger.warning('Input shape could not be inferred. '
                        'If alibi_detect.models.tensorflow.embedding.TransformerEmbedding '
-                       'is used as preprocessing step, a saved detector cannot be reinitialized.')
+                       'is used as preprocessing step, a saved detector cannot be reinitialized. '
+                       'Additionally, if the detector is a subclassed `tf.keras.Model`, it is '
+                       'unlikely to be serializable.')
```
mauicv, line 123: Perhaps a silly question, but what is this warning about specifically? Does it just tell the user that we can't infer the input shape because the model can't be initialized? And that's because the models cannot be serialized? If that's the case should we issue a warning at the save step?

alibi_detect/saving/_tensorflow/tests/test_saving_tf.py
```diff
@@ -54,29 +59,63 @@ def test_load_optimizer_type(backend):
...
 @parametrize_with_cases("data", cases=ContinuousData.data_synthetic_nd, prefix='data_')
-@parametrize('model', [encoder_model])
+@parametrize('model', [encoder_model, encoder_model_subclassed])
 @parametrize('layer', [None, -1])
 def test_save_model_tf(data, model, layer, tmp_path):
     """
     Unit test for _save_model_config and _load_model_config with tensorflow model.
     """
+    if layer is not None and isinstance(model, EncoderTF):
+        pytest.skip("Don't test `layer != None` when the model is a subclassed model.")
```
mauicv, line 69: Perhaps add the reason why

alibi_detect/saving/loading.py
```diff
@@ -498,7 +505,7 @@ def resolve_config(cfg: dict, config_dir: Optional[Path]) -> dict:
         elif isinstance(src, dict):
             backend = cfg.get('backend', Framework.TENSORFLOW)
             if key[-1] in ('model', 'proj'):
-                obj = _load_model_config(src)
+                obj = _load_model_config(src, **kwargs)
```
mauicv, line 508: Just a general thought, but is it an issue if later we want multiple functions to receive `kwargs`? Is that likely to happen? I we had a `tf_load_Kwargs` would it be better? :thinking:

ascillitoe: Yeh true it would be an issue... but maybe we could just deal with that when/if we need to in the future?!

mauicv (approved):
LGTM!

alibi_detect/saving/_tensorflow/saving.py
```diff
@@ -101,24 +98,31 @@ def save_model(model: tf.keras.Model,
...
+    if filepath.suffix == '.h5':
+        model_dir = filepath.parent
+        save_format = 'h5'
+    else:
+        model_dir = filepath
+        save_format = 'tf'
 
-    # save model
-    model_path = model_path.joinpath(filename + '.h5') if save_format == 'h5' else model_path
+    # create folder to save model in  # TODO - is this needed?
```
ascillitoe, line 113: Note to self, check this...

ascillitoe:
The commits from [f61af31](https://github.com/SeldonIO/alibi-detect/pull/628/commits/f61af31012f815f8f98f50934b51ac602a3ec899) onwards contain two primary changes (based on discussion with @mauicv):

1. Subclassed tf models must have been built/called before they can be saved. It was decided that attempting to perform a dummy call prior to saving was risky. Instead, the `ValueError` that occurs in `alibi_detect.saving._tensorflow.save_model` is caught and re-raised with a more informative error message.
2. For custom layers or subclassed models, various errors can occur during inference or cloning if custom objects are not properly provided at load time. The errors are often unclear, and surface from a variety of places e.g. in detector `predict` methods etc. `alibi_detect.saving._tensorflow.load_model` runs some basic checks on the loaded model and raises a warning if any problems are detected. This allows problems to be discovered when the detector is first loaded, instead of having to wait until prediction time.

alibi_detect/saving/_pytorch/loading.py
```diff
@@ -18,6 +18,7 @@
 
 def load_model(filepath: Union[str, os.PathLike],
```
jklaise, line 19: Extra functionality sneaking into this PR... Worth adding changelog entries to this PR so everything is documented and not missed upon release?

alibi_detect/saving/_tensorflow/loading.py
```diff
@@ -21,6 +21,7 @@
 from alibi_detect.models.tensorflow.autoencoder import (AE, AEGMM, VAE, VAEGMM,
                                                         DecoderLSTM,
                                                         EncoderLSTM, Seq2Seq)
+from alibi_detect.utils.tensorflow.misc import check_model
```
jklaise, line 24: Minor, but why not keep the function in this module since it's specifically used during loading?

alibi_detect/utils/tensorflow/misc.py
```diff
@@ -95,3 +96,40 @@ def clone_model(model: tf.keras.Model) -> tf.keras.Model:
...
+                             'a subclassed model has been loaded without all custom objects being provided.')
+
+        # Check model cloning doesn't raise error
+        clone_model(model)
+
+    except Exception as error:
+        if raise_error:
+            raise ValueError(msg) from error
+        else:
+            warnings.warn(msg + f"Original error message: \n\t{error}")
```
jklaise, line 135: Not sure this works that well together, e.g. we check if cloning works but it any case, the error is never raised. So effectively this would blow up again when using inside detectors that do use cloning? Or is the intention to call this method there also?

mauicv (approved):

alibi_detect/saving/_tensorflow/loading.py
```diff
@@ -36,36 +37,31 @@
...
     """
-    # TODO - update this to accept tf format - later PR.
-    model_dir = Path(filepath)
-    model_name = filename + '.h5'
-    # Check if model exists
-    if model_name not in [f.name for f in model_dir.glob('[!.]*.h5')]:
-        raise FileNotFoundError(f'{model_name} not found in {model_dir.resolve()}.')
-    model = tf.keras.models.load_model(model_dir.joinpath(model_name), custom_objects=custom_objects)
+    # Load model
+    model = tf.keras.models.load_model(filepath, **kwargs)
```
jklaise, line 61: Seems like we're throwing away the validation code for the existence of the model? Or is it done from higher up in another caller?

alibi_detect/saving/_tensorflow/loading.py
ascillitoe: Mmn I put it here so that its next to `clone_model`, which is quite related. But can move back to `saving` since we only use it there...

alibi_detect/saving/_pytorch/loading.py
ascillitoe: Not 100% clear what you mean here? The functionality of passing kwargs to `load_detector`? It is extra functionality but is interlinked with the PR, as `custom_objects` needs to be passed to `load_detector`.

alibi_detect/saving/_tensorflow/loading.py
```diff
@@ -36,36 +37,31 @@
 
 
 def load_model(filepath: Union[str, os.PathLike],
-               filename: str = 'model',
-               custom_objects: dict = None,
                layer: Optional[int] = None,
```
jklaise, line 40: Minor, but I don't fully agree with omitting `filename` from these internal save/load functions. What we save for in function signature, we pay at every callsite, having to remember to do `.joinpath(filename)`. (OTOH in the old behaviour, having a default `model` name is also likely not desirable as forgetting to set it would result in a perhaps unexpected default).

doc/source/overview/saving.md
```diff
@@ -120,28 +121,47 @@ cd = ClassifierDrift(x_ref, model, backend='sklearn', p_val=.05, preds_type='pro
...
+[here](https://www.tensorflow.org/tutorials/keras/save_and_load#saving_custom_objects)).
 ```
 
-```{tab-item} PyTorch
-PyTorch models are serialized by saving the [entire model](https://pytorch.org/tutorials/beginner/saving_loading_models.html#save-load-entire-model)
-using the [dill](https://dill.readthedocs.io/en/latest/index.html) module. Therefore, Alibi Detect should support any PyTorch 
-model that can be saved and loaded with `torch.save(..., pickle_module=dill)` and `torch.load(..., pickle_module=dill)`.
+```{note}
+
+- The {obj}`~alibi_detect.cd.tensorflow.HiddenOutput` utility class is not currently compatible with subclassed models.
```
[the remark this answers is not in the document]
ascillitoe: `HiddenOutput` works by building a new `tf.keras.Model` from the original model's `input` and `layers` attributes.

https://github.com/SeldonIO/alibi-detect/blob/c0c5e644a1faaadc23d375f31e27a13110ce6bac/alibi_detect/cd/tensorflow/preprocess.py#L67-L85

This works fine for `tf.keras.Model`'s constructed from `tf.keras.Sequential` etc as they have a pre-defined structure to their these attributes. It doesn't work out-of-the-box for subclassed models, presumably because there are quite a few different ways you can construct these models internally. Probably worth opening an issue to explore this one further...

alibi_detect/saving/_pytorch/loading.py
jklaise: No need, but would appreciate a changelog as part of the PR.

doc/source/overview/saving.md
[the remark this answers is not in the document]
jklaise: I see, feels like a docstring is required for `HiddenOutput`(and other utility classes) to describe behaivour/limitations. For another PR though.

jklaise (approved):
Generally LGTM, but would appreciate a changelog with all the changes. Also, noting that there's a few subtle changes in the behaviour of internal functions for saving/loading, need to be extra vigilant new bugs haven't been introduced.

alibi_detect/utils/tensorflow/misc.py
ascillitoe: The problem I am having here is that there are a large number of possible failure modes that can occur due to not specifying the `custom_objects` properly. Depending on the exact tf version, and whether the model itself is subclassed, or just layers, either `ValueError`, `TypeError`'s or `NotImeplemented` errors are raised, and this is either during inference or cloning. 

The widest net I've been able to cast is to simply test if cloning works, and to check if there might be problems at inference, check whether the model is a `RevivedNetwork`, which is what it's loaded as if custom objects are missing (I don't try to actually call the model as data isn't available at this point). 

A wide net is nice in term of hopefully catching most errors, but has the downside of throwing errors when things might have actually worked. For example, the `RevivedNetwork`'s generally work for inference but not cloning. So a user can actually get away with not passing `custom_objects` if the model is just to be used for preprocessing. Hence why I went with a warning instead of error...

Maybe one compromise is to raise a warning if the model is a `RevivedNetwork`, since this can cause issues, but raise the `ValueError` (with the more coherent message) if cloning fails?

doc/source/overview/saving.md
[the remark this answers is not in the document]
ascillitoe: I'll open an issue shortly 👍🏻

alibi_detect/utils/tensorflow/misc.py
jklaise: I'm not sure, cloning failure is only relevant for a specific set of detectors, so somehow feels like it should be checked only for that subset. Maybe leave this as is and wrap the relevant detector calls to `clone()` in `try/except` with a customized Alibi error message?

alibi_detect/saving/_tensorflow/loading.py
jklaise: OK that makes sense.

alibi_detect/saving/_tensorflow/loading.py
ascillitoe: Yeh very true. Pretty much the only reason I made this change is that the `filename` is only there for legacy loading (legacy as in, saving to `.h5`, and legacy as in loading files with different names such as `encoder.h5`). For the "modern" loading we simply do:

```python
    if flavour == Framework.TENSORFLOW:
        model = load_model_tf(src, layer=layer, **kwargs)
```

So I saw it as a trade-off wrt to carrying around more complexity in `load_model` to facilitate the legacy functionality in `load_detector_legacy`, or add some complexity to the calls in `load_detector_legacy` to simplify the `load_model` function... (which just happens to be used for modern and legacy loading). 

Similar story for `save_model`...

ascillitoe: > Also, noting that there's a few subtle changes in the behaviour of internal functions for saving/loading, need to be extra vigilant new bugs haven't been introduced.

p.s. below is very true though... tweaking anything to do with legacy save/load does bring the potential for bugs like the ones `v0.10.5` fixed. I've run the same tests of loading old artefacts we ran in https://github.com/SeldonIO/alibi-detect/pull/732 and everything passes, but that isn't 100% comprehensive...

alibi_detect/utils/tensorflow/misc.py
ascillitoe: Sorry, just coming back on this, I've thought about it a bit more. There are just so many different failure modes, and the problem with just checking if cloneable is that there is one nasty failure mode where inference works, cloning works, BUT inference after cloning doesn't work 🤯 (this occurs in example 5 [here](https://colab.research.google.com/gist/ascillitoe/bab98843256d99be14ed6fa4a85ff640/tensorflow_save-load.ipynb), the custom `call` method is lost when cloning...). 

This is what motivates casting the wide net by checking if a `RevivedNetwork` too. But then throwing a warning for this would prevent users getting away with cases that would otherwise work with no `custom_objects` provided (generally when no cloning). 

I see two options:
1. Leave pretty much as is. Downside at the moment is by swallowing the `ValueError` from cloning and raising as a warning, it can be somewhat lost in the later actual errors that occur (i.e. when cloning down for real). 
2. Turn the warning into an error. Be up front about requiring `custom_objects` to be passed in all cases where there are custom objects involved. This is slightly less convenient for the user, but means we don't have to worry about these many different failure modes that might also change in future versions.

ascillitoe: Yes exactly. It would mean any models containing custom classes loaded without providing those custom classes (via registering, or `custom_objects` kwarg) would be pretty much guaranteed to fail. Thinking about it more this would probably be what I would lean towards. Although to be clear, this does also mean that, for example, if a user had a relatively simple `tf.keras.Sequential`, but with one custom layer, they would have to provide this layer at load time even if they don't use the model in a detector with any cloning...

ascillitoe: The `tensorflow` behaviour with this has been so fluid in versions 2.9, 2.10 and 2.11 that I do think it'd be the safest option though... (for now)

ascillitoe:
Postponing this to `v0.12.0` so that we can combine it with deprecation of legacy saving and #723 .

PR #635, opened 2022-09-27
Minor save/load refactoring

ascillitoe:
This PR contains a number of minor changes to the `saving` submodule. Some of these are cherry-picked from #628, and it is intended to be a pre-cursor to the incoming PR's for `sklearn` and `pytorch` serialization support.

Changes:
- Adust the way attributes in `LARGE_ARTEFACTS` are handled in `alibi_detect/base.py`. Specifically, don't attempt to remove/add them from/to the config unless the `_nested_detector` actually possesses the attribute. This change isn't strictly necessary right now, but will be helpful for changes like those in #638, where some attributes such as `model` might not exist in `_nested_detector`'s e.g. a `model` doesn't exist in the `KSDrift` detector nested inside of `RegressorUncertaintyDrift`. (Note that more thought is required as to whether to include `model` in `LARGE_ARTEFACTS`, since we need to be careful about state leaking into config...).
- Add unsupported backends to the error message raised in `_load_detector_config`.
- Fix a few bugs in `test_saving` and rename some variables. 
- Add a "Supported ML models" section to the docs. This will be fleshed out further in the incoming PR's.
- Made the `alibi_detect.saving.tensorflow` submodule private. This was the original intention, but it was changed in #537. Having it private will allow us to make changes to the internals without worrying about breaking changes and deprecation warnings. In the future we should review public vs private patterns in other backend submodules (e.g. in `cd` etc).


alibi_detect/base.py
```diff
@@ -119,16 +119,14 @@ def get_config(self) -> dict:  # TODO - move to BaseDetector once config save/lo
...
             # Get config (stored in top-level self)
             cfg = self.config
-            # Get low-level nested detector (if needed)
-            detector = self._detector if hasattr(self, '_detector') else self  # type: ignore[attr-defined]
-            detector = detector._detector if hasattr(detector, '_detector') else detector  # type: ignore[attr-defined]
             # Add large artefacts back to config
             for key in LARGE_ARTEFACTS:
-                if key in cfg:  # self.config is validated, therefore if a key is not in cfg, it isn't valid to insert
-                    cfg[key] = getattr(detector, key)
+                if key in cfg and hasattr(self._nested_detector, key):
```
jklaise, line 124: Just to clarify, the previous implementation would cause an opaque error if the key doesn't exist in the nested detector? What's a concrete example of a `cfg` for which this would happen?

alibi_detect/saving/tests/test_saving.py
```diff
@@ -350,7 +353,7 @@ def test_save_ksdrift_nlp(data, preprocess_fn, max_len, enc_dim, tmp_path):
                  p_val=P_VAL,
                  preprocess_fn=preprocess_fn,
                  preprocess_at_init=True,
-                 input_shape=(max_len,),
+                 input_shape=(768,),  # hardcoded to bert-base-cased for now
```
jklaise, line 356: Why/what was the bug for using `max_len` parameter previously?

jklaise (approved):
LGTM

alibi_detect/base.py
ascillitoe: For the last implementation and the new one, we are working on the assumption that there is a one-to-one mapping between a detector's pydantic model and its args/kwargs, which are passed into `_set_config` as `inputs`. Therefore, in the previous implementation, if a `key` was in `inputs`, it was set to `None`. Then, later in `get_config`, we could safely check whether it needed to be set back to its actual value in `cfg` by checking if the `key` existed in `cfg` (because a `key` existing in `cfg` means it must have been in `inputs` earlier). 

The above is all good for the current set of `LARGE_ARTEFACTS = ['x_ref', 'c_ref', 'preprocess_fn']`, since we can safely  get the real `key` value back from its corresponding attribute in `_nested_detector` (note that the `_nested_detector` bit is because for detectors like `MMDDrift`, `x_ref` etc are stored in `self._detector.x_ref` not `self.x_ref`). The problem might arise in the future for some artefacts like `model`. For `ClassifierDrift` it would be fine because the `model` key is stored in the `self.model` attribute. For `RegressorUncertaintyDrift` we have a problem, because `model` isn't stored as an attribute anywhere. Another solution would be to store it, but the question then is where. It is a weird one because its `_nested_detector` is a `KSDrift` which itself doesn't have a `model` attribute:

https://github.com/SeldonIO/alibi-detect/blob/df7edca2f565b241062c5fdf4811ef50821e3ce6/alibi_detect/cd/model_uncertainty.py#L118-L127

So if we add a `model` attribute to `self` we're in a situation where `x_ref` would be in `self._detector.x_ref` but `model` is in `self.model`. 

The current implementation deals with this by simply skipping the `LARGE_ARTEFACT` procedure for the uncertainty detectors. I think this probably leads to duplication of the `model`, since it is in the `cfg` and the `self.model_fn` (in the future we could add a `model` `@property` to the uncertainty detectors to be more clever about this...). In any case, because `model` is more complciated, and I'm worried about state leaking into config (if the model has been trained), I've left `model` out of `LARGE_ARTEFACTS` for now.

**TL'DR**: The new implementation shouldn't actually change anything right now, but it should be more robust in any cases where an artefact is not stored in `_nested_detector`. `model` is one example, but there could be others I haven't thought of...

alibi_detect/saving/tests/test_saving.py
ascillitoe: The `input_shape` should be the model input data's shape, after it has passed through any nlp embedding and tokenizer (i.e. 768 for bert-base-cased). I had incorrectly set it to `max_len`, which is passed to the HuggingFace tokenizer's `max_length` kwarg, which sets the max length for the tokenizer to pad/truncate sentences to. The bug doesn't actually cause a problem at the moment, but it will for subclassed tensorflow models in the future (once we moved to `SavedModel` format).

alibi_detect/saving/_tensorflow/__init__.py
```diff
@@ -11,11 +11,23 @@
...
     names=['save_detector_legacy', 'save_model_config']
 )
 
 get_tf_dtype = import_optional(
-    'alibi_detect.saving.tensorflow._conversions',
+    'alibi_detect.saving._tensorflow.conversions',
     names=['get_tf_dtype']
 )
+
+__all__ = [
```
ascillitoe, line 23: @mauicv I just added this `__all__` here so that `import_optional` etc doesn't show up in `_tensorflow`. Guess it doesn't really matter since `_tensorflow` is private but maybe makes sense to add for convention anyway?

alibi_detect/saving/_tensorflow/saving.py
```diff
@@ -150,6 +150,9 @@ def save_embedding_config(embed: TransformerEmbedding,
     return cfg_embed
 
 
+#######################################################################################################
+# TODO: Everything below here is legacy saving code, and will be removed in the future
```
ascillitoe, line 154: This comment was in the wrong place previously.

ascillitoe:
@jklaise apologies, I've just added two more minor things since your approval. See comments above.

mauicv (approved):
LGTM

PR #633, opened 2022-09-27
Limit sphinx-autodoc-typehints upper bound

ascillitoe:
Limit `sphinx-autodoc-typehints` to `<1.19.3` due to https://github.com/tox-dev/sphinx-autodoc-typehints/issues/259 and https://github.com/tox-dev/sphinx-autodoc-typehints/issues/260.

`alibi` equivalent: https://github.com/SeldonIO/alibi/pull/790

PR #636, opened 2022-09-27
Fix numbered list in docstring

ascillitoe:
The current docstring leads to the following sphinx error:

```bash
docstring of alibi_detect.utils.missing_optional_dependency.err_msg_template:3: ERROR: Unexpected indentation.
```


mauicv (approved):
Ah, I was wondering why this was causing an error on another PR I just submitted! LGTM!

ascillitoe:
Cheers @mauicv !

ascillitoe:
P.s. @mauicv this wasn't actually the cause of the error in your PR. This error is swallowed during the sphinx build so doesn't cause a CI failure. 

Your error is likely down to a problem with `sphinx-autodoc-typehints` 1.19.3, which should be fixed by https://github.com/SeldonIO/alibi-detect/pull/633.

PR #639, opened 2022-09-29
Revert Python 3.10.6 pin

### LINKED ISSUE #620: Monitor `mypy` releases and revert to Python 3.10 in CI when issue fixed
#619 explicitly sets our CI to use `Python 3.10.6` until [this issue](https://github.com/python/mypy/issues/13627) is fixed. We should monitor the upcoming `mypy 0.980` and `0.990` releases and revert to setting `3.10` in CI once the issue is fixed.

ascillitoe:
Revert #619 now that mypy 0.981 is out. See https://github.com/python/mypy/issues/13627#issuecomment-1259095263.

Resolves #620.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,12 +30,12 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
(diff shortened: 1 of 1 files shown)


mauicv (approved):
LGTM!

PR #641, opened 2022-09-29
Remove config_spec versioning

ascillitoe:
This PR removes the `config_spec` version number. This was added in v0.10.0, as an addition to the usual `version` number. This was intended to be distinct from `version`, with the idea being it would be used to signify changes to the format of the alibi-detect detector configuration file. However, it has proven to be of limited use, and leads to confusion (+ additional maintainability version). Since we don't guarantee save/load compatibility across `version`'s, the `config_spec` number is simply not needed.


jklaise (approved):
LGTM. Perhaps worth keeping with the convention of adding lines to the development version of the changelog?

ascillitoe:
> LGTM. Perhaps worth keeping with the convention of adding lines to the development version of the changelog?

Good idea! Will do that now

PR #642, opened 2022-09-29
Support for serializing detectors with scikit-learn backends and/or models

ascillitoe:
This PR implements save/load support for sklearn (and xgboost) backends and models.

The primary change addition is a `saving._sklearn` subpackage, containg `save_model` and `load_model` functions that use `joblib` to serialize/deserialize sklearn (and xgboost) models.

Additionally, since this PR introduces a new "flavour" of models/backends to serialize (in addition to the existing `tensorflow`), this PR includes a number of changes to better handle serialization of multiple flavours/models:

1. The detector `backend` config field is no longer tied to the `model` (or `preprocess_fn`) type. This gives us the flexibility to allow mixing and matching of models in the future e.g. a scikit-learn preprocessing model could be used with a `ClassifierDrift(..., backend='tensorflow')` detector (note: an sklearn `preprocess_drift` does not yet exist, just an example!). Also, this change means we do not have the confusing side-effect of `backend='tensorflow'` being written to detector config files without a backend, such as for `KSDrift`. We can also be more be granular when validating `backend` for different detectors.
2. The above is achieved via a new `flavour` field in `ModelConfig` and `EmbeddingConfig`. The aim is to make these configs more self-contained i.e. `model.flavour` tells us whether `model.src` refers to a tensorflow, pytorch or sklearn model, and we can therefore decide how to load it. Self-contained/atomic configs should help with the move to OOP serialization that @mauicv and I have been talking about...
3. The saving tests have been significantly refactored, with backend-specific tests being moved to the backend-specific saving subpackages.


alibi_detect/saving/_sklearn/__init__.py
```diff
@@ -0,0 +1,7 @@
+from alibi_detect.saving._sklearn.saving import save_model_config as save_model_config_sk
+from alibi_detect.saving._sklearn.loading import load_model as load_model_sk
```
ascillitoe, line 2: Added `__init__.py`'s here for consistency with the `_tensorflow` subpackage. Not particularly important since `_saving._sklearn` is private anyway...

alibi_detect/saving/_sklearn/tests/test_saving_sk.py
```diff
@@ -0,0 +1,32 @@
...
+from alibi_detect.saving.loading import _load_model_config
+from alibi_detect.saving.saving import _path2str, _save_model_config
+from alibi_detect.saving.schemas import ModelConfig
+
+backend = param_fixture("backend", ['sklearn'])
+
+
+@parametrize_with_cases("data", cases=ContinuousData.data_synthetic_nd, prefix='data_')
+@parametrize('model', [classifier_model, xgb_classifier_model])
+def test_save_model_sk(data, model, tmp_path):
```
ascillitoe, line 15: We arguably could be more specific about what tests we keep in `saving/tests/test_saving.py` and what ones we keep here.

For now, I've adopted moving tests that are very backend specific to the backend saving subpackages (e.g. `_save_model_config` is tested in `saving/_sklearn/tests/test_saving_sk.py` when `model` is a sklearn or xgboost model). However, for the more generic functional tests such as `test_save_ksdrift`, I test them with both `backend`'s in `saving/tests/test_saving.py` to minimise duplication. There are some other tests in `saving/tests/test_saving.py` such as `test_save_optimizer` that we could probably think about moving to the backend-specific testing files in the future...

Also note, our more usual convention would be to test `saving._sklearn.save_model` here, instead of the parent function `saving._save_model_config`. I am testing `_save_model_config` here so that we also test the logic inside, which should call `saving._sklearn.save_model` if it recognises `model` to be a `sklearn.base.BaseEstimator`.

alibi_detect/saving/_tensorflow/loading.py
```diff
@@ -69,7 +69,7 @@ def load_model(filepath: Union[str, os.PathLike],
     return model
 
 
-def prep_model_and_emb(model: Optional[Callable], emb: Optional[TransformerEmbedding]) -> Callable:
+def prep_model_and_emb(model: Callable, emb: Optional[TransformerEmbedding]) -> Callable:
```
ascillitoe, line 72: `model` no longer `Optional`, since the logic [here](https://github.com/SeldonIO/alibi-detect/pull/642/files#diff-1da3b4e01d459b964951c3287ce414a19a433192bb74ea41bc4520f026f7b9dfR226-R231) is changed so that `prep_model_and_emb` is not called if `model is None`.

alibi_detect/saving/loading.py
```diff
@@ -216,14 +215,20 @@ def _load_preprocess_config(cfg: dict,
             emb = kwargs.pop('embedding')  # embedding passed to preprocess_drift as `model` therefore remove
 
             # Backend specifics
-            if backend == 'tensorflow':
+            if has_tensorflow and isinstance(model, SupportedModels_tf):
```
ascillitoe, line 218: By the time we reach here, `model` has already been resolved. Hence we cannot read the `flavour` field in `ModelConfig` in order to determine whether it is a `tensorflow`, `pytorch` or `sklearn` model. Instead we infer this by examining the model type. The `has_tensorflow` check is to avoid calling `isinstance` when `SupportedModels_tf` is empty.

alibi_detect/saving/loading.py
```diff
@@ -216,14 +215,20 @@ def _load_preprocess_config(cfg: dict,
...
                 model = prep_model_and_emb_tf(model, emb)
                 kwargs.pop('device')
-            elif backend == 'pytorch':  # TODO - once optional deps implemented
+            elif has_pytorch and isinstance(model, SupportedModels_torch):
                 raise NotImplementedError('Loading preprocess_fn for PyTorch not yet supported.')
 #               device = cfg['device'] # TODO - device should be set already - check
 #               kwargs.update({'model': kwargs['model'].to(device)})  # TODO - need .to(device) here?
 #               kwargs.update({'device': device})
+            elif model is None:
```
ascillitoe, line 226: Handle the `model is None` case separately here, so that we don't miss the `kwargs.pop('device')` when `model is None` (it used to be done inside `prep_model_and_emb_tf` when `backend == 'tensorflow' even if `model is None`). When we add pytorch support a bit more logic might be needed here to avoid pop'ing `device` when the embedding is a pytorch model...

alibi_detect/saving/loading.py
```diff
@@ -235,25 +240,22 @@ def _load_preprocess_config(cfg: dict,
...
-    backend
-        The backend.
 
     Returns
     -------
     The loaded model.
     """
 
     # Load model
+    flavour = cfg['flavour']
```
ascillitoe, line 258: `ModelConfig` and `EmbeddingConfig` now contain a new field called `flavour`, which specs whether the model is `tensorflow`, `sklearn`, or `pytorch` model. Struggled to think of a good name for this and am open to suggestions!

alibi_detect/saving/loading.py
```diff
@@ -262,24 +264,24 @@ def _load_model_config(cfg: dict,
...
+        model = load_model_sk(src, load_dir='.')
     else:
-        raise NotImplementedError('Loading of non-tensorflow models not currently supported')
+        raise NotImplementedError('Loading of PyTorch models not currently supported')
 
     return model
 
 
-def _load_embedding_config(cfg: dict, backend: str) -> Callable:  # TODO: Could type return more tightly
+def _load_embedding_config(cfg: dict) -> Callable:  # TODO: Could type return more tightly
```
ascillitoe, line 277: No longer need to be passed `backend` as `flavour` is contained in the `EmbeddingConfig`. 

The ultimate goal here is to make these artefact config's self-contained. This will pave the way towards a more OOP approach! (@mauicv )

alibi_detect/saving/loading.py
```diff
@@ -492,15 +495,15 @@ def resolve_config(cfg: dict, config_dir: Optional[Path]) -> dict:
...
             elif key[-1] == 'embedding':
-                obj = _load_embedding_config(src, backend)
+                obj = _load_embedding_config(src)
             elif key[-1] == 'tokenizer':
                 obj = _load_tokenizer_config(src)
             elif key[-1] == 'optimizer':
                 obj = _load_optimizer_config(src, backend)
             elif key[-1] == 'preprocess_fn':
-                obj = _load_preprocess_config(src, backend)
+                obj = _load_preprocess_config(src)
```
ascillitoe, line 506: `backend` not passed to some of these functions since the artefact configs themselves now contain `flavour`.

alibi_detect/saving/loading.py
```diff
@@ -492,15 +495,15 @@ def resolve_config(cfg: dict, config_dir: Optional[Path]) -> dict:
...
             if key[-1] in ('model', 'proj'):
-                obj = _load_model_config(src, backend)
+                obj = _load_model_config(src)
             elif key[-1] == 'embedding':
-                obj = _load_embedding_config(src, backend)
+                obj = _load_embedding_config(src)
             elif key[-1] == 'tokenizer':
                 obj = _load_tokenizer_config(src)
             elif key[-1] == 'optimizer':
                 obj = _load_optimizer_config(src, backend)
```
ascillitoe, line 504: Some of these functions still have a `backend` arg since we still want some of these objects' backends to be constrained by the detector `backend`. For example, it would not make sense to allow a pytorch optimizer to be loaded with a tensorflow backend detector.

alibi_detect/saving/saving.py
```diff
@@ -123,9 +124,9 @@ def _save_detector_config(detector: ConfigurableDetector, filepath: Union[str, o
         File path to save serialized artefacts to.
     """
     # Get backend, input_shape and detector_name
-    backend = detector.meta.get('backend', 'tensorflow')
-    if backend != 'tensorflow':
-        raise NotImplementedError("Currently, saving is only supported with backend='tensorflow'.")
+    backend = detector.meta.get('backend', None)
+    if backend not in (None, 'tensorflow', 'sklearn'):
```
ascillitoe, line 128: `None` is now included since detectors such as `KSDrift` now have no `backend` in their config.

alibi_detect/saving/schemas.py
```diff
@@ -52,19 +49,22 @@ def coerce_int2list(value: int) -> List[int]:
         return value
 
 
-def validate_model(model: Callable, values: dict) -> Callable:
-    """Validator to check the model is compatible with the given backend"""
-    backend = values['backend']
-    if backend == 'tensorflow' and not isinstance(model, SupportedModels_tf):
-        raise ValueError('A TensorFlow backend is not available for this model')
-    elif backend == 'pytorch' and not isinstance(model, SupportedModels_torch):
-        raise ValueError('A PyTorch backend is not available for this model')
```
ascillitoe, line 61: I realised that these `ValueError`'s did not necessarily explain what we are actually validating. The validation was intended to check that the given `model` matched `backend`, but the error message suggested something else. We can reword if we decide to still validate the `model` type against `backend` (see next comment).

alibi_detect/saving/schemas.py
```diff
@@ -52,19 +49,22 @@ def coerce_int2list(value: int) -> List[int]:
...
-    elif backend == 'sklearn' and not isinstance(model, SupportedModels_sklearn):
-        raise ValueError('A sklearn backend is not available for this model')
-    return model
+class SupportedModelsType:
+    """
```
ascillitoe, line 53: Previously we typed `model` fields as:

```python
model: Optional[Callable] = None
```

and then validated them with:

```python
_validate_model = validator('model', allow_reuse=True, pre=True)(validate_model)
```

This did not work for sklearn models since `sklearn.base.BaseEstimator` is not a `Callable`. We could simple change to `model: Optional[Any] = None`. However I think a more elegent solution is to use a pydantic [custom data type](https://pydantic-docs.helpmanual.io/usage/types/#custom-data-types). This can then be used directly as:

```python
model: Optional[SupportedModelsType] = None
```

Note, that as of now the validation we are doing has changed. We are simply validating if the `model` type is one of the `SupportedModels`, which itself is conditional on the installed optional deps. We could add in more specific checks to validate `model` against `SupportModels_tf` etc depending on `backend`. (just thought if this and will look into now!)

alibi_detect/saving/schemas.py
```diff
@@ -97,8 +97,6 @@ class DetectorConfig(CustomBaseModel):
     """
     name: str
     "Name of the detector e.g. `MMDDrift`."
-    backend: Literal['tensorflow', 'pytorch', 'sklearn', 'keops'] = 'tensorflow'
```
ascillitoe, line 100: No `backend` in `DetectorConfig` as this is now specific to each detector.

alibi_detect/saving/tests/test_saving.py
```diff
@@ -76,213 +72,6 @@
 # TODO - future: Some of the fixtures can/should be moved elsewhere (i.e. if they can be recycled for use elsewhere)
 
 
-@fixture
```
ascillitoe, line 79: Moved to `models.py`.

alibi_detect/saving/tests/test_saving.py
```diff
@@ -1170,69 +994,14 @@ def test_save_preprocess_nlp(data, preprocess_fn, tmp_path, backend):
...
-        embedding_load = preprocess_fn_load.keywords['model'].encoder.layers[0]
-    assert isinstance(embedding_load.model, type(embedding.model))
-    assert embedding_load.emb_type == embedding.emb_type
-    assert embedding_load.hs_emb.keywords['layers'] == embedding.hs_emb.keywords['layers']
-
-
-@parametrize_with_cases("data", cases=ContinuousData.data_synthetic_nd, prefix='data_')
-@parametrize('model', [encoder_model])
-@parametrize('layer', [None, -1])
-def test_save_model(data, model, layer, backend, tmp_path):
```
ascillitoe, line 1186: Moved to `test_saving_tf.py`.

requirements/dev.txt
```diff
@@ -23,3 +23,4 @@ tox>=3.21.0, <4.0.0 # used to generate licence info via `make licenses`
 twine>3.2.0, <4.0.0  # 4.x causes deps clashes with testing/requirements.txt, as requires rich>=12.0.0 -> requires typing-extensions>=4.0.0 -> too high for spacy and thinc!
 packaging>=19.0, <22.0 # Used to check scipy version for CVMDrift test. Can be removed once python 3.6 support dropped (and scipy lower bound >=1.7.0).
 codecov>=2.0.15, <3.0.0
+xgboost>=1.3.2, <2.0.0 # Install for use in testing since we support serialization of xgboost models under the sklearn API
```
ascillitoe, line 26: Need to install this for the xgboost model defined [here](https://github.com/SeldonIO/alibi-detect/pull/642/files#diff-18d404d6b5ac8c3a8692660e2de7eb57a989b7b260f7f15aecf0ed8ce09b4fe9R155) and used in a test [here](https://github.com/SeldonIO/alibi-detect/pull/642/files#diff-349b14885d352b4861af40d39a1a767f3794394c030d01173832d7277ec36900R14).

alibi_detect/saving/schemas.py
ascillitoe: Note that the:

```python
raise ValueError('A TensorFlow backend is not available for this model')
```

Has been changed to:

```python
raise TypeError("`backend='tensorflow'` but the `model` doesn't appear to be a TensorFlow supported model.")
```

as I think the previous error message was misleading.

jklaise:
Very likely outside the scope of this PR, but is it true that we only support the `sklearn` interface for `xgboost` models where they can be used in the library (`ClassifierDrift` being the only place?) ?

Reason for asking is two-fold:
 - should we support `xgboost` native format for detectors like `ClassifierDrift`?
 - the native format seems to have good json-based serialization: https://xgboost.readthedocs.io/en/stable/tutorials/saving_model.html

alibi_detect/saving/tests/test_saving.py
```diff
@@ -406,12 +196,15 @@ def test_save_cvmdrift(data, preprocess_custom, tmp_path):
...
 @parametrize_with_cases("data", cases=ContinuousData, prefix='data_')
-def test_save_mmddrift(data, kernel, preprocess_custom, backend, tmp_path, seed):
+def test_save_mmddrift(data, kernel, preprocess_custom, backend, tmp_path, seed):  # noqa: F811
     """
     Test MMDDrift on continuous datasets, with UAE as preprocess_fn.
 
     Detector is saved and then loaded, with assertions checking that the reinstantiated detector is equivalent.
     """
+    if backend not in ('tensorflow', 'pytorch', 'keops'):
+        pytest.skip("Detector doesn't have this backend")
```
jklaise, line 206: Slightly confused about these statements, is this future-proofing if we had more backends? Also looking at the `backend` fixture it's only parametrized with `tensorflow` and `sklearn`, so how does it work for `keops`, `pytorch`?

jklaise (commented):
Mostly LGTM, I left some questions in places I'm unsure about.

alibi_detect/saving/_sklearn/loading.py
```diff
@@ -0,0 +1,29 @@
...
+    Load scikit-learn (or xgboost) model. Models are assumed to be a subclass of :class:`~sklearn.base.BaseEstimator`.
+    This includes xgboost models following the scikit-learn API
+    (see https://xgboost.readthedocs.io/en/latest/python/python_api.html#module-xgboost.sklearn).
+
+    Parameters
+    ----------
+    filepath
+        Saved model directory.
+    load_dir
+        Name of saved model folder within the filepath directory.
```
mauicv, line 22: Not quite sure of the difference between the args? Is `filepath` the saved detector directory and the `load_dir` the location within the detector directory? Why not just have `filepath`?

alibi_detect/saving/loading.py
mauicv: I vote for `flavorflav` :rofl:?

doc/source/overview/saving.md
```diff
@@ -127,13 +127,17 @@ documented below.
...
+### Scikit-learn
 
-%### scikit-learn
+Scikit-learn models are serialized using [joblib](https://joblib.readthedocs.io/en/latest/persistence.html).
+Any scikit-learn model that is a subclass of {py:class}`sklearn.base.BaseEstimator` is supported, including 
+[xgboost](https://xgboost.readthedocs.io/en/latest/python/python_api.html#module-xgboost.sklearn) models following 
+the scikit-learn API.
 
+
+%### PyTorch
```
mauicv, line 143: Is the % intentional?

alibi_detect/saving/loading.py
ascillitoe: Yeah, boyeee!

alibi_detect/saving/_tensorflow/loading.py
ascillitoe: I'm not sure we can use `SupportedModelsType` as is, since it contains if-statements conditional on the installed optional deps, so I don't think a static type checker (mypy) will like it? I wonder if we could do something clever by combining it with a `Protocol` (https://www.madelyneriksen.com/blog/validated-container-types-python-pydantic)? However, I've just realised that `UAE` and `HiddenOutput` etc are all subclasses of `tf.Keras.Model` so we could probably just simplify the type annotations anyway... 

Will open a follow-up PR for this!

alibi_detect/saving/_tensorflow/saving.py
```diff
@@ -78,7 +81,10 @@ def save_model_config(model: Callable,
     if model is not None:
         filepath = base_path.joinpath(local_path)
         save_model(model, filepath=filepath, save_dir='model')
-        cfg_model = {'src': local_path.joinpath('model')}
+        cfg_model = {
+            'flavour': 'tensorflow',
```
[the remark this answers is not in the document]
ascillitoe: Will open a follow-up PR for this.

alibi_detect/saving/loading.py
```diff
@@ -127,9 +129,9 @@ def _load_detector_config(filepath: Union[str, os.PathLike]) -> ConfigurableDete
     logger.info('Validated resolved config.')
 
     # Backend
-    backend = cfg.pop('backend')  # popping so that cfg left as kwargs + `name` when passed to _init_detector
-    if backend.lower() != 'tensorflow':
-        raise NotImplementedError('Loading detectors with PyTorch, sklearn or keops backend is not yet supported.')
+    backend = cfg.get('backend', None)
```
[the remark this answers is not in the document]
ascillitoe: It's no longer needed as pydantic now only populates the config with `backend` for detectors who have a `backend` kwarg (before we added `backend` to all detector configs as it also represented the "flavour" of preprocessing).

alibi_detect/saving/tests/test_saving.py
ascillitoe: Yes future proofing, but more in the sense of future-proofing in anticipation of serialization support being extended. 

This file (`test_saving.py`) is currently only parametrized with `backend = param_fixture("backend", ['tensorflow', 'sklearn'])` because we currently only support saving with `tensorflow` and `sklearn`. As we add support for more backends, we should add them to this parametrization. 

The `if backend not in ('tensorflow', 'pytorch', 'keops'):` is intended to skip the test if any other `backend` is passed to it (e.g. atm `backend=='sklearn'` is passed), because `MMDDrift` only has `'tensorflow'`, `'pytorch'` and `'keops'` backends. There are various ways we could do this but I thought the above logic was the most maintainable because there is a one-to-one mapping between the contents of the tuple and the actual backends supported by the detector.

doc/source/overview/saving.md
ascillitoe: Yes. It will be uncommented once we add pytorch serialization support.

alibi_detect/saving/_sklearn/loading.py
ascillitoe: Fair point. This is really only for consistency with the `_tensorflow.loading.load_model` function. It has these two separate kwarg's for use in the legacy `load_detector` here: https://github.com/SeldonIO/alibi-detect/blob/6fa17081735c11275dc7e551ddc672674e679b93/alibi_detect/saving/_tensorflow/loading.py#L275-L276.

Happy to remove now for sklearn if you think best?

ascillitoe: Removed in [c3fd225](https://github.com/SeldonIO/alibi-detect/pull/642/commits/c3fd2256c92bc2ff96b8d28dde946d90cde8cefa) :)

mauicv (approved):

PR #645, opened 2022-10-06
Experimental Outlier detectors

mauicv:
This PR is an experimental implementation of:

- knn outlier detectors
- ensemble outlier detectors
- deepsvdd outlier detectors
- serialization and deserialization of the above detectors

General notes on the implemnetation are [here](https://www.notion.so/Implementation-details-65267847104046f2b6676a1cdb7a22e0) notion page. This PR follows on from #643, #603, #601 and #596. Notes on these PRs are [here](https://www.notion.so/Outlier-Revamp-Experimental-Work-a16269c0b42c4dfe9d52613bf8661dba). Specific notes on the serialization changes relvenant to this PR are [here](https://www.notion.so/Outlier-Revamp-Experimental-Work-part-2-43cd3ff3acc04e1ea8d39bfa6c51e2e8).

**Note**: This is a POC and should not be merged!


alibi_detect/od/knn.py
```diff
@@ -0,0 +1,56 @@
...
+X_REF_FILENAME = 'x_ref.npy'
+
+backends = {
+    'pytorch': KNNTorch,
+    'keops': KNNKeops
+}
+
+@registry.register('KNN')
+class KNN(OutlierDetector, ConfigMixin):
+    CONFIG_PARAMS = ('k', 'kernel', 'aggregator', 'normaliser', 'backend')
```
jklaise, line 22: Wondering if we can mostly automate this by setting it in the constructor using `locals()` (with appropriate exclusion rules). Or perhaps this is good enough if we have appropriate tests to ensure everything we want to (and don't want to) serialize is done for every detector. To be fair, in either case we'll have to do some manual lists of arguments to include/exclude so having it at the top as a class attribute at least makes it fully visible.

jklaise: On further inspection I see that there is a standardized call `self_set_config(locals())` which then looks at `CONFIG_PARAMS` to decide what to do with each argument. I think this is a good approach overall.

alibi_detect/od/backends/torch/kernels.py
```diff
@@ -0,0 +1,126 @@
+import imp
```
arnaudvl, line 1: All this is to test as it is a copy of the `utils.torch` equivalent?

PR #650, opened 2022-10-12
Follow-up to sklearn serialization

ascillitoe:
Minor follow-up PR to https://github.com/SeldonIO/alibi-detect/pull/642.

## Summary
Contains 3 main changes, and 1 requiring discussion (see **Outstanding**):
1. The backend frameworks (`'tensorflow'`, `'pytorch'` etc) are set via an `Enum` to avoid typo's, as discussed in https://github.com/SeldonIO/alibi-detect/pull/642#discussion_r990044848. Note, I haven't updated the tests to use this...
2. The old `SupportedModels_tf` etc defined in `saving.schemas.py` has been moved to `utils._types`, since it is more to do with typing than schemas specifically. It has also been renamed to `supported_models` since the camel case implied that these were classes, when in fact they are simply tuples of types. 
3. We had missed something when making TensorFlow an optional dependency. We still had ` optimizer: Optional['tf.keras.optimizers.Optimizer']` in some of the detector pydantic models. I think this would cause an error if the serialized detector was loaded when tensorflow isn't installed, since pydantic does not play nicely with forward refs. I've added the `SupportedOptimizers` custom pydantic data type to handle this in the same way `model`'s are handled.

## Outstanding

I investigated https://github.com/SeldonIO/alibi-detect/pull/642#discussion_r990039862. The issue here is that `SupportedModels` or similar pydantic data types cannot be used as type hints, since they are populated dynamically (conditional on installed optional deps), and mypy is a static type checker. For a simple solution, I have added `model_types_tf` etc for now for use in type hints. I won't update the rest of the code to use these until we have decided if these are a suitable solution.

### DIFF
--- a/alibi_detect/cd/keops/learned_kernel.py
+++ b/alibi_detect/cd/keops/learned_kernel.py
@@ -10,6 +10,7 @@ from typing import Callable, Dict, List, Optional, Union, Tuple
 from alibi_detect.cd.base import BaseLearnedKernelDrift
 from alibi_detect.utils.pytorch import get_device, predict_batch
 from alibi_detect.utils.pytorch.data import TorchDataset
+from alibi_detect.utils.frameworks import Framework
 
 
 class LearnedKernelDriftKeops(BaseLearnedKernelDrift):
@@ -130,7 +131,7 @@ class LearnedKernelDriftKeops(BaseLearnedKernelDrift):
             input_shape=input_shape,
             data_type=data_type
         )
-        self.meta.update({'backend': 'keops'})
+        self.meta.update({'backend': Framework.KEOPS})
 
         # Set device, define model and training kwargs
         self.device = get_device(device)
--- a/alibi_detect/cd/keops/learned_kernel.py
+++ b/alibi_detect/cd/keops/learned_kernel.py
@@ -131,7 +131,7 @@ class LearnedKernelDriftKeops(BaseLearnedKernelDrift):
             input_shape=input_shape,
             data_type=data_type
         )
-        self.meta.update({'backend': Framework.KEOPS})
+        self.meta.update({'backend': Framework.KEOPS.value})
 
         # Set device, define model and training kwargs
         self.device = get_device(device)
--- a/alibi_detect/saving/_sklearn/saving.py
+++ b/alibi_detect/saving/_sklearn/saving.py
@@ -35,7 +35,7 @@ def save_model_config(model: BaseEstimator,
     filepath = base_path.joinpath(local_path)
     save_model(model, filepath=filepath, save_dir='model')
     cfg_model = {
-        'flavour': Framework.SKLEARN,
+        'flavour': Framework.SKLEARN.value,
         'src': local_path.joinpath('model')
     }
     return cfg_model
--- a/alibi_detect/saving/_sklearn/saving.py
+++ b/alibi_detect/saving/_sklearn/saving.py
@@ -2,9 +2,10 @@ import logging
 import os
 from pathlib import Path
 from typing import Union
-
 import joblib
 from sklearn.base import BaseEstimator
+
+from alibi_detect.utils.frameworks import Framework
 
 logger = logging.getLogger(__name__)
 
@@ -34,7 +35,7 @@ def save_model_config(model: BaseEstimator,
     filepath = base_path.joinpath(local_path)
     save_model(model, filepath=filepath, save_dir='model')
     cfg_model = {
-        'flavour': 'sklearn',
+        'flavour': Framework.SKLEARN,
         'src': local_path.joinpath('model')
     }
     return cfg_model
--- a/alibi_detect/saving/schemas.py
+++ b/alibi_detect/saving/schemas.py
@@ -141,7 +141,7 @@ class ModelConfig(CustomBaseModel):
         src = "model/"
         layer = -1
     """
-    flavour: Literal[Framework.TENSORFLOW, Framework.PYTORCH, Framework.SKLEARN]
+    flavour: Literal['tensorflow', 'pytorch', 'sklearn']
     """
     Whether the model is a `tensorflow`, `pytorch` or `sklearn` model. XGBoost models following the scikit-learn API
     are also included under `sklearn`.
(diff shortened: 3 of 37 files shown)


ascillitoe:
@jklaise @mauicv this is WIP pending discussion of the item in **Outstanding**.

alibi_detect/cd/keops/learned_kernel.py
```diff
@@ -130,7 +131,7 @@ def __init__(
             input_shape=input_shape,
             data_type=data_type
         )
-        self.meta.update({'backend': 'keops'})
+        self.meta.update({'backend': Framework.KEOPS})
```
jklaise, line 134: Even though for `str enums` in a python interpreter `Framework.KEOPS == 'keops'` is `True`, I believe for user facing string values we need to do `Framework.KEOPS.value`...

jklaise: That being said `json.dumps({'backend': Framework.KEOPS})` does the right thing, so perhaps this is mostly a question of representing the metadata visually in the Python interpreter.

mauicv (approved):
LGTM!

alibi_detect/saving/_sklearn/saving.py
```diff
@@ -34,7 +35,7 @@ def save_model_config(model: BaseEstimator,
     filepath = base_path.joinpath(local_path)
     save_model(model, filepath=filepath, save_dir='model')
     cfg_model = {
-        'flavour': 'sklearn',
+        'flavour': Framework.SKLEARN.value,
```
jklaise, line 38: Inconsistency here where everywhere else `Framework.SKLEARN` was used, does it matter?

alibi_detect/saving/schemas.py
```diff
@@ -13,31 +13,16 @@
     For detector pydantic models, the fields match the corresponding detector's args/kwargs. Refer to the
     detector's api docs for a full description of each arg/kwarg.
 """
-
+from __future__ import annotations
```
jklaise, line 16: What was the reason for this? Could be potentially breaking?

alibi_detect/utils/_types.py
```diff
@@ -51,3 +53,33 @@ def _validate(cls: Type, val: Any, field: ModelField) -> np.ndarray:
...
+supported_models_sklearn = (BaseEstimator, )  # type: ignore
+supported_models_all = supported_models_tf + supported_models_torch + supported_models_sklearn
+supported_optimizers_all = supported_optimizers_tf + supported_optimizers_torch
+
+# 2. Type unions
+model_types_tf = Type['tf.keras.Model']
+model_types_torch = Union['torch.nn.Module', 'torch.nn.Sequential']
+model_types_sklearn = Type[BaseEstimator]  # no ForwardRef since sklearn a core dep
+optimizer_types_tf = Type['tf.keras.optimizers.Optimizer']
+optimizer_types_torch = Union['torch.optim.Optimizer']
```
jklaise, line 82: `Union` -> `Type`?

jklaise: Btw what's the difference between `Type[SomeClass]` and just `SomeClass` when used in type hints?

alibi_detect/saving/schemas.py
```diff
@@ -49,7 +34,7 @@ def coerce_int2list(value: int) -> List[int]:
         return value
 
 
-class SupportedModelsType:
+class SupportedModels:
```
jklaise, line 37: Nitpick: should this (and the optimizer one) be singular as it defines a `SupportedModel` type?

jklaise (commented):
Left some questions, overall looks good!

alibi_detect/saving/schemas.py
ascillitoe: Ooops thanks, I was experimenting and forgot to remove this. Deffo shouldn't be here!

alibi_detect/cd/keops/learned_kernel.py
ascillitoe: I think you're right, this was actually why I changed to `Framework.SKLEARN.value` in https://github.com/SeldonIO/alibi-detect/pull/650#discussion_r994429416, as the `Framework.SKLEARN` was causing the pydantic validation to fail. I (think) I agree that anything public facing should have a `.value`.

alibi_detect/utils/_types.py
ascillitoe: Without `Type` I get the following mypy error when `model_types_tf` is added to `model_types_all`:

```
error: Variable "alibi_detect.utils._types.model_types_tf" is not valid as a type  [valid-type]
```

To be honest I'm wondering if we should just remove lines 77 onwards for now, since we are not using them elsewhere in the code yet anyway. I've also just realised that `models_types_all` might no actually work properly anyway, since we are still defining this at runtime hence mixing dynamic and static typing again.

jklaise:
Re the outstanding stuff, I don't know of any better way to handle it. Perhaps there is some info out there on a harmonious use of `pydantic` models and static type checkers, but I would say low priority.

ascillitoe:
> Re the outstanding stuff, I don't know of any better way to handle it. Perhaps there is some info out there on a harmonious use of `pydantic` models and static type checkers, but I would say low priority.

Yeh I've done some research but haven't found much. I had hoped to define some custom model types that could be used both for static type checking (mypy) and runtime type checking (pydantic, and `isinstance`). I haven't found much that would work though, except for `Protocol`'s, but this would involve us trying to define `Protocol`'s to "duck type" tf and torch models, which seems dodgy...

In the latest commit I've removed the `models_types_all` etc entirely. I'm not sure its worth complicating matters with these right now, since mypy isn't checking tf, torch or sklearn types anyway (due to lack of stubs...).

alibi_detect/utils/_types.py
ascillitoe: Removed (for now) in e4ed219fb6b8d559eac503bbcd0200d78301e811.

alibi_detect/saving/schemas.py
ascillitoe: Removed in dc686f35b1120ede44268106ae00be4c66bbb6df.

alibi_detect/cd/keops/learned_kernel.py
ascillitoe: Changed to all `self.meta` updates to use `.value` in 44a7200.

alibi_detect/saving/_sklearn/saving.py
ascillitoe: See https://github.com/SeldonIO/alibi-detect/pull/650#discussion_r994464274.

jklaise (approved):

Issue #484, opened 2022-04-14
Some public classes and functions are missing docstrings


ascillitoe:
After the change to `sphinx-autoapi` (https://github.com/SeldonIO/alibi-detect/pull/482), there are some new objects showing up in the api docs which are undesired, for example the `logger`:

![image](https://user-images.githubusercontent.com/32061685/163418971-33db606f-720a-40f5-b873-bfb2c565e1e7.png)

This is a symptom of the `undoc-members` setting, which means even objects without docstrings are documented. Since it is bad practice to have undocumented public members in any case, we should ensure all public members have docstrings (e.g. see [alibi_detect.cd.tensorflow.UAE](https://docs.seldon.io/projects/alibi-detect/en/latest/api/alibi_detect/cd/tensorflow/index.html?highlight=UAE#alibi_detect.cd.tensorflow.UAE)). Once this is done we can turn off `undoc-members`, giving us more control over what is documented.

Issue #463, opened 2022-03-16
Type detector kwarg's as Literal where relevant


ascillitoe:
Some detector kwarg's are "magic strings" with a limited set of options, e.g. `correction` and `alternative` is the univariate detectors. We should update these kwarg's to type them as `Literal`. For example, in `ksdrift.py`:

```
            correction: str = 'bonferroni',
            alternative: str = 'two-sided',
```

would become:

```
            correction: str = Literal['bonferroni', 'fdr']
            alternative: str = Literal['two-sided', 'less', 'greater']
```

The same changes should be made in the pydantic schema's once the config PR is merged (#389), in order to tighten the validation of these kwarg's.

ascillitoe:
The same should also be done for other public facing classes, such as `DeepKernel`.

PR #656, opened 2022-10-19
Support for serializing detectors with PyTorch backends and/or models 

### LINKED ISSUE #527: Potential inconsistency in cd_text_imdb.ipynb
Hello,

I have a question regarding the consistency of the MMD Tensorflow detector and the MMD PyTorch detector in the [`cd_text_imdb.ipynb` Jupyter Notebook](https://github.com/SeldonIO/alibi-detect/blob/cb512067bc8206816eeddff37f004ef492ccd72c/doc/source/examples/cd_text_imdb.ipynb). In my opinion, the preprocessing function passed for the class instantiation of `alibi_detect.cd.MMDDrift` (cf. `preprocessing_fn` argument) is not identical across the two frameworks (Tensorflow and PyTorch, respectively). `preprocess_fn` is defined as follows:
- **Tensorflow**: 
```
preprocess_fn = partial(preprocess_drift, model=uae, tokenizer=tokenizer, 
                        max_len=max_len, batch_size=32)
```
with 
```
uae = UAE(input_layer=embedding, shape=shape, enc_dim=enc_dim)
```
such that `embedding` represents the embeddings derived from a transformer model in Tensorflow (cf. `alibi_detect.models.tensorflow.TransformerEmbedding` ) and `UAE` is the Tensorflow implementation of an untrained autoencoder (cf. [`alibi_detect.cd.tensorflow.preprocess`](https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/cd/tensorflow/preprocess.py)). If I am not mistaken, there is no UAE implemented in PyTorch - I'll come back to this later.

- **PyTorch**:
```
preprocess_fn = partial(preprocess_drift, model=model, tokenizer=tokenizer, 
                        max_len=max_len, batch_size=32, device=device)
```
with 
```
model = nn.Sequential(
    embedding_pt,
    nn.Linear(768, 256),
    nn.ReLU(),
    nn.Linear(256, enc_dim)
).to(device).eval()
```
such that `embedding_pt` represents the embeddings derived from a transformer model in PyTorch (cf. `alibi_detect.models.pytorch.TransformerEmbedding`).

I think that the main difference between the **Tensorflow** and **PyTorch** snippets comes from the respective definitions of the `model` argument. Since the UAE class does not exist (yet) in `alibi_detect.cd.pytorch`, the **PyTorch** part requires us to define a dimensionality reduction step. The dimension of the embeddings is reduced by:
(i) adding a dense layer on top of BERT embeddings (input dimension: 768; output dimension: 256)
(ii) applying a ReLU activation function, and 
(iii) adding another dense layer (input dimension: 256; output dimension: `enc_dim`). 

Although one reads that the detector with the PyTorch backend is identical to the detector with the Tensorflow backend (first sentence at the beginning of the [`MMD PyTorch detector > Initialize` subsection](https://docs.seldon.io/projects/alibi-detect/en/stable/examples/cd_text_imdb.html#id4)), the dimensionality reduction step in **PyTorch** is different from that in **Tensorflow**, _i.e._, the `uae` object (cf. implementation of the MMD detector with the Tensorflow backend) does not reduce the dimensionality of BERT embeddings in the same way as the `model` object used in PyTorch. Indeed, `uae` is an instance of the `UAE` class and the associated constructor indicates that we will use a `self.encoder` with 
(i) a dense layer (input dimension: 768; output dimension: 522, since 522 = 32 + 2 x ( int([768 - 32]/3) ) = 32 + 2 x 245)
(ii) a ReLU activation function
(iii) a dense layer (input dimension: 522; output dimension: 277, since 277 = 32 + 1 x ( int([768 - 32]/3) ) = 32 + 1 x 245)
(iv) a ReLU activation function
(v) a dense layer (input dimension: 277; output dimension: 32)
The above MLP is [defined in the constructor of the `_Encoder` class](https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/cd/tensorflow/preprocess.py#L24-L30).

Based on what I described here, and assuming that there is no misunderstanding on my part, is it intended that the dimensionality reduction steps have not been defined in a unique way when using the Tensorflow and PyTorch backends respectively? 
Many thanks for your attention!
(Tagging @ascillitoe since he took over the latest issue related to cd_text_imdb.ipynb (issue #437) 😉 )

ascillitoe:
This PR implements save/load support for PyTorch backends and models.

The primary change addition is a `saving._pytorch` subpackage, containing save/load support for pytorch models, optimizers, kernels etc. For consistency with TensorFlow in the way `preprocess_drift` functions are serialized, a pytorch equivalent of `cd.preprocess.tensorflow.UAE` has also been implemented (this will also resolve #527 once examples are updated to use it).

### pytorch model saving

To avoid the need to provide torch model classes at load time (either via kwargs to `load_detector` or our registry), the currently implementation saves the entire pytorch model (see [pytorch docs](https://pytorch.org/tutorials/beginner/saving_loading_models.html#save-load-entire-model)). When `torch.save(...)` is called with the default kwargs, `pickle` is used for this, which would result in _"The disadvantage of this approach is that the serialized data is bound to the specific classes and the exact directory structure used when the model is saved."_. Instead, the current implementation uses `dill` (i.e. `torch.save(..., pickle_module=dill)`), which seems to fix the above issue. [Link to experiments on this](https://www.notion.so/Serialize-torch-backends-preprocessing-f34d1c7cc0bd4f68b0e80b89dff7f73d#830bffb3d4f24877bbda234225420053).

### TODO

- [x] Investigate difference in number of parameters in first layer of the pytorch `UAE` versus the tensorflow one. - _This was me mistakenly building the tensorflow `UAE` model with `input_shape = (256, )` instead of `input_shape = (1, 256)`._
- [x] Related to above, decide on using `Linear` vs `LazyLinear` for `UAE` (see [comment](https://github.com/SeldonIO/alibi-detect/pull/656/files#discussion_r1000612590) and [comment](https://github.com/SeldonIO/alibi-detect/pull/656/files#discussion_r1000619584)). _Can untick but IMO sticking to `Linear` is fine_.
- [x] Fix optional deps test problem. (thanks @mauicv )
- [x] Check if there are any relevant limitations to the extent of `torch` save/load support via `dill`.


alibi_detect/cd/mmd_online.py
```diff
@@ -150,3 +150,16 @@ def score(self, x_t: Union[np.ndarray, Any]) -> float:
         Squared MMD estimate between reference window and test window.
         """
         return self._detector.score(x_t)
+
+    def get_config(self) -> dict:  # Needed due to self.x_ref being a torch.Tensor when backend='pytorch'
```
ascillitoe, line 154: Need to override `get_config` in order to convert `self.x_ref` back to `np.ndarray` (it is a `torch.Tensor` in the pytorch version of `MMDDriftOnline`).

alibi_detect/cd/pytorch/preprocess.py
```diff
@@ -7,6 +7,63 @@
...
+            raise ValueError('Need to provide either `enc_dim` and `step_dim` or a '
+                             'nn.Module `mlp`')
+
+    def forward(self, x: Union[np.ndarray, torch.Tensor, Dict[str, torch.Tensor]]) -> torch.Tensor:
+        if isinstance(self.input_layer, nn.Module):
+            x = self.input_layer(x)
+        return self.mlp(x)
+
+
+class UAE(nn.Module):
```
ascillitoe, line 42: Colab notebook for prototyping:
https://gist.github.com/ascillitoe/0c967eeac59a941a17f4c6c258b1ef81

alibi_detect/cd/pytorch/preprocess.py
```diff
@@ -7,6 +7,63 @@
                                                    predict_batch_transformer)
 
 
+class _Encoder(nn.Module):
+    def __init__(
+            self,
+            input_layer: Union[nn.Module, int],
+            mlp: Optional[nn.Module] = None,
+            input_dim: Optional[int] = None,
```
ascillitoe, line 15: Additional `input_dim` kwarg (compared to tensorflow version) because unlike keras `Dense` layers, pytorch `Linear` layers need input shape to be given. 

TODO: Explore using `LazyLinear` for consistency with tensorflow implementation...

alibi_detect/saving/_pytorch/loading.py
```diff
@@ -0,0 +1,143 @@
...
+
+    Returns
+    -------
+    The loaded optimizer class.
+    """
+    class_name = cfg.get('class_name', None)
+    try:
+        return getattr(import_module('torch.optim'), class_name)
+    except AttributeError:
+        raise ValueError(f"{class_name} is not a recognised optimizer in `torch.optim`.")
```
ascillitoe, line 120: The pytorch versions of `ClassifierDrift` etc expect `optimizer` to be a class, which is then instantiated internally. Therefore, `load_optimizer` here simply imports the class without instantiating.

alibi_detect/saving/_pytorch/loading.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    filepath = Path(filepath).joinpath('model.pt')
+    model = torch.load(filepath, pickle_module=dill)
+    # Optionally extract hidden layer
+    if isinstance(layer, int):
+        model = HiddenOutput(model, layer=layer)
+    return model
+
+
+def prep_model_and_emb(model: nn.Module, emb: Optional[TransformerEmbedding]) -> nn.Module:
```
ascillitoe, line 45: Pretty much same as tensorflow equivalent.

alibi_detect/saving/_pytorch/loading.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    class_name = cfg.get('class_name', None)
+    try:
+        return getattr(import_module('torch.optim'), class_name)
+    except AttributeError:
+        raise ValueError(f"{class_name} is not a recognised optimizer in `torch.optim`.")
+
+
+def load_embedding(src: str, embedding_type, layers) -> TransformerEmbedding:
+    """
```
ascillitoe, line 124: Pretty much same as tensorflow equivalent.

alibi_detect/saving/_pytorch/saving.py
```diff
@@ -0,0 +1,134 @@
...
+    cfg_model = None  # type: Optional[Dict[str, Any]]
+    cfg_embed = None  # type: Optional[Dict[str, Any]]
+    if isinstance(model, UAE):
+        layers = list(model.encoder.children())
+        if isinstance(layers[0], TransformerEmbedding):  # if UAE contains embedding and encoder
+            # embedding
+            embed = layers[0]
+            cfg_embed = save_embedding_config(embed, base_path, local_path.joinpath('embedding'))
+            # preprocessing encoder
+            model = layers[1]
```
ascillitoe, line 49: When `model` is a `UAE`, we unpick the `TransformerEmbedding` and encoder model, and save both separately. This "unpicking" is much more simple than the pytorch equivalent because pytorch models don't need to be built/called before saving (since they already know their dimensions etc). If we replaced `Linear` in `UAE` with `LazyLinear` we would have to do the same as for tensorflow...

alibi_detect/saving/_pytorch/saving.py
```diff
@@ -0,0 +1,134 @@
...
+        else:  # If UAE is simply an encoder
+            model = model.encoder
```
ascillitoe, line 51: The rest is pretty much the same as tensorflow version...

alibi_detect/saving/_pytorch/saving.py
```diff
@@ -0,0 +1,134 @@
...
+    # save model
+    model_path = model_path.joinpath('model.pt')
+
+    if isinstance(model, (nn.Module, nn.Sequential)):
+        torch.save(model, model_path, pickle_module=dill)
+    else:
+        raise ValueError('The extracted model to save is not a `nn.Module` or `nn.Sequential`. Cannot save.')
+
+
+def save_embedding_config(embed: TransformerEmbedding,
```
ascillitoe, line 102: Also pretty much same as tensorflow version.

alibi_detect/saving/_tensorflow/loading.py
```diff
@@ -128,10 +129,9 @@ def load_kernel_config(cfg: dict) -> Callable:
     return kernel
 
 
-def load_optimizer(cfg: dict) -> tf.keras.optimizers.Optimizer:
+def load_optimizer(cfg: dict) -> Union[Type[tf.keras.optimizers.Optimizer], tf.keras.optimizers.Optimizer]:
```
ascillitoe, line 132: `Type[...]` because `optimizer` may be passed to `ClassifierDrift` etc either as an instantiated object or a class. For example, see:

https://github.com/SeldonIO/alibi-detect/blob/6df9bd4b0fb6b27a1df41f0cdae76c804dc3cd55/alibi_detect/cd/tensorflow/classifier.py#L141

As an aside, we should update type hints to show this:
https://github.com/SeldonIO/alibi-detect/blob/6df9bd4b0fb6b27a1df41f0cdae76c804dc3cd55/alibi_detect/cd/tensorflow/classifier.py#L34

ascillitoe: Serialization logic wrt to above is as follows:

1. If the `optimizer` config is a tensorflow optimizer config, e.g. 
   ```toml
        [optimizer]
        class_name = "Adam"

        [optimizer.config]
        name = "Adam"
        learning_rate = 0.001
        decay = 0.0
   ```
   
   This is deserialized into a `tf.keras.optimizers.Optimizer` object.

2. If the `optimizer` config only specifies the class name e.g. 
      ```toml
        [optimizer]
        class_name = "Adam"
   ```
  Only the class is loaded (i.e. it isn't instantiated). **This approach is the only approach available in the torch implementation**, since torch optimizers cannot be instantiated without providing model parameters.

alibi_detect/saving/_tensorflow/saving.py
```diff
@@ -158,6 +158,21 @@ def save_embedding_config(embed: TransformerEmbedding,
     return cfg_embed
 
 
+def save_optimizer_config(optimizer: tf.keras.optimizers.Optimizer):
```
ascillitoe, line 161: This was missing before!

alibi_detect/saving/loading.py
```diff
@@ -217,19 +219,22 @@ def _load_preprocess_config(cfg: dict) -> Optional[Callable]:
...
                 model = prep_model_and_emb_tf(model, emb)
-                kwargs.pop('device')
             elif has_pytorch and isinstance(model, supported_models_torch):
-                raise NotImplementedError('Loading preprocess_fn for PyTorch not yet supported.')
-#               device = cfg['device'] # TODO - device should be set already - check
-#               kwargs.update({'model': kwargs['model'].to(device)})  # TODO - need .to(device) here?
-#               kwargs.update({'device': device})
+                model = prep_model_and_emb_pt(model, emb)
             elif model is None:
-                kwargs.pop('device')
```
ascillitoe, line 227: Cleaner to pop device below.

alibi_detect/saving/registry.py
```diff
@@ -67,4 +67,4 @@ def my_function(x: np.ndarray) -> np.ndarray:
     registry.register('utils.pytorch.kernels.GaussianRBF', func=GaussianRBF_torch)
     registry.register('utils.pytorch.kernels.sigma_median', func=sigma_median_torch)
     registry.register('cd.pytorch.context_aware._sigma_median_diag', func=_sigma_median_diag_torch)
-    registry.register('utils.pytorch.preprocess.preprocess_drift', func=preprocess_drift_torch)
```
ascillitoe, line 70: This was wrong before.

alibi_detect/saving/saving.py
```diff
@@ -178,6 +184,11 @@ def _save_detector_config(detector: ConfigurableDetector, filepath: Union[str, o
         model_cfg, _ = _save_model_config(model, base_path=filepath, input_shape=cfg['input_shape'])
         cfg['model'] = model_cfg
 
+    # Serialize optimizer
+    optimizer = cfg.get('optimizer', None)
```
ascillitoe, line 188: Was missing!

alibi_detect/saving/saving.py
```diff
@@ -503,3 +516,22 @@ def _save_kernel_config(kernel: Callable,
...
+    ----------
+    optimizer
+        The optimizer to save.
+
+    Returns
+    -------
+    Optimizer config dict.
+    """
+    if isinstance(optimizer, type):
+        return {'class_name': optimizer.__name__}
```
ascillitoe, line 535: Simply saves the class name.

alibi_detect/saving/saving.py
```diff
@@ -503,3 +516,22 @@ def _save_kernel_config(kernel: Callable,
...
+    else:
+        return save_optimizer_config_tf(optimizer)
```
ascillitoe, line 537: If `optimizer` is an object, it must be a tensorflow optimizer, serialize it into a tensorflow conifg dict.

alibi_detect/saving/schemas.py
```diff
@@ -19,18 +19,10 @@
...
 
 from alibi_detect.utils.frameworks import Framework
-from alibi_detect.utils._types import (Literal, NDArray, supported_models_all, supported_models_tf,
+from alibi_detect.utils._types import (Literal, supported_models_all, supported_models_tf,
                                        supported_models_sklearn, supported_models_torch, supported_optimizers_tf,
                                        supported_optimizers_torch, supported_optimizers_all)
-
-
-# Custom validators (defined here for reuse in multiple pydantic models)
-def coerce_int2list(value: int) -> List[int]:
```
ascillitoe, line 28: Moved to a new `validators.py` file, since we are beginning to add lots of these...

alibi_detect/saving/schemas.py
```diff
@@ -353,7 +349,11 @@ class KernelConfig(CustomBaseModelWithKwargs):
     "A string referencing a filepath to a serialized kernel in `.dill` format, or an object registry reference."
 
     # Below kwargs are only passed if kernel == @GaussianRBF
-    sigma: Optional[NDArray[np.float32]] = None
+    flavour: Literal['tensorflow', 'pytorch']
```
ascillitoe, line 352: Validation of this is essentially a 2-stage process. It is validated against the `Literal`, before being passed to `validate_framework` below to check against installed optional deps.

alibi_detect/saving/schemas.py
```diff
@@ -367,6 +367,9 @@ class KernelConfig(CustomBaseModelWithKwargs):
     should match :py:func:`~alibi_detect.utils.tensorflow.kernels.sigma_median`. If `None`, it is set to
     :func:`~alibi_detect.utils.tensorflow.kernels.sigma_median`.
     """
+    # Validators
+    _validate_flavour = validator('flavour', allow_reuse=True, pre=False)(validate_framework)
+    _coerce_sigma2tensor = validator('sigma', allow_reuse=True, pre=False)(coerce_2tensor)
```
ascillitoe, line 372: Almost all of our public API expects `list`'s or `np.ndarray`'s. However, a peculiarity of our `GaussianRBF` kernels is that they are public, but expect `sigma` to be a `tf.Tensor` or `torch.Tensor`. Previously we were coercing `sigma` into a `np.ndarray` (via the `NDArray` pydantic type). It wasn't strictly correct to pass this to the tensorflow `GaussianRBF`, however we got away with it since `sigma` was coerced to a `tf.Tensor` internally.  The pytorch `GaussianRBF` does not do this, hence passing `sigma` as an `np.ndarray` fails. Therefore, the `coerce_2tensor` validator has been added to coerce `sigma` to the correct `Tensor` depending on the value of `flavour`. Depending on @Srceh's kernel updates we can retain this functionality or remove as needed.

alibi_detect/saving/tests/models.py
```diff
@@ -64,12 +70,15 @@ def encoder_dropout_model(backend, current_cases):
                [
                    tf.keras.layers.InputLayer(input_shape=(input_dim,)),
                    tf.keras.layers.Dense(5, activation=tf.nn.relu),
-                   tf.keras.layers.Dropout(0.5),
+                   tf.keras.layers.Dropout(0.0),  # 0.0 to ensure determinism
```
ascillitoe, line 73: Unlikely to give a sensible model for `RegressorUncertaintyDrift`, but if > 0.0 detector predictions are not deterministic. OK for testing serialization I think...

alibi_detect/saving/tests/models.py
```diff
@@ -93,18 +102,43 @@ def kernel(request, backend):
...
+            if sigma is not None and not isinstance(sigma, torch.Tensor):
+                sigma = torch.tensor(sigma)
+            kernel = GaussianRBF_pt(sigma=sigma, **kernel_cfg)
         else:
             pytest.skip('`kernel` only implemented for tensorflow and pytorch.')
     return kernel
 
 
+@fixture
+def optimizer(request, backend):
```
ascillitoe, line 123: This has been added so that we can test the full e2e save/load of an `optimizer` kwarg. This is only used in the `test_save_classifier` for now.

alibi_detect/saving/validators.py
```diff
@@ -0,0 +1,84 @@
...
+
+from alibi_detect.utils.frameworks import has_tensorflow, has_pytorch, has_keops, Framework
+
+if has_tensorflow:
+    import tensorflow as tf
+if has_pytorch:
+    import torch
+
+
+def coerce_int2list(value: int) -> List[int]:
```
ascillitoe, line 15: Moved from `schemas.py`.

alibi_detect/saving/validators.py
```diff
@@ -0,0 +1,84 @@
...
+        return framework
+    elif framework == Framework.SKLEARN:  # sklearn is a core dep
+        return framework
+    else:
+        raise ImportError(f"`{field.name} = '{framework}'` not possible since {framework} is not installed.")
+
+
+# NumPy NDArray pydantic validator type
+# The code below is adapted from https://github.com/cheind/pydantic-numpy.
+T = TypeVar("T", bound=np.generic)
```
ascillitoe, line 36: Moved from `utils._types`.

alibi_detect/utils/_types.py
```diff
@@ -15,57 +11,18 @@
...
+supported_optimizers_tf, supported_optimizers_torch = None, None
 if has_tensorflow:
     import tensorflow as tf
-    supported_models_tf = (tf.keras.Model, )  # type: ignore
-    supported_optimizers_tf = (tf.keras.optimizers.Optimizer, )  # type: ignore
+    supported_models_tf = tf.keras.Model
+    supported_optimizers_tf = (tf.keras.optimizers.Optimizer, type)
 if has_pytorch:
     import torch
-    supported_models_torch = (torch.nn.Module, torch.nn.Sequential)  # type: ignore
```
ascillitoe, line 67: `nn.Sequential` is a subclass of `nn.Module` hence `nn.Sequential` removed.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -89,6 +91,29 @@ def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch
         kernel_mat = torch.exp(- torch.cat([(g * dist)[None, :, :] for g in gamma], dim=0))  # [Ns, Nx, Ny]
         return kernel_mat.mean(dim=0)  # [Nx, Ny]
 
+    def get_config(self) -> dict:
+        """
+        Returns a serializable config dict (excluding the input_sigma_fn, which is serialized in alibi_detect.saving).
+        """
+        cfg = self.config.copy()
+        if isinstance(cfg['sigma'], torch.Tensor):
+            cfg['sigma'] = cfg['sigma'].detach().cpu().numpy().tolist()
```
ascillitoe, line 100: Need to convert `sigma` from `torch.Tensor` back to `list`.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -89,6 +91,29 @@ def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch
...
+        cfg.update({'flavour': Framework.PYTORCH.value})
```
ascillitoe, line 101: We add in `flavour` to config since it is needed by the pydantic validation in order to coerce `sigma` to the correct type.

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -89,6 +91,29 @@ def forward(self, x: Union[np.ndarray, torch.Tensor], y: Union[np.ndarray, torch
...
+    def from_config(cls, config):
+        """
+        Instantiates a kernel from a config dictionary.
+
+        Parameters
+        ----------
+        config
+            A kernel config dictionary.
+        """
+        config.pop('flavour')
```
ascillitoe, line 114: `flavour` must be removed since isn't a kwarg of `GaussianRBF`

alibi_detect/tests/test_dep_management.py
```diff
@@ -205,13 +206,34 @@ def test_saving_tf_dependencies(opt_dep):
...
     from alibi_detect.saving import _tensorflow as tf_saving
     check_correct_dependencies(tf_saving, dependency_map, opt_dep)
 
 
+def test_saving_torch_dependencies(opt_dep):
+    """Tests that the alibi_detect.saving._pytorch module correctly protects against uninstalled optional
+    dependencies.
+    """
+
+    dependency_map = defaultdict(lambda: ['default'])
```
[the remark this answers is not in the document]
mauicv: the `opt_dep` parameter passed to the tests that tells it which optional dependency bucket is installed, in the case of PyTorch, is `torch` but in the dependencies, for each imported object you call it `pytorch`. tbh, we should catch both cases and we should also log these tests better. I'll make PR to do this.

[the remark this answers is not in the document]
mauicv: Basically:

```py
('load_embedding_pt', ['pytorch']),
('load_kernel_config_pt', ['pytorch']),
...
```

should be

```py
('load_embedding_pt', ['torch']),
('load_kernel_config_pt', ['torch']),
...
```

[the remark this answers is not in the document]
ascillitoe: aaaah got you! Its probs worth me changing these to `torch` anyway I guess? (for consistency)

alibi_detect/cd/pytorch/preprocess.py
ascillitoe: IMO we should stick to `Linear`, since it simplifies `save_model_config`. Compare the [pytorch implementation](https://github.com/ascillitoe/alibi-detect/blob/cf3a9e153d791fe6aa81db1b58f0d46c7269c473/alibi_detect/saving/_pytorch/saving.py#L42-L51) to to the [tensorflow one](https://github.com/SeldonIO/alibi-detect/blob/6df9bd4b0fb6b27a1df41f0cdae76c804dc3cd55/alibi_detect/saving/_tensorflow/saving.py#L55-L69).

Requiring `input_dim` to be passed to `_Encoder` also doesn't significantly complicate the public api, since `_Encoder` is private, and `shape` is already passed to `UAE`.

ascillitoe:
@arnaudvl I've requested a review from you mainly to discuss the new pytorch `UAE`, since you wrote the tensorflow one. Would also be great to get your thoughts/ideas on any pytorch models you think might break the dill-based model saving (@jklaise, @mauicv - input welcome on these topics too of course!).

alibi_detect/saving/validators.py
```diff
@@ -0,0 +1,84 @@
...
+def coerce_int2list(value: int) -> List[int]:
+    """Validator to coerce int to list (pydantic doesn't do this by default)."""
+    if not isinstance(value, list):
+        return [value]
+    else:
+        return value
+
+
+# Framework validator (validates `flavour` and `backend` fields)
+def validate_framework(framework: str, field: ModelField) -> str:
```
jklaise, line 24: Is `ModelField` something to do with code completion?

alibi_detect/saving/validators.py
```diff
@@ -0,0 +1,84 @@
...
+
+def _coerce2NDArray(cls: Type, val: Any, field: ModelField) -> np.ndarray:
+    if field.sub_fields is not None:
+        dtype_field = field.sub_fields[0]
+        return np.asarray(val, dtype=dtype_field.type_)
+    else:
+        return np.asarray(val)
+
+
+def coerce_2tensor(value: Union[float, List[float]], values: dict):
```
jklaise, line 72: Type information ambiguous, e.g. no return type and also seems `None` is an option?

alibi_detect/saving/validators.py
```diff
@@ -0,0 +1,84 @@
...
+    if value is None:
+        return value
+    framework = values.get('backend', None) or values.get('flavour', None)
```
jklaise, line 75: `.get` returns `None` as default if key doesn't exist (see previous comment).

alibi_detect/saving/_pytorch/saving.py
```diff
@@ -0,0 +1,134 @@
...
+from alibi_detect.cd.pytorch import UAE, HiddenOutput
+from alibi_detect.models.pytorch import TransformerEmbedding
+from alibi_detect.utils.frameworks import Framework
+
+logger = logging.getLogger(__name__)
+
+
+def save_model_config(model: Callable,
+                      base_path: Path,
+                      input_shape: Optional[tuple],
```
jklaise, line 19: `input_shape` unused?

alibi_detect/saving/_pytorch/saving.py
```diff
@@ -0,0 +1,134 @@
...
+            # preprocessing encoder
+            model = layers[1]
+        else:  # If UAE is simply an encoder
+            model = model.encoder
+    elif isinstance(model, TransformerEmbedding):
+        cfg_embed = save_embedding_config(model, base_path, local_path.joinpath('embedding'))
+        model = None
+    elif isinstance(model, HiddenOutput):
+        model = model.model
+    elif isinstance(model, nn.Module):  # Last as TransferEmbedding and UAE are nn.Module's
```
jklaise, line 57: typo in comment -> TransformerEmbedding

alibi_detect/saving/_pytorch/saving.py
jklaise: more simple than `tensorflow` I assume?

alibi_detect/saving/_pytorch/saving.py
```diff
@@ -0,0 +1,134 @@
...
+    # create folder to save model in
+    model_path = Path(filepath).joinpath(save_dir)
+    if not model_path.is_dir():
+        logger.warning('Directory {} does not exist and is now created.'.format(model_path))
+        model_path.mkdir(parents=True, exist_ok=True)
+
+    # save model
+    model_path = model_path.joinpath('model.pt')
+
+    if isinstance(model, (nn.Module, nn.Sequential)):
```
jklaise, line 96: Do we need to have both types here?

alibi_detect/saving/schemas.py
```diff
@@ -164,6 +156,8 @@ class ModelConfig(CustomBaseModel):
     :class:`~alibi_detect.cd.pytorch.preprocess.HiddenOutput` model is returned (dependent on `flavour`).
     Only applies to 'tensorflow' and 'pytorch' models.
     """
+    # Validators
+    _validate_flavour = validator('flavour', allow_reuse=True, pre=False)(validate_framework)
```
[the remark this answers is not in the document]
jklaise: What was the reason for passing validation in the case that `tensorflow` wasn't installed? I suppose we need to clearly delineate that now any validation is conditional on appropriate optional dependencies being available at runtime (e.g. a config may be "valid" if someone is just reading it, but it's not valid if they want to use it but a specified optional dependency isn't available).

alibi_detect/saving/tests/models.py
```diff
@@ -139,11 +175,15 @@ def classifier_model(backend, current_cases):
...
     if backend == 'tensorflow':
-        inputs = tf.keras.Input(shape=(input_dim,))
-        outputs = tf.keras.layers.Dense(2, activation=tf.nn.softmax)(inputs)
-        model = tf.keras.Model(inputs=inputs, outputs=outputs)
+        model = tf.keras.Sequential(
+               [
+                   tf.keras.layers.InputLayer(input_shape=(input_dim,)),
+                   tf.keras.layers.Dense(2, activation=tf.nn.softmax),
+               ]
+           )
```
jklaise, line 183: Any particular reason to switching to `Sequential`?

jklaise (commented):
LGTM! Please see comments.

alibi_detect/saving/schemas.py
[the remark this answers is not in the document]
ascillitoe: For the case where `backend='tensorflow'` but tensorflow isn't installed; previously it wasn't so much a design decision to allow this to pass validation, it was more that we simply didn't check against installed deps anywhere. The nice side-effect of this was for applications where we might want to generate configs and validate them, but not actually instantiate a detector, we could avoid having to install heavy deps like tensorflow.

However, now that we coerce to `Tensor`'s during validation, we need tf and torch to be installed. It makes sense to do the validation of `flavour` against installed deps here, as it simplifies the logic later on when we coerce to `Tensor`.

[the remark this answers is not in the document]
ascillitoe: > I suppose we need to clearly delineate that now any validation is conditional on appropriate optional dependencies being available at runtime (e.g. a config may be "valid" if someone is just reading it, but it's not valid if they want to use it but a specified optional dependency isn't available).

Yes exactly. This is a subtle but important change. We are now validating a config wrt to whether it is "valid" on a given enviroment.

alibi_detect/utils/_types.py
ascillitoe: Good point. issue opened: https://github.com/SeldonIO/alibi-detect/issues/668

Fair point about the naming now... I thought it might be worth keeping as plural for consistency with `supported_models_all`, and in case we need to make them collections again in the future. Can change to singular if you think its better though...

alibi_detect/cd/pytorch/preprocess.py
```diff
@@ -7,6 +7,63 @@
...
+                nn.ReLU(),
+                nn.Linear(enc_dim + step_dim, enc_dim)
+            )
+        else:
+            raise ValueError('Need to provide either `enc_dim` and `step_dim` or a '
+                             'nn.Module `mlp`')
+
+    def forward(self, x: Union[np.ndarray, torch.Tensor, Dict[str, torch.Tensor]]) -> torch.Tensor:
+        if isinstance(self.input_layer, nn.Module):
+            x = self.input_layer(x)
```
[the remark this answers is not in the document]
ascillitoe: Thanks, good spot. This is an error. An earlier iteration accepted `input_layer` as an `int`. I'll remove.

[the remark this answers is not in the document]
ascillitoe: Removing as an error. See https://github.com/SeldonIO/alibi-detect/pull/656#discussion_r1006863953.

alibi_detect/cd/pytorch/preprocess.py
```diff
@@ -7,6 +7,63 @@
...
+
+
+class UAE(nn.Module):
+    def __init__(
+            self,
+            encoder_net: Optional[nn.Module] = None,
+            input_layer: Optional[nn.Module] = None,
+            shape: Optional[tuple] = None,
+            enc_dim: Optional[int] = None
+    ) -> None:
```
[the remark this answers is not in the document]
ascillitoe: We have an issue regarding this, but maybe we should increase the priority?

https://github.com/SeldonIO/alibi-detect/issues/484

alibi_detect/utils/_types.py
jklaise: Why not keep them tuples of one member, e.g. `(torch.nn.Module, )`? Would keep consistency and perhaps extensibility (if it turns out, for any reason, that we might only support a subset of model types.

alibi_detect/saving/_pytorch/saving.py
ascillitoe: Oops... yes tensorflow

alibi_detect/cd/pytorch/preprocess.py
```diff
@@ -7,6 +7,63 @@
...
+        super().__init__()
+        is_enc = isinstance(encoder_net, nn.Module)
+        is_enc_dim = isinstance(enc_dim, int)
+        if is_enc:
+            self.encoder = encoder_net
+        elif not is_enc and is_enc_dim:  # set default encoder
+            input_dim = np.prod(shape)
+            step_dim = int((input_dim - enc_dim) / 3)
+            self.encoder = _Encoder(input_layer, input_dim=input_dim, enc_dim=enc_dim, step_dim=step_dim)
```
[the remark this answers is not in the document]
ascillitoe: I've reworked this in e24e7bb3cae512df912463bf25ab9c4df3a202af. `input_layer` is now `Optional` within the `_Encoder`, and it is just skipped in `forward` if it is `None`.

alibi_detect/cd/pytorch/lsdd.py
```diff
@@ -116,6 +116,8 @@ def _configure_normalization(self, x_ref: torch.Tensor, eps: float = 1e-12):
         x_ref_means = x_ref.mean(0)
         x_ref_stds = x_ref.std(0)
         self._normalize = lambda x: (torch.as_tensor(x) - x_ref_means) / (x_ref_stds + eps)  # type: ignore[assignment]
+        self._unnormalize = lambda x: (torch.as_tensor(x) * (x_ref_stds + eps)  # type: ignore[assignment]
```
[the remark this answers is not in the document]
ascillitoe: The `+ x_ref_means).cpu().numpy()` is on the next line :)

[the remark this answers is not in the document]
jklaise: Foiled by a line break...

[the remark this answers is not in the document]
ascillitoe: Its the need to have the '# type: ignore' on the first line that always tricks me!

alibi_detect/utils/pytorch/kernels.py
```diff
@@ -113,12 +138,16 @@ class DeepKernel(nn.Module):
     def __init__(
         self,
         proj: nn.Module,
-        kernel_a: nn.Module = GaussianRBF(trainable=True),
-        kernel_b: Optional[nn.Module] = GaussianRBF(trainable=True),
+        kernel_a: Union[nn.Module, str] = 'rbf',
+        kernel_b: Optional[Union[nn.Module, str]] = 'rbf',
```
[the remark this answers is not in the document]
ascillitoe: There is also already an issue for this! https://github.com/SeldonIO/alibi-detect/issues/463

alibi_detect/cd/pytorch/lsdd_online.py
```diff
@@ -114,6 +115,7 @@ def _configure_normalization(self, eps: float = 1e-12):
         x_ref_means = x_ref.mean(0)
         x_ref_stds = x_ref.std(0)
         self._normalize = lambda x: (x - x_ref_means) / (x_ref_stds + eps)
+        self._unnormalize = lambda x: (torch.as_tensor(x) * (x_ref_stds + eps) + x_ref_means).cpu().numpy()
```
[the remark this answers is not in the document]
ascillitoe: I didn't notice this. I've done some digging and think the difference seems to be due to where they are used:

- `LSDDDriftOnline`: `_normalize` only used within the torch backend `score` method:
https://github.com/SeldonIO/alibi-detect/blob/1af9fc053ac224a599f16fd78e0d894de12f89f1/alibi_detect/cd/pytorch/lsdd_online.py#L205-L207
   Therefore, `_normalize` always recieves a `torch.Tensor`.
   
- `LSDDDrift`: `_normalize` is used in the `torch` backend `score` method, but also in the base `predict` method (in order to serve the `update_reference` functionality:
https://github.com/SeldonIO/alibi-detect/blob/1af9fc053ac224a599f16fd78e0d894de12f89f1/alibi_detect/cd/base.py#L797-L803
   In this case, `_normalize` recieves a `np.ndarray`. The type signature of `_configure_normalize` is still correct however, since this is called within the torch backend hence recieves a `torch.Tensor`.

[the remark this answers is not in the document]
ascillitoe: For the `_unnormalize` functions it's a similar story. The `x_ref_means` and `x_ref_stds` are `torch.Tensor`'s, but `_unnormalize` is called in the top-level `LSDDDrift.get_config`, hence `x` is a `np.ndarray` and needs converting back to `torch.Tensor`.

alibi_detect/saving/validators.py
ascillitoe: See https://github.com/SeldonIO/alibi-detect/pull/656#discussion_r1007851569

alibi_detect/saving/loading.py
```diff
@@ -260,16 +265,13 @@ def _load_model_config(cfg: dict) -> Callable:
     custom_obj = cfg['custom_objects']
     layer = cfg['layer']
     src = Path(src)
-    if not src.is_dir():
-        raise FileNotFoundError("The `src` field is not a recognised directory. It should be a directory containing "
-                                "a compatible model.")
```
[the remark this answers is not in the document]
ascillitoe: No you're right, it doesn't make sense to remove this. Will add back in!

alibi_detect/saving/validators.py
ascillitoe: Yep. In the code-snippet in [this section](https://pydantic-docs.helpmanual.io/usage/types/#generic-classes-as-types) of the pydantics docs they suggest including `ModelField` for code completion.

alibi_detect/saving/validators.py
```diff
@@ -0,0 +1,84 @@
...
+
+        @classmethod
+        def validate(cls, val: Any, field: ModelField) -> Optional[np.ndarray]:
+            return _coerce2NDArray(cls, val, field)
+
+
+def _coerce2NDArray(cls: Type, val: Any, field: ModelField) -> np.ndarray:
+    if field.sub_fields is not None:
+        dtype_field = field.sub_fields[0]
+        return np.asarray(val, dtype=dtype_field.type_)
```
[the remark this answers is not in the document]
ascillitoe: FYI this is not new. This is part of the `NDArray` custom type class which I moved from `utils._types` to here. The `_coerce2NDArray` used to be called `_validate` (see below). I renamed it for consistency with `coerce_2tensor` etc.
https://github.com/SeldonIO/alibi-detect/blob/1af9fc053ac224a599f16fd78e0d894de12f89f1/alibi_detect/utils/_types.py#L50-L55

I am actually going to tweak these to be `coerce_2_ndarray` and `coerce_2_tensor`.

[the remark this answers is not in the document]
ascillitoe: RE how the work. They work a bit differently because `_coerece2NDArray` is called by the `NDArray` pydantic [custom data type ](https://pydantic-docs.helpmanual.io/usage/types/#custom-data-types), which is used in `schemas` as follows:

```
sigma: Optional[NDArray[np.float32]] = None
```

whilst `coerce_2tensor` is itself a validator, so is used like:

```
    # Validators
    _validate_flavour = validator('flavour', allow_reuse=True, pre=False)(validate_framework)
    _coerce_sigma2tensor = validator('sigma', allow_reuse=True, pre=False)(coerce_2tensor)
```

Wrt to what is happening in `dtype_field = field.sub_fields[0]`, the `np.float32` given to `NDArray` in my example is inserted as the first element in `field.sub_fields[0]`. The `val` is the actual value of `sigma` passed to the validator. Hence we coerce `val` to `np.float32`.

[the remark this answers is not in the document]
ascillitoe: Final bit of info. I did want to do `coerce_2tensor` in the same way as we do `NDArray` (i.e. as a custom data type), however I couldn't figure out a way to get the `flavour` info in there properly. So went for the less elegant validator approach for this.

alibi_detect/saving/validators.py
ascillitoe: Thinking about this more, I can add a unit test for `coerce_2tensor` to `test_validate`. But for `validate_framework` I'm not so sure what the best approach is i.e. how to mock `has_*`. @mauicv any ideas?

alibi_detect/saving/_pytorch/saving.py
ascillitoe: Thanks. Have removed for now (pending discussion in #668 )

alibi_detect/utils/_types.py
```diff
@@ -17,12 +17,12 @@
...
 if has_pytorch:
     import torch
-    supported_models_torch = torch.nn.Module
-    supported_optimizers_torch = type  # Note type not object!
-supported_models_sklearn = BaseEstimator
+    supported_models_torch = (torch.nn.Module, )
+    supported_optimizers_torch = (type, )  # Note type not object!
+supported_models_sklearn = (BaseEstimator, )
 supported_models_all = (supported_models_tf, supported_models_torch, supported_models_sklearn)
 supported_optimizers_all = (supported_optimizers_tf, supported_optimizers_torch)
```
jklaise, line 28: Is this a bug now that the members are tuples of length 1 (if so, why isn't anything failing?)? Should we use addition to create a new tuple?

alibi_detect/saving/tests/test_validate.py
```diff
@@ -80,3 +84,35 @@ def test_validate_config_wo_meta(cfg):
...
+    # Define a kernel config
+    kernel_cfg = {
+        'src': f'@utils.{flavour}.kernels.GaussianRBF',
+        'flavour': flavour,
+        'sigma': sigma
+    }
+
+    # Pass through validation and check results
+    kernel_cfg_val = KernelConfig(**kernel_cfg).dict()
+    print(kernel_cfg_val)
```
jklaise, line 109: unnecessary

jklaise (changes requested):
Would like to see that detailed CHANGELOG on this PR as there are quite a few things going on here.

alibi_detect/utils/_types.py
ascillitoe: Passing tuples of tuples to `isinstance` seems to be fine. For example:

```python
a = 2
test = ((str,), (float,))
isinstance(a, test)  # gives False
```

However, I agree its ugly, so will change to `supported_optimizers_tf + supported_optimizers_torch`

alibi_detect/saving/saving.py
mauicv: Do we never need to save the `optimizer` state in the case of pytorch?

jklaise (approved):

alibi_detect/saving/_pytorch/saving.py
```diff
@@ -0,0 +1,131 @@
...
+    if not filepath.is_dir():
+        logger.warning('Directory {} does not exist and is now created.'.format(filepath))
+        filepath.mkdir(parents=True, exist_ok=True)
+
+    # Populate config dict
+    cfg_embed = {}  # type: Dict[str, Any]
+    cfg_embed.update({'type': embed.emb_type})
+    cfg_embed.update({'layers': embed.hs_emb.keywords['layers']})
+    cfg_embed.update({'src': local_path})
+    cfg_embed.update({'flavour': Framework.PYTORCH.value})
```
mauicv, line 125: Does it make sense to just have a `get_config` method on the embedding model? like how we do with the [tensorflow kernel](https://github.com/SeldonIO/alibi-detect/blob/57488644af894a7e4e528718c7d4ef86156ee2fa/alibi_detect/utils/tensorflow/kernels.py#L91).

mauicv: hmm, although looking again at [this](https://github.com/SeldonIO/alibi-detect/blob/57488644af894a7e4e528718c7d4ef86156ee2fa/alibi_detect/utils/tensorflow/kernels.py#L91) i've realised that these things aren't exactly similar... the tf kernel `get_config` overide is just used to convert `sigma` from `torch.tensor` to a `numpy` array. I guess what I mean is would it make sense to move the config stuff to the embedding object instead of doing it here?

alibi_detect/saving/schemas.py
```diff
@@ -668,6 +681,8 @@ class MMDDriftConfig(DriftDetectorConfig):
     n_permutations: int = 100
     batch_size_permutations: int = 1000000
     device: Optional[Literal['cpu', 'cuda']] = None
+    # Validators
+    _validate_flavour = validator('backend', allow_reuse=True, pre=False)(validate_framework)
```
mauicv, line 685: Can this be abstracted out into a base class?

ascillitoe: For clarification, you mean adding something like a `DriftDetectorWithBackendConfig` base pydantic model, similar to the existing `DriftDetectorConfig`, but with the `_validate_framework` validator, and then having all pydantic models such as `MMDDriftConfig` inherit from this? 

That might work, the only thing I am unsure about is that the `backend: Literal['tensorflow', 'pytorch'] = 'tensorflow'` field needs to be defined in the specific detector pydantic models, because different detectors have different `backend`'s. I'm not sure if having the `validate_framework` validator in the base model will work in that case, because `backend` isn't defined in the base model. Will look into it...

alibi_detect/cd/pytorch/preprocess.py
mauicv: Lazy Linear is also [in active development](https://github.com/pytorch/pytorch/blob/4618371da56c887195e2e1d16dad2b9686302800/torch/nn/modules/lazy.py#L180) and if we did use them we might want to wait until they're stable.

alibi_detect/saving/schemas.py
mauicv: w.r.t. clarification: yes exactly.

mauicv (approved):
Some minor questions but otherwise LGTM!

alibi_detect/saving/schemas.py
ascillitoe: Cheers for this suggestion, https://github.com/SeldonIO/alibi-detect/pull/656/commits/11c760944e80b08090a53d7b0fb53e3bb25e5aac moves `validate_framework` to `DetectorConfig`. This just required setting `check_fields=False` so that pydantic does not throw an error because `backend` is not present in `DetectorConfig`. If `backend` is present the detector config (e.g. `MMDDriftConfig`), the validator is applied. If it isn't present (e.g. `KSDriftConfig`), the validator isn't applied.

A bit cleaner I think!

mauicv: Nice!

jklaise:
@ascillitoe fine to leave it I think, but in that case would add a line to changelog.

ascillitoe:
> @ascillitoe fine to leave it I think, but in that case would add a line to changelog.

Thanks, will do!

alibi_detect/saving/_pytorch/saving.py
arnaudvl: Follow up?

alibi_detect/saving/_pytorch/loading.py
arnaudvl: Types of `embedding_type` and `layers`.

arnaudvl: What is the benefit of `load_embedding` if it just calls `TransformerEmbedding` with the same args/kwargs? Consistency in naming?

arnaudvl:
Is it worth adding a small example on serialisation?

alibi_detect/cd/mmd_online.py
ascillitoe: Based on offline discussion with @mauicv, we've decided to leave this one for now since it is quite specific to a very small number of detectors.

alibi_detect/saving/_pytorch/loading.py
ascillitoe: Yeh just for consistency in the top-level (backend agnostic) `load_embedding_config`:

https://github.com/SeldonIO/alibi-detect/blob/ac01d352b31607473970307080096a9c7b4d5744/alibi_detect/saving/loading.py#L292-L300

ascillitoe: Actually sorry, that's not quite true. It's also hides the below import away in the `_pytorch.saving` submodule:

https://github.com/SeldonIO/alibi-detect/blob/ac01d352b31607473970307080096a9c7b4d5744/alibi_detect/saving/_pytorch/loading.py#L13

Hiding the backend-specific stuff in backend-specific submodules is something we've been generally trying to follow since the optional deps work (although admittedly here it leads to a bit of a needless abstraction...)

alibi_detect/saving/_tensorflow/loading.py
ascillitoe: Done in https://github.com/SeldonIO/alibi-detect/pull/656/commits/f3e94d33c140810983a627684622758770b93455.

ascillitoe:
> Is it worth adding a small example on serialisation?

@arnaudvl I've added this to the MMD/CIFAR10 example in [9a5fc10](https://github.com/SeldonIO/alibi-detect/pull/656/commits/9a5fc101d94865164019498b80f3ec05c15f48a1).

arnaudvl (approved):
LGTM!

PR #661, opened 2022-10-21
Fix incorrect default `alternative` kwarg and docs in FET detectors

ascillitoe:
- The default for `alternative` was wrong in `FETDrift`, meaning it could not be instantiated with default kwarg's. 
- Added clarification and corrections to docs.
- Changed defaults to `less` for offline and online. Wondering if this makes more sense than `greater`, given that typically `0` might represent an incorrect prediction and `1` a correct prediction. i.e. by default we check for a decrease in model performance. @ojcobb thoughts on this one?


ascillitoe:
CI triggered here:
https://github.com/SeldonIO/alibi-detect/actions/runs/3296720117

jklaise (approved):
LGTM (pending CI)

ascillitoe:
@jklaise just realised the CI is not testing the PR, but of course is testing the original `patch/v0.10.4` branch (i.e. at `v0.10.3` tag). I'll merge this and then we can check CI before moving ahead with release. IMO we should name all branches like this `release/v...` and then turn on CI for PR's into `release/` branches...

PR #665, opened 2022-10-24
Return the OOF instances for classifier-based drift detectors (ClassifierDrift and SpotTheDiffDrift)

arnaudvl:
When using the ClassifierDrift or SpotTheDiffDrift detectors, you can return the prediction probabilities on the out-of-fold instances. However, when you use `train_size` instead of cross-validation, you don't know which instances these probabilities correspond to, since you don't know which reference and test instances are used for training or prediction. This PR adds the associated reference and test instances to the detector's return dictionary.


alibi_detect/cd/tensorflow/classifier.py
```diff
@@ -144,7 +144,8 @@ def __init__(
         if isinstance(train_kwargs, dict):
             self.train_kwargs.update(train_kwargs)
 
-    def score(self, x: np.ndarray) -> Tuple[float, float, np.ndarray, np.ndarray]:  # type: ignore[override]
+    def score(self, x: np.ndarray) \
```
jklaise, line 147: `mypy` is failing because of the new line break causing the ignore to be on the next line

alibi_detect/cd/base.py
```diff
@@ -240,7 +240,7 @@ def test_probs(
         return p_val, dist
 
     @abstractmethod
-    def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray, np.ndarray]:
+    def score(self, x: Union[np.ndarray, list]) -> Tuple[float, float, np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
```
jklaise, line 243: Note for future - reconsider making `score` a private method. This would also allow us to use more maintainable data structures for the return type, e.g. a dataclass.

jklaise (commented):
LGTM pending CI passing.

alibi_detect/cd/base.py
arnaudvl: Opened #666 for this

alibi_detect/cd/sklearn/classifier.py
```diff
@@ -272,24 +273,36 @@ def _score(self, x: Union[np.ndarray, list]) \
...
         idx_oob = np.where(np.all(~np.isnan(self.model.oob_decision_function_), axis=1))[0]
         probs_oob = self.model.oob_decision_function_[idx_oob]
-        x_oob, y_oob = x[idx_oob], y[idx_oob]
+        y_oob = y[idx_oob]
+        if isinstance(x, np.ndarray):
+            x_oob = x[idx_oob]
+        elif isinstance(x, list):
+            x_oob = [x[_] for _ in idx_oob]
+        else:
+            raise TypeError(f'x needs to be of type np.ndarray or list and not {type(x)}.')
```
jklaise, line 305: Noting that this is different to the pattern in previous parts where we only have a single `if/else` and no raising of error otherwise.

jklaise (approved):
LGTM, would suggest updating the changelog as well to reduce burden come release time.

alibi_detect/cd/sklearn/classifier.py
arnaudvl: Indeed, because elsewhere a check has already been carried out during training for the TypeError so it can be dropped later. Here this isn't case.

PR #675, opened 2022-11-14
Fix notebook tests

### LINKED ISSUE #664: Notebook CI failing
Our notebook tests are failing with:

```
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/site-packages/pytest_randomly/__init__.py", line 194, in pytest_runtest_setup
    _reseed(item.config, -1)
  File "/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/site-packages/pytest_randomly/__init__.py", line 172, in _reseed
    reseed(seed)
  File "/opt/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/site-packages/thinc/util.py", line 70, in fix_random_seed
    numpy.random.seed(seed)
  File "mtrand.pyx", line 246, in numpy.random.mtrand.RandomState.seed
  File "_mt19937.pyx", line 1[66](https://github.com/SeldonIO/alibi-detect/actions/runs/3257690768/jobs/5349135028#step:5:67), in numpy.random._mt19937.MT19937._legacy_seeding
  File "_mt19937.pyx", line 180, in numpy.random._mt19937.MT19937._legacy_seeding
ValueError: Seed must be between 0 and 2**32 - 1
```

It appears to involve the `numpy` seed setting, `pytest_randomly`, and `thinc`. Perhaps `thinc`'s `pytest_randomly` entry point is overriding ours? 

Requires further investigation...

ascillitoe:
`pytest-randomly` is disabled in the notebook CI in order to resolve #664 (see https://github.com/pytest-dev/pytest-randomly/issues/495). This isn't an ideal long-term solution, we might want to think about replacing `pytest-randomly` if https://github.com/pytest-dev/pytest-randomly/issues/495 is not going to be fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,7 +81,9 @@ jobs:
           if [ "$RUNNER_OS" == "macOS" ]; then  # Avoid numba/OpenMP segfault in CVMDrift (https://github.com/SeldonIO/alibi-detect/issues/648)
             export NUMBA_THREADING_LAYER="workqueue"
           fi
-          pytest alibi_detect
+          pytest --randomly-seed=0 alibi_detect
+          # Note: The pytest-randomly seed is fixed at 0 for now. Once the legacy np.random.seed(0)'s 
+          # are removed from tests, this can be removed, allowing all tests to use random seeds.
 
       - name: Upload coverage to Codecov
         if: ${{ success() }}
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,13 +11,8 @@ addopts =
     --tb native
     -W ignore
     --cov=alibi_detect
-    --randomly-dont-reorganize
-    --randomly-seed=0
     #-n auto
     #--forked
-# randomly-dont-reorganize prevents pytest-randomly from reordering tests
-# The pytest-randomly seed is fixed at 0 for now. Once the legacy np.random.seed(0)'s 
-# are removed from tests, this can be removed, allowing all tests to use random seeds.
 
 [flake8]
 max-line-length = 120
--- a/.github/workflows/test_all_notebooks.yml
+++ b/.github/workflows/test_all_notebooks.yml
@@ -45,4 +45,4 @@ jobs:
 
       - name: Run notebooks
         run: |
-          pytest --suppress-no-test-exit-code --no-cov -rA --durations=0 -vv testing/test_notebooks.py
+          pytest --suppress-no-test-exit-code --no-cov -rA --durations=0 -vv -p no:randomly testing/test_notebooks.py
--- a/.github/workflows/test_changed_notebooks.yml
+++ b/.github/workflows/test_changed_notebooks.yml
@@ -65,4 +65,4 @@ jobs:
         # adding the `or` quantifier between the names and concatenating with the test name `test_notebook_execution`.
         run: |
           tests="test_notebook_execution[$(echo ${FILES} | sed 's|doc/source/examples/||g' | sed 's| | or |g')]" &&
-          pytest --suppress-no-test-exit-code --no-cov -rA --durations=0 -vv testing/test_notebooks.py -k "$tests"
+          pytest --suppress-no-test-exit-code --no-cov -rA --durations=0 -vv -p no:randomly testing/test_notebooks.py -k "$tests"


ascillitoe:
Notebook CI run:
https://github.com/ascillitoe/alibi-detect/actions/runs/3460987064

setup.cfg
```diff
@@ -11,13 +11,8 @@ addopts =
     --tb native
     -W ignore
     --cov=alibi_detect
-    --randomly-dont-reorganize
```
mauicv, line 14: Why do we remove this here?

mauicv (approved):
One minor question otherwise LGTM!

setup.cfg
ascillitoe: For `--randomly-seed=0`, I've moved it to `ci.yml` because having it in the global pytest config caused an error in the notebook CI because `pytest-randomly` is now disabled there.

For `--randomly-dont-reorganize`, I decided we could remove it anyway. This controls whether or not the order of tests is randomly reorganized. From the `pytest-randomly` docs:

> By randomly ordering the tests, the risk of surprising inter-test dependencies is reduced - a technique used in many places, for example Google’s C++ test runner [googletest](https://code.google.com/p/googletest/wiki/V1_5_AdvancedGuide#Shuffling_the_Tests). Research suggests that “dependent tests do exist in practice” and a random order of test executions can effectively detect such dependencies [[1]](https://pypi.org/project/pytest-randomly/#footnote-1).

ascillitoe:
@jklaise @mauicv I'm now wondering whether we park this until we've decided what to do about `pytest-randomly`. I added this plugin in https://github.com/SeldonIO/alibi-detect/pull/496 to give the ability to set random seeds within a context manager for testing (useful when repeatedly perfoming random operations i.e. instantiating some detectors, during testing). 

However, https://github.com/SeldonIO/alibi-detect/issues/664 uncovered a problematic issue with the way custom `pytest_randomly` seeders are defined via `entry_points`. Essentially at present there is no way to disable entry points added by our deps (in this case `thinc`). I came opened a PR on `pytest-randomly` with an idea for a fix, but the devs don't seem to be super responsive at the moment. I'm starting to think we either need to fork `pytest-randomly`, drop it, or be patient!

setup.cfg
ascillitoe: btw we can add `--randomly-dont-reorganize` to `ci.yml` if you think this change shouldn't be made without further discussion...

mauicv: Ah yeah, I was curious about `--randomly-dont-reorganise`, it seems useful no? I suppose because we don't really deal with persistent state it's not the type of thing that's going to help us much. Although perhaps it's worth keeping in for the saving tests, that's the only place where I can image one test leaking into another...

ascillitoe: Yeh agree that I can't think of any obvious places where tests might be leaky. I hope that the saving ones are ok since they use pytest's [tmp_path](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html). I sort of think if we keep `pytest_randomly` it doesn't hurt to randomise test order, but if we're removing it its not worth implementing an alternative solution to randomise tests.

mauicv: yeah, aggreed

.github/workflows/ci.yml
```diff
@@ -81,7 +81,9 @@ jobs:
           if [ "$RUNNER_OS" == "macOS" ]; then  # Avoid numba/OpenMP segfault in CVMDrift (https://github.com/SeldonIO/alibi-detect/issues/648)
             export NUMBA_THREADING_LAYER="workqueue"
           fi
-          pytest alibi_detect
+          pytest --randomly-seed=0 alibi_detect
+          # Note: The pytest-randomly seed is fixed at 0 for now. Once the legacy np.random.seed(0)'s 
+          # are removed from tests, this can be removed, allowing all tests to use random seeds.
```
jklaise, line 86: Not sure I understand this change, also isn't this independent of the notebook tests?

setup.cfg
jklaise: Removing `--randomly-dont-reorganise` actually enables random order, right? In which case it seems useful to have?

jklaise:
@ascillitoe I think it would be great if we could get the notebook tests passing again, I don't think it's a big deal to disable `pytest-randomly` for those for the time being.

setup.cfg
ascillitoe: Yep, random order has been enabled by removing the flag.

ascillitoe: @jklaise just to confirm, are you saying random order seems useful to have? i.e. we stick to having the flag removed...

jklaise: Correct

ascillitoe: OK thanks. Think this is good to merge then 👍🏻

PR #681, opened 2022-11-23
Support for serialising detectors with keops backends

ascillitoe:
Following on from #656, this PR adds support for serialising detectors with `backend='keops'`. Since keops detectors are still pytorch based, this is a relatively simple addition. The main changes involve removing `NotImplementedError`'s and support for saving keops kernels.

### DIFF
--- a/alibi_detect/saving/loading.py
+++ b/alibi_detect/saving/loading.py
@@ -15,6 +15,7 @@ from alibi_detect.saving._tensorflow import load_detector_legacy, load_embedding
     load_model_tf, load_optimizer_tf, prep_model_and_emb_tf, get_tf_dtype
 from alibi_detect.saving._pytorch import load_embedding_pt, load_kernel_config_pt, load_model_pt, \
     load_optimizer_pt, prep_model_and_emb_pt, get_pt_dtype
+from alibi_detect.saving._keops import load_kernel_config_ke
 from alibi_detect.saving._sklearn import load_model_sk
 from alibi_detect.saving.validate import validate_config
 from alibi_detect.base import Detector, ConfigurableDetector
@@ -135,11 +136,6 @@ def _load_detector_config(filepath: Union[str, os.PathLike]) -> ConfigurableDete
     cfg = validate_config(cfg, resolved=True)
     logger.info('Validated resolved config.')
 
-    # Backend
-    backend = cfg.get('backend')
-    if backend is not None and backend.lower() not in (Framework.TENSORFLOW, Framework.PYTORCH, Framework.SKLEARN):
-        raise NotImplementedError('Loading detectors with keops backend is not yet supported.')
-
     # Init detector from config
     logger.info('Instantiating detector.')
     detector = _init_detector(cfg)
@@ -186,8 +182,10 @@ def _load_kernel_config(cfg: dict, backend: str = Framework.TENSORFLOW) -> Calla
     """
     if backend == Framework.TENSORFLOW:
         kernel = load_kernel_config_tf(cfg)
-    else:
+    elif backend == Framework.PYTORCH:
         kernel = load_kernel_config_pt(cfg)
+    else:  # backend=='keops'
+        kernel = load_kernel_config_ke(cfg)
     return kernel
 
 
--- a/alibi_detect/saving/tests/models.py
+++ b/alibi_detect/saving/tests/models.py
@@ -21,12 +21,14 @@ from alibi_detect.utils.pytorch.kernels import GaussianRBF as GaussianRBF_pt
 from alibi_detect.utils.pytorch.kernels import DeepKernel as DeepKernel_pt
 from alibi_detect.utils.tensorflow.kernels import GaussianRBF as GaussianRBF_tf
 from alibi_detect.utils.tensorflow.kernels import DeepKernel as DeepKernel_tf
-from alibi_detect.utils.keops.kernels import GaussianRBF as GaussianRBF_ke
-from alibi_detect.utils.keops.kernels import DeepKernel as DeepKernel_ke
 from alibi_detect.models.pytorch import TransformerEmbedding as TransformerEmbedding_pt
 from alibi_detect.models.tensorflow import TransformerEmbedding as TransformerEmbedding_tf
 from alibi_detect.cd.pytorch import HiddenOutput as HiddenOutput_pt
 from alibi_detect.cd.tensorflow import HiddenOutput as HiddenOutput_tf
+from alibi_detect.utils.frameworks import has_keops
+if has_keops:
+    from alibi_detect.utils.keops.kernels import GaussianRBF as GaussianRBF_ke
+    from alibi_detect.utils.keops.kernels import DeepKernel as DeepKernel_ke
 
 LATENT_DIM = 2  # Must be less than input_dim set in ./datasets.py
 DEVICE = "cuda" if torch.cuda.is_available() else "cpu"
--- a/alibi_detect/saving/tests/models.py
+++ b/alibi_detect/saving/tests/models.py
@@ -26,7 +26,7 @@ from alibi_detect.models.tensorflow import TransformerEmbedding as TransformerEm
 from alibi_detect.cd.pytorch import HiddenOutput as HiddenOutput_pt
 from alibi_detect.cd.tensorflow import HiddenOutput as HiddenOutput_tf
 from alibi_detect.utils.frameworks import has_keops
-if has_keops:
+if has_keops:  # pykeops only installed in Linux CI
     from alibi_detect.utils.keops.kernels import GaussianRBF as GaussianRBF_ke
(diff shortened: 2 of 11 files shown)


alibi_detect/utils/keops/kernels.py
```diff
@@ -115,13 +117,36 @@ def forward(self, x: LazyTensor, y: LazyTensor, infer_sigma: bool = False) -> La
             kernel_mat = kernel_mat.sum(-1) / len(self.sigma)
         return kernel_mat
 
+    def get_config(self) -> dict:
+        """
+        Returns a serializable config dict (excluding the input_sigma_fn, which is serialized in alibi_detect.saving).
+        """
+        cfg = self.config.copy()
+        if isinstance(cfg['sigma'], torch.Tensor):
+            cfg['sigma'] = cfg['sigma'].detach().cpu().numpy().tolist()
```
jklaise, line 126: Are we certain we don't need a deep copy here (depends on the dict values)? Wouldn't be ideal if `self.config` turns out to not be idempotent (different depending whether `get_config` has been called or not.

jklaise: I guess the same question applies to every component for which we have `self.config` and which may have mutable values - do we need to do an audit / have tests?

alibi_detect/utils/keops/kernels.py
```diff
@@ -115,13 +117,36 @@ def forward(self, x: LazyTensor, y: LazyTensor, infer_sigma: bool = False) -> La
...
+
 
 class DeepKernel(nn.Module):
     def __init__(
         self,
         proj: nn.Module,
-        kernel_a: nn.Module = GaussianRBF(trainable=True),
-        kernel_b: Optional[nn.Module] = GaussianRBF(trainable=True),
+        kernel_a: Union[nn.Module, str] = 'rbf',
+        kernel_b: Optional[Union[nn.Module, str]] = 'rbf',
```
jklaise, line 149: I would say to use `Literal`s but I know you'll point me to the other open issue... :)

alibi_detect/utils/keops/kernels.py
```diff
@@ -176,3 +205,10 @@ def forward(self, x_proj: LazyTensor, y_proj: LazyTensor, x: Optional[LazyTensor
         if self.kernel_b is not None:
             similarity = (1-self.eps)*similarity + self.eps*self.kernel_b(x, y)
         return similarity
+
+    def get_config(self) -> dict:
+        return self.config.copy()
```
jklaise, line 210: No `flavour` here?

alibi_detect/saving/tests/models.py
```diff
@@ -25,6 +25,10 @@
 from alibi_detect.models.tensorflow import TransformerEmbedding as TransformerEmbedding_tf
 from alibi_detect.cd.pytorch import HiddenOutput as HiddenOutput_pt
 from alibi_detect.cd.tensorflow import HiddenOutput as HiddenOutput_tf
+from alibi_detect.utils.frameworks import has_keops
+if has_keops:
```
jklaise, line 29: Is this conditional purely because we skip `keops` on some platforms in CI? Should we add a comment for this? Same question for the other test module.

jklaise (commented):
A few minor comments but otherwise LGTM!

alibi_detect/utils/keops/kernels.py
ascillitoe: Upon review of this I think you're right. We seem to get away with it in our tests (and in the informal tests I did previously) because when we update items in `config` we are typically redefining the entire item, rather than mutating it (e.g. we change `sigma` from `None` to a `Tensor`, rather than mutating an element in `sigma`). I think this means the shallow copy is OK since the reference is broken, e.g.:

```python
orig_a = [1,2,3]
dict1 = {'a': deepcopy(orig_a), 'b': 'cat'}
dict2 = dict1.copy()

dict1['a'] = [1,1,1]

dict2['a'] == orig_a

>>> True
```

However, as you say, if we were to mutate an item in `config`, we would have a problem:
```python
orig_a = [1,2,3]
dict1 = {'a': deepcopy(orig_a), 'b': 'cat'}
dict2 = dict1.copy()

dict1['a'][1] = 1 

dict2['a'] == orig_a

>>> False
```

I might be missing something but seems we should add `deepcopy` to all of these. I can update these ones in this PR and then do a follow-up for the rest of them?

jklaise: Yea I think better safe than sorry.

alibi_detect/utils/keops/kernels.py
ascillitoe: Ha ha I hope I'm not developing a reputation for deflecting to other issues!

I can see why `Literal` might be a good idea here since we really do only accept `rbf` as a `str`. Since we'd want to do this for the pytorch and tensorflow kernels it might be best for me to follow-up straight after this PR?

alibi_detect/utils/keops/kernels.py
ascillitoe: Bear with me on this one! We have `flavour` in `ModelConfig` etc (see `saving.schemas`) because we use this to decide which `load_model_...` function to use (i.e. `_tf`, `_pt` or `_sk`). I've kept `flavour` decoupled from the detector `backend` here so that we have the option of using different flavour preprocessing to the model backend (e.g. a scikit-learn preprocessor with a tensorflow MMD detector). Plus, we might have a `model` even when there is no backend (e.g. any preprocessing model with KS detector). 

The situation is a little different for kernels. These are only used for detectors which have backends, and it doesn't make sense to decouple their `flavour` from `backend`. Ideally we would not actually have `flavour` at all for kernels (we actually decide which `load_kernel_...` function to use based on the detector `backend`. The only reason we have `flavour` is so that the `coerce_2_tensor` pydantic validator knows what type of `Tensor` to coerce `sigma` to (this happens before we get to to `load_kernel_...`. 

We don't need to do any coercion for `DeepKernel` hence why no `flavour` in `DeepKernelConfig`. I could add it in just for consistency? But it wouldn't actually be used at load time.

alibi_detect/saving/tests/models.py
ascillitoe: Yep precisely. I'll add notes to both modules now!

alibi_detect/utils/keops/kernels.py
ascillitoe: Tracker: https://github.com/SeldonIO/alibi-detect/issues/683

alibi_detect/utils/keops/kernels.py
ascillitoe: Tracker: https://github.com/SeldonIO/alibi-detect/issues/683

ascillitoe: Done for keops kernel only in 669a8c60bc4cdb1bdd6e6f6ead9008bfa7d2f3a2

alibi_detect/saving/loading.py
```diff
@@ -186,8 +182,10 @@ def _load_kernel_config(cfg: dict, backend: str = Framework.TENSORFLOW) -> Calla
     """
     if backend == Framework.TENSORFLOW:
         kernel = load_kernel_config_tf(cfg)
-    else:
+    elif backend == Framework.PYTORCH:
         kernel = load_kernel_config_pt(cfg)
+    else:  # backend=='keops'
+        kernel = load_kernel_config_ke(cfg)
```
mauicv, line 188: Not a massive deal but rather than:

```py
if backend == Framework.TENSORFLOW:
        kernel = load_kernel_config_tf(cfg)
    elif backend == Framework.PYTORCH:
        kernel = load_kernel_config_pt(cfg)
    else:  # backend=='keops'
        kernel = load_kernel_config_ke(cfg)
```

we might be able to do:

```py
return {
    Framework.TENSORFLOW: load_kernel_config_tf,
    Framework.PYTORCH: load_kernel_config_pt,
    Framework.KEOPS: load_kernel_config_ko
}[backend]()
```

mauicv (approved):
LGTM

alibi_detect/saving/loading.py
ascillitoe: @mauicv happy to make this change, do agree it looks nice. My only concern is if any of the `load_kernel_config_` functions needed kwarg's for whatever reason later on. Could always change it back then though...

ascillitoe:
@jklaise would you be able to look into the responses to your comments when you have a chance, please?

jklaise (approved):
LGTM

alibi_detect/saving/loading.py
mauicv: Yeah fair enough

ascillitoe: @mauicv so I can merge this now, I have added this change as a follow-up in https://github.com/SeldonIO/alibi-detect/issues/683.

PR #710, opened 2023-01-11
Replace indexing with `tf.gather` in `MMDDriftOnlineTF._update_state`

### LINKED ISSUE #706: Intermittent `IndexError` in `MMDDriftOnlineTF` with older Python (and tensorflow?) versions
## Problem

A not-seen-before error is occurring during CI [here](https://github.com/SeldonIO/alibi-detect/actions/runs/3883607868/jobs/6625111418), in the `test_save_multivariate_online_state` test function. The error is:

```
IndexError: too many indices for array: array is 2-dimensional, but 31 were indexed
```

It occurs in `MMDDriftOnlineTF._update_state()` at the line:

```python
kernel_col = self.kernel(self.x_ref[self.ref_inds], x_t)
```

For context, `self.x_ref` is `np.ndarray: (50,4)` and `self.ref_inds` is `tf.EagerTensor: (31,)`. Strangely, this error doesn't occur in the `test_save_onlinemmddrift` test function where  `self.x_ref` is `np.ndarray: (50,2)` and `self.ref_inds` is `tf.EagerTensor: (41,)`.

## Possible cause
The error only occurs in the Python 3.7 build but not the 3.10 build. The `tensorflow` version is the same in both (2.10.1), but the Python 3.7 build uses an older `numpy` version (`1.21.6` compared to `1.23.5` in 3.10 build). I suspect the error might be an obscure compatibly error between the old-ish `numpy` version and new-ish `tensorflow` version. The `self.x_ref[self.ref_inds]` operation involves indexing a ndarray (`self.x_ref`) with a tf Tensor (`self.ref_inds`). Supporting evidence is the fact that `self.x_ref[self.ref_inds.numpy()]` fixes the issue.

## Possible solutions

We could change the problematic line to `self.x_ref[self.ref_inds.numpy()]`, or alternatively `self.x_ref[self.ref_inds, :]` appears to fix things (I think this is fine since we always expect `x_ref` and `x` to have shape (`n_instances`, `n_features`), even when `n_features=1`.

ascillitoe:
Fixes #706

### DIFF
--- a/alibi_detect/cd/tensorflow/mmd_online.py
+++ b/alibi_detect/cd/tensorflow/mmd_online.py
@@ -167,7 +167,7 @@ class MMDDriftOnlineTF(BaseMultiDriftOnline):
 
     def _update_state(self, x_t: np.ndarray):  # type: ignore[override]
         self.t += 1
-        kernel_col = self.kernel(self.x_ref[self.ref_inds], x_t)
+        kernel_col = self.kernel(tf.gather(self.x_ref, self.ref_inds), x_t)
         self.test_window = tf.concat([self.test_window[(1 - self.window_size):], x_t], axis=0)
         self.k_xy = tf.concat([self.k_xy[:, (1 - self.window_size):], kernel_col], axis=1)
 


mauicv (approved):
LGTM!

PR #712, opened 2023-01-12
Add test execution times

jklaise:
For discussion see here: https://github.com/SeldonIO/alibi/pull/849#issuecomment-1380076542

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,6 +11,8 @@ addopts =
     --tb native
     -W ignore
     --cov=alibi_detect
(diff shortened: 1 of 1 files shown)


ascillitoe (approved):
Looks good to merge. Shall we open an issue to investigate some of the long run times? Or just leave it until (if) they become problematic?

jklaise:
Don't mind either way.

ascillitoe:
I vote for lets merge! Leave the investigation _for the future_ ™

PR #713, opened 2023-01-12
Update github actions `checkout` and `setup-python`

jklaise:
Updating actions due to starting to receive warnings that the older versions will be deprecated.

Same as here: https://github.com/SeldonIO/alibi/pull/856


ascillitoe (approved):
LGTM!

PR #714, opened 2023-01-12
Added `num_workers` kwarg to PyTorch `LearnedKernelDrift`

### LINKED ISSUE #611: Add `num_workers` to PyTorch/KeOps backend where relevant
Allow to change the `num_workers` argument for the torch dataloaders where relevant.

ascillitoe:
`num_workers` was included as a kwarg to the KeOps `LearnedKernelDrift` backend in #602 (`num_workers` is an option passed to the PyTorch `DataLoader`). This adds the same option to the PyTorch `LearnedKernelDrift` backend for consistency.

Resolves #611 

### TODO

- [x] Add `num_workers` to `LearnedKernelDrift` wrapper (and pop for `backend='tensorflow'`).


ascillitoe:
Need to add `num_workers` to `LearnedKernelDrift` wrapper (and pop for `backend='tensorflow'`).

ascillitoe:
Looks like there are known issues with  `num_workers>1` for PyTorch `DataLoader`'s on Windows. One possible solution is to set `persistent_workers=True` on Windows only.

ascillitoe:
@mauicv @jklaise could I get your opinion on this one please? Passing `persistent_workers=True` to the torch `DataLoader` _might_ fix the error we are seeing here, but we'd probably only want to do this for Windows since [there are downsides](https://discuss.pytorch.org/t/what-are-the-dis-advantages-of-persistent-workers/102110/2). However, I'm hesitant to start adding `if platform == "win32"` or similar into detector code (don't think we do this anywhere currently...).

A simpler alternative is to just add a note to the docstring saying to stick to `num_workers=1` on Windows.

alibi_detect/cd/keops/learned_kernel.py
```diff
@@ -101,7 +101,7 @@ def __init__(
         epochs
             Number of training epochs for the kernel. Corresponds to the smaller of the reference and test sets.
         num_workers
-            Number of workers for the dataloader.
+            Number of workers for the dataloader. Note: Setting `num_workers>0` may be unreliable on Windows.
```
arnaudvl, line 104: >1 instead of >0?

alibi_detect/cd/pytorch/learned_kernel.py
```diff
@@ -35,6 +35,7 @@ def __init__(
             batch_size: int = 32,
             preprocess_batch_fn: Optional[Callable] = None,
             epochs: int = 3,
+            num_workers: int = 4,
```
arnaudvl, line 38: Any reason for 4 as default `num_workers`?

alibi_detect/saving/schemas.py
```diff
@@ -905,6 +905,7 @@ class LearnedKernelDriftConfig(DriftDetectorConfig):
     batch_size_predict: int = 1000000
     preprocess_batch_fn: Optional[str] = None
     epochs: int = 3
+    num_workers: int = 4
```
arnaudvl, line 908: Same comment here and a bit further on `num_workers = 4`.

arnaudvl:
> @mauicv @jklaise could I get your opinion on this one please? Passing `persistent_workers=True` to the torch `DataLoader` _might_ fix the error we are seeing here, but we'd probably only want to do this for Windows since [there are downsides](https://discuss.pytorch.org/t/what-are-the-dis-advantages-of-persistent-workers/102110/2). However, I'm hesitant to start adding `if platform == "win32"` or similar into detector code (don't think we do this anywhere currently...).
> 
> A simpler alternative is to just add a note to the docstring saying to stick to `num_workers=1` on Windows.

Docstring would work for me.

alibi_detect/cd/keops/learned_kernel.py
ascillitoe: I'm saying `>0` because the default `0` means multi-processing data loading is turned off (from [torch docs](https://pytorch.org/docs/stable/data.html#single-and-multi-process-data-loading)). `>0` causes errors on Windows. We could change the description to `>=1` if that is clearer?

alibi_detect/cd/pytorch/learned_kernel.py
ascillitoe: Only to match what you had in the KeOps one:

https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/cd/keops/learned_kernel.py#L38

alibi_detect/cd/keops/learned_kernel.py
arnaudvl: Ok got it, that's fine then. Was confused because initially in the PR it said issues were caused on windows > 1.

alibi_detect/cd/pytorch/learned_kernel.py
arnaudvl: Well that backfired quickly! Let me check why I did that.

arnaudvl: I believe I used it as `num_workers = 4 * num_gpus` looks like a reasonable default. For KeOps we could assume (with a bit of a stretch) that people use it with GPU (which is the whole point). However this is less the case for the normal PyTorch one. Maybe we should change them all to the PyTorch dataloader's default 0?

alibi_detect/cd/keops/learned_kernel.py
ascillitoe: Yeh sorry noticed my typo after you commented 🙈

alibi_detect/cd/pytorch/learned_kernel.py
ascillitoe: Good idea, I'll change default to `0` for all.

alibi_detect/cd/keops/learned_kernel.py
ascillitoe: Updated default to `num_workers=0` in ffe1b78

alibi_detect/cd/pytorch/learned_kernel.py
ascillitoe: Updated default to `num_workers=0` in ffe1b78

alibi_detect/saving/schemas.py
ascillitoe: Updated default to `num_workers=0` in ffe1b78

mauicv:
> @mauicv @jklaise could I get your opinion on this one please? Passing `persistent_workers=True` to the torch `DataLoader` _might_ fix the error we are seeing here, but we'd probably only want to do this for Windows since [there are downsides](https://discuss.pytorch.org/t/what-are-the-dis-advantages-of-persistent-workers/102110/2). However, I'm hesitant to start adding `if platform == "win32"` or similar into detector code (don't think we do this anywhere currently...).
> 
> A simpler alternative is to just add a note to the docstring saying to stick to ~`num_workers=1`~ `num_workers=0` (see [#714 (comment)](https://github.com/SeldonIO/alibi-detect/pull/714#discussion_r1084133395)) on Windows.

I'm in favour of a note in the docstring and maybe something in the docs.

mauicv (approved):
LGTM!

arnaudvl (approved):

PR #725, opened 2023-01-19
Replace Python 2 type comments with type annotations

ascillitoe:
This PR replaces Python 2 type comments with type annotations. This will allow CI to pass with `flake8 >= 6.0` (https://github.com/SeldonIO/alibi-detect/pull/682).

**Note**: Will need to do the same with `alibi`. The command `grep -r "# type:" alibi/* | grep -Eiv "ignore" can be used to find all typing comments whilst excluding `# type: ignore` comments.

### DIFF
--- a/alibi_detect/od/mahalanobis.py
+++ b/alibi_detect/od/mahalanobis.py
@@ -1,7 +1,7 @@
 import logging
 import numpy as np
 from scipy.linalg import eigh
-from typing import Dict, Union
+from typing import Dict, Optional
 from alibi_detect.utils.discretizer import Discretizer
 from alibi_detect.utils.distance import abdm, mvdm, multidim_scaling
 from alibi_detect.utils.mapping import ohe2ord, ord2num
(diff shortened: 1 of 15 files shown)


alibi_detect/od/mahalanobis.py
```diff
@@ -63,10 +63,10 @@ def __init__(self,
         # keys = categorical columns; values = numerical value for each of the categories
         self.cat_vars = cat_vars
         self.ohe = ohe
-        self.d_abs = {}  # type: Dict
+        self.d_abs: Dict = {}
 
         # initial parameter values
-        self.clip = None  # type: Union[None, list]
+        self.clip: Union[None, list] = None
```
jklaise, line 69: Use the opportunity to convert to `Optional[list]`?

jklaise (approved):

alibi_detect/od/mahalanobis.py
ascillitoe: Done in [9c910ac](https://github.com/SeldonIO/alibi-detect/pull/725/commits/9c910ac80265f21097499c6c3207cb12b70997af)

PR #739, opened 2023-02-07
Refactor to support TensorFlow 2.11

ascillitoe:
This PR includes a number of refactorings necessary for compatibility with [TensorFlow 2.11](https://github.com/tensorflow/tensorflow/releases/tag/v2.11.0). 

The majority of changes are to support the new family of optimizers introduced in 2.11, whilst there is one minor change to `MMDDriftOnlineTF` due to a change in the behavior of `tf.concat`.

# Compatibility with new optimizers

TensorFlow 2.11 introduces a [new optimizers API](https://blog.tensorflow.org/2022/11/whats-new-in-tensorflow-211.html#keras:~:text=The%20new%20Keras%20Optimizers%20API%20is%20ready). The optimizers largely behave the same as the legacy optimizers (now located at `tf.keras.optimizers.legacy`). However, the way they handle their own state has changed, which causes problems with a number of Alibi Detect's detectors (see below). 

A problem arises when the new optimizers are asked to optimize different models. A "different model" may mean an entirely different model, or a clone of the same model. For example:

```python
from alibi_detect.models.tensorflow import trainer
from alibi_detect.utils.tensorflow.misc import clone_model

# Define optimizer
optimizer = tf.keras.optimizers.Adam()

# Define original model
model = MyModel()

# Train model
trainer(model, optimizer)

# Clone model
model = clone_model(model)
trainer(model, optimizer)  # Leads to error!
```

This will raise an error like:

```
KeyError: 'The optimizer cannot recognize variable dense_1733/kernel:0. This usually means you are trying to call the optimizer to update different parts of the model separately. Please call `optimizer.build(variables)` with the full list of trainable variables before the training loop or use legacy optimizer `tf.keras.optimizers.legacy.{self.__class__.__name__}.'
```

See [this notebook](https://gist.github.com/ascillitoe/d6c9a8797c09491dab16f66aeaf7729c) for an applied example of this problem.

## `ClassifierDrift` detector

The above issue breaks the `ClassifierDriftTF` detector because `self.model` is cloned on each `predict` call when `retrain_from_scratch=True` (in order to reinitialise the model). Effectively, `self.model` is no longer the same model the `optimizer` saw during the last `predict` call when `optimizer.apply_gradients` was called in `trainer`. 

### Possible fixes

1. The error message suggests to call `optimizer.build(variables)` prior to using the `optimizer` (where `variables` is the variables to be optimized). However, this doesn't appear to work when `optimizer` is used on a cloned model (the variables in the cloned model appear to have the same names as previously. It isn't clear what changes in the cloned model lead to the optimizer raising an error). If this issue specific to cloned models was fixed in a future tf version, an `optimizer.build()` call in `trainer` might be enough.
2. Call `optimizer = optimizer.__class__.from_config(optimizer.get_config())` before training. This fixes the issue because the `optimizer` is effectively a fresh clone during each `predict` call.
3. Avoid cloning `self.model` during `predict`. Instead, reinitialise `self.model` to the original model by getting/setting model weights. This fixes the problem since `self.model`'s state is changed upon each `self.predict` call, but it is essentially still the same model. This would also fix https://github.com/SeldonIO/alibi-detect/issues/740.

### Chosen fix

Option 2. has currently been chosen due to its simplicity...

## `SpotTheDiff` detector

For this detector, the `self.Model` is defined as a `SpotTheDiffDriftTF.InterpretableClf` model internally, before passing it to `ClassifierDriftTF`. The model therefore now has to be called on sample data internally, before it is passed to `ClassifierDriftTF`.

### Outlier and adversarial detectors

For most of the outlier and adversarial detectors, the `optimizer` kwarg was set as:

```python
optimizer: tf.keras.optimizers = tf.keras.optimizers.Adam(learning_rate=1e-3)
```

This was a problem because the `optimizer` was being instantiated in the default kwarg when a detector was first imported (not instantiated!), meaning detectors would be broken when instantiated multiple times (since new detectors and models all used the same optimizer instantiated at import).


alibi_detect/cd/tensorflow/spot_the_diff.py
```diff
@@ -129,7 +129,14 @@ def __init__(
             if len(initial_diffs) != n_diffs:
                 raise ValueError("Should have initial_diffs.shape[0] == n_diffs")
 
+        # Define interpretable classifier model
         model = SpotTheDiffDriftTF.InterpretableClf(kernel, x_ref_proc, initial_diffs)
+        # Build the model so that it can be passed to ClassifierDriftTF
+        model(x_ref_proc)
```
ascillitoe, line 135: Performance-wise, it would be preferable to do `model.build(x_ref_proc.shape)` instead of `model(x_ref_proc)`. However, `build` doesn't currently work due to the design of `GaussianRBF`. It would probably also be too restrictive on user-provided custom `kernel`'s.

alibi_detect/cd/tensorflow/mmd_online.py
```diff
@@ -168,7 +168,7 @@ def _configure_thresholds(self):
                     2 * tf.reduce_sum(k_xy_col_sums[w:w + w_size]))
                     for k_xx_sum, y_inds_w, k_xy_col_sums in zip(k_xx_sums_all, y_inds_all_w, k_xy_col_sums_all)
                     ]
-            mmds = tf.concat(mmds, axis=0)  # an mmd for each bootstrap sample
+            mmds = tf.stack(mmds, axis=0)  # an mmd for each bootstrap sample
```
ascillitoe, line 171: To fix the following error:

```python
Traceback (most recent call last):
  File "/home/runner/work/alibi-detect/alibi-detect/alibi_detect/cd/tests/test_mmd_online.py", line 25, in test_mmddriftonline
    cd = MMDDriftOnline(x_ref=x_ref, ert=25, window_size=5, backend=backend, n_bootstraps=100)
  File "/home/runner/work/alibi-detect/alibi-detect/alibi_detect/cd/mmd_online.py", line 99, in __init__
    self._detector = MMDDriftOnlineTF(*args, **kwargs)  # type: ignore
  File "/home/runner/work/alibi-detect/alibi-detect/alibi_detect/cd/tensorflow/mmd_online.py", line 87, in __init__
    self._configure_thresholds()
  File "/home/runner/work/alibi-detect/alibi-detect/alibi_detect/cd/tensorflow/mmd_online.py", line 171, in _configure_thresholds
    mmds = tf.concat(mmds, axis=0)  # an mmd for each bootstrap sample
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/site-packages/tensorflow/python/util/traceback_utils.py", line [153](https://github.com/SeldonIO/alibi-detect/pull/739/checks#step:8:154), in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/opt/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/site-packages/tensorflow/python/framework/ops.py", line 7215, in raise_from_not_ok_status
    raise core._status_to_exception(e) from None  # pylint: disable=protected-access
tensorflow.python.framework.errors_impl.InvalidArgumentError: {{function_node __wrapped__ConcatV2_N_100_device_/job:localhost/replica:0/task:0/device:CPU:0}} ConcatOp : Can't concatenate scalars (use tf.stack instead) [Op:ConcatV2] name: concat
```

This has only begun occuring in tf 2.11. Prior to reaching the `tf.concat` call, `mmds` is a list of `tf.Tensor`'s each with `shape=()` (i.e. scalars!). The error message "_ConcatOp : Can't concatenate scalars_" therefore makes some sense. However, I cannot find anything in the 2.11 release notes about the behavior of `tf.concat` changing...

alibi_detect/ad/adversarialae.py
```diff
@@ -139,7 +139,7 @@ def fit(self,
             loss_fn: tf.keras.losses = loss_adv_ae,
             w_model: float = 1.,
             w_recon: float = 0.,
-            optimizer: tf.keras.optimizers = tf.keras.optimizers.Adam(learning_rate=1e-3),
+            optimizer: tf.keras.optimizers.Optimizer = tf.keras.optimizers.Adam,
```
ascillitoe, line 142: Avoid instantiation in default kwarg, since this meant `optimizer` was instantiated upon detector import instead of instantiation. See [PR discussion](https://github.com/SeldonIO/alibi-detect/pull/739#:~:text=to%20ClassifierDriftTF.-,Outlier%20and%20adversarial%20detectors,-For%20most%20of).

ascillitoe: Note: the type signature needs to be updated here, but we need to discuss the best way to do this since it could get messy. Technically the type is now `Union[Type[tf.keras.optimizers.Optimizer], tf.keras.optimizers.Optimizer, Type[tf.keras.optimizers.legacy.Optimizer], tf.keras.optimizers.legacy.Optimizer]`

alibi_detect/cd/tensorflow/classifier.py
```diff
@@ -125,16 +125,21 @@ def __init__(
...
         if preds_type not in ['probs', 'logits']:
             raise ValueError("'preds_type' should be 'probs' or 'logits'")
 
         self.meta.update({'backend': Framework.TENSORFLOW.value})
 
         # define and compile classifier model
-        self.original_model = model
-        self.model = clone_model(model)
+        self.model: tf.keras.Model = model
+        if not self.model.built:
```
ascillitoe, line 137: Check if model has been built. See discussion in PR...

alibi_detect/cd/tests/test_spot_the_diff.py
```diff
@@ -57,7 +57,7 @@ def test_stddrift(stddrift_params):
         kernel = MyKernelTF((n_features,))
     else:
         kernel = None
-    x_ref = np.random.randn(*(n, n_features))
+    x_ref = np.random.randn(*(n, n_features)).astype(np.float32)
```
ascillitoe, line 60: `x_ref` was not previously used during detector instantiation, but now is used to build the model internally. The model expects float32 data...

alibi_detect/saving/_tensorflow/tests/test_saving_tf.py
```diff
@@ -5,27 +5,30 @@
...
 from alibi_detect.saving.schemas import ModelConfig
 import tensorflow as tf
+import numpy as np
 
 backend = param_fixture("backend", ['tensorflow'])
 
 
 # Note: The full save/load functionality of optimizers (inc. validation) is tested in test_save_classifierdrift.
-def test_load_optimizer_object(backend):
+@parametrize('legacy', [True, False])
```
ascillitoe, line 17: Test saving/loading of new and legacy optimizers.

ascillitoe:
**Note re 'Custom>'**: The output of [tf.keras.optimizers.serialize](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/serialize) for the new tf 2.11 optimizers is currently "strange":

```python
tf.keras.optimizers.serialize(tf.keras.optimizers.SGD())
```

gives:

```
{'class_name': 'Custom>SGD',
 'config': {'name': 'SGD',
  'weight_decay': None,
  'clipnorm': None,
  'global_clipnorm': None,
  'clipvalue': None,
  'use_ema': False,
  'ema_momentum': 0.99,
  'ema_overwrite_frequency': None,
  'jit_compile': True,
  'is_legacy_optimizer': False,
  'learning_rate': 0.01,
  'momentum': 0.0,
  'nesterov': False}}
```

whereas the legacy optimizer gives `{'class_name': 'Custom>SGD', ...`. This is likely an oversight and will presumably be updated in a future tf version. It doesn't impact our serialisation but the test would need updating...

setup.py
```diff
@@ -19,7 +19,7 @@ def readme():
     # https://github.com/SeldonIO/alibi-detect/issues/375 and 387
     "tensorflow": [
         "tensorflow_probability>=0.8.0, <0.19.0",
-        "tensorflow>=2.2.0, !=2.6.0, !=2.6.1, <2.11.0",  # https://github.com/SeldonIO/alibi-detect/issues/375 and 387
+        "tensorflow>=2.2.0, !=2.6.0, !=2.6.1, <2.12.0",  # https://github.com/SeldonIO/alibi-detect/issues/375 and 387
```
ascillitoe, line 22: Supersedes #678

alibi_detect/od/llr.py
```diff
@@ -144,6 +144,10 @@ def fit(self,
             Callbacks used during training.
         """
         input_shape = X.shape[1:]
+        optimizer = optimizer() if isinstance(optimizer, type) else optimizer
```
ascillitoe, line 147: This detector is a special case. The `optimizer` is called on two separate models during `fit`, which now leads to errors. To fix the `optimizer` is cloned so that separate optimizers are used for each model.

alibi_detect/od/llr.py
```diff
@@ -144,6 +144,10 @@ def fit(self,
...
+        # Separate into two separate optimizers, one for semantic model and one for background model
+        optimizer_s = optimizer
+        optimizer_b = optimizer.__class__.from_config(optimizer.get_config())
```
ascillitoe, line 150: This detector is a special case. The optimizer is called on two separate models during fit, which now leads to errors. To fix the optimizer is cloned so that separate optimizers are used for each model.

alibi_detect/cd/tensorflow/classifier.py
```diff
@@ -125,16 +125,21 @@ def __init__(
...
         )
         if preds_type not in ['probs', 'logits']:
             raise ValueError("'preds_type' should be 'probs' or 'logits'")
 
         self.meta.update({'backend': Framework.TENSORFLOW.value})
 
         # define and compile classifier model
-        self.original_model = model
-        self.model = clone_model(model)
+        self.model: tf.keras.Model = model
```
ascillitoe, line 136: This change removes the public attribute `self.original_model`. This is still present in the scikit-learn and PyTorch backends. We might want to make `original_model` private in those detectors, although there are many things that are public but should be private, so leaving this for another PR. Below are some arguments for and against keeping a public `original_model` attribute.


- **Arguments for keeping**: The `model` provided to the detector will be trained during detector predictions. `detector.original_model` provides the user with access to their original model (if public attribute). If private, the original model might still be useful in the future e.g. if adding explanations functionality to the sklearn backend (based on discussion w/ @RobertSamoilescu, this was part of the reason for including `original_model` in the sklearn backend). 
- **Arguments for removing**: Unnecessary complexity. Would be good to remove all `model_clone`'s eventually (issues with subclassed models etc). Duplication of model (although weights need duplicating anyway...). If the user cares about the original model, they could just save (or clone) it before passing it to the detector. The attribute is currently undocumented so may as well be private...

alibi_detect/cd/tensorflow/classifier.py
```diff
@@ -176,8 +176,13 @@ def score(self, x: np.ndarray) -> Tuple[float, float, np.ndarray, np.ndarray,  #
...
             ds_tr = self.dataset(x_tr, y_tr)
-            self.model = clone_model(self.original_model) if self.retrain_from_scratch \
-                else self.model
+            if self.retrain_from_scratch:
+                # clone model to re-initialise
+                self.model = clone_model(self.original_model)
+                # If a new (tf>=2.11) optimizer, clone this too to prevent error due to cloned model
+                optimizer = self.train_kwargs['optimizer']
+                if isinstance(optimizer, tf.keras.optimizers.Optimizer):  # If not a legacy optimizer
+                    self.train_kwargs['optimizer'] = optimizer.__class__.from_config(optimizer.get_config())
```
ascillitoe, line 185: This is "option 2" of the possible fixes. Cloning the `optimizer` as well as the `model` prevents it from complaining about the `model` changing on successive `predict` calls.

jklaise:
It might be worth investigating the pros and cons of avoiding `model.clone` calls entirely (and reset weights instead as you mention).

ascillitoe:
> It might be worth investigating the pros and cons of avoiding `model.clone` calls entirely (and reset weights instead as you mention).

Sorry I did actually investigate this and meant to add more details. Prior to https://github.com/SeldonIO/alibi-detect/pull/739/commits/d441a293336dc6ec99d6e3888827a3332fb6939a I had gone with option 3 (getting/setting weights). The nice thing about this option is that it resolves #740 at the same time. However, the problem with it was that subclassed models that have not yet been built return a list of zero length when `get_weights` is called, and this then causes an error when it is given to `set_weights` in `predict`. A solution is to require all subclassed models to be built (via `.build` or calling the model on sample data) before passing them to the `ClassifierDrift` detector. Prior to  https://github.com/SeldonIO/alibi-detect/pull/739/commits/d441a293336dc6ec99d6e3888827a3332fb6939a I'd added a warning to the docs setting out this expectation. I had two concerns with this:

1. It adds different expectations for models in different places e.g. models passed to other detectors and `preprocess_fn` don't need to be called. 
2. Many models don't have a workable `.build` method, meaning it has to be called on dummy data instead. This isn't always easy to do. For example, when I tried to retrofit the Amazon example notebook, I realised it was a little tricky to do this because I'd have to extract the tokenizer from the batch preprocess function, pass the dummy data through that, and then pass the tokens through the `model` to build it. We have a similar requirement that subclassed models have to be built before `save_detector` will work (see #628 ), but this is easier to do as the detector containing the `model` will have already been instantiated so a `detector.predict` call will build the model.

alibi_detect/ad/adversarialae.py
jklaise: We should be getting some `mypy` errors from tests in that case? (in fact the old type doesn't make sense...)

alibi_detect/od/llr.py
jklaise: I assume we don't have to worry about potentially sharing optimizer state between the two models or diverging optimizer states @arnaudvl ?

arnaudvl: No they don't have to share state.

alibi_detect/cd/tensorflow/mmd_online.py
ascillitoe: Yes, at least for `tensorflow >= 2.2.0` which is the lower bound we support. The difference between the two is that `tf.concat` only works when `axis` is within the existing dimensions, whilst `tf.stack` creates a new dimension when `axis` is outside the existing dimensions. I think `stack` makes more sense here since we're concatenating scalars with zero dimensions into a tensor with dimension `(n,)`. In other words, it seems that 2.11 has simply made `tf.concat` more strict about "misuse", rather than it changing the behavior of `stack` or `concat`.

alibi_detect/ad/adversarialae.py
ascillitoe: Yeh the old types don't make sense, but this is a wider issue with typing of  `kernel`, `optimizer`, `dataset` at maybe other kwargs, so think it makes sense to leave the fix to a follow-up>

alibi_detect/cd/tensorflow/classifier.py
```diff
@@ -176,8 +176,13 @@ def score(self, x: np.ndarray) -> Tuple[float, float, np.ndarray, np.ndarray,  #
...
                 raise TypeError(f'x needs to be of type np.ndarray or list and not {type(x)}.')
             ds_tr = self.dataset(x_tr, y_tr)
-            self.model = clone_model(self.original_model) if self.retrain_from_scratch \
-                else self.model
+            if self.retrain_from_scratch:
+                # clone model to re-initialise
+                self.model = clone_model(self.original_model)
+                # If a new (tf>=2.11) optimizer, clone this too to prevent error due to cloned model
+                optimizer = self.train_kwargs['optimizer']
+                if isinstance(optimizer, tf.keras.optimizers.Optimizer):  # If not a legacy optimizer
```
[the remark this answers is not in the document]
ascillitoe: Thanks, you're right this didn't make sense. I haven't found a sensible backward compatible way to check new vs old optimizers. The `tf.keras.optimizers.legacy` and `tf.keras.optimizers.experimental` submodules don't exist in much older tf versions, and it isn't clear what will happen to these in the future anyway. To keep things simple I've just removed the check so that we clone old optimizers too. I've just checked with tf 2.2.0 and it works OK.

alibi_detect/ad/adversarialae.py
ascillitoe: p.p.s on second thoughts, the `optimizer` typing issue is introduced by this PR, so we should fix it here...

The question is how to do this, the new `Union` would be quite ugly, so maybe we should declare the `Union` in `alibi_detect.utils._types` instead?

jklaise: Yea, I would agree with having a well-named type alias.

jklaise: Since `tensorflow` is optional, there might be some issue with having these types transparently available...

alibi_detect/cd/tests/test_spot_the_diff.py
```diff
@@ -57,7 +57,7 @@ def test_stddrift(stddrift_params):
         kernel = MyKernelTF((n_features,))
     else:
         kernel = None
-    x_ref = np.random.randn(*(n, n_features)).astype(np.float32)
+    x_ref = np.random.randn(*(n, n_features))
```
jklaise, line 60: ?

alibi_detect/cd/tensorflow/classifier.py
```diff
@@ -51,8 +52,7 @@ def __init__(
         x_ref
             Data used as reference distribution.
         model
-            TensorFlow classification model used for drift detection. If this is a subclassed model, it should be
-            built/called before passing it to the detector.
+            TensorFlow classification model used for drift detection.
```
jklaise, line 55: I don't fully follow, can you clone if it's not built? I thought that was the original issue why the model needed to be built.

alibi_detect/cd/tests/test_spot_the_diff.py
ascillitoe: I added `astype(np.float32)` because this was necessary when I added a dummy model call in `SpotTheDiffDrift` to build the model before passing to `ClassifierDrift` (prior to [d441a29](https://github.com/SeldonIO/alibi-detect/pull/739/commits/d441a293336dc6ec99d6e3888827a3332fb6939a)). We're no longer doing any dummy calls because we've reverted to cloning the model instead of getting/setting weights.

Technically the data should be float32 if we're actually going to call the model or detector, but we don't do that in this test. Just trying to keep unrelated changes to a minimum in the PR...

alibi_detect/cd/tensorflow/classifier.py
ascillitoe: There are two main limitations with subclassed models:

1. Cloning is a problem if the correct `custom_objects` were not passed during loading. See discussion in https://github.com/SeldonIO/alibi-detect/pull/628.
2. `get_weights()` will return a length zero list which causes `set_weights()` to fail, if the model has not been built. 

Removing the cloning in this PR and replacing with getting/setting weights would resolve the problem in https://github.com/SeldonIO/alibi-detect/pull/628, in that users would be able to instantiate and run detectors without worrying about providing `custom_objects` at load time. However, due to 2. above, it would add a different limitation, in that models would have to be built before passing to the detector, so that the getting/setting of weights would work. On balance, I think the latter (having to build the model) is more of a limitation than the former (providing custom objects). Although this is subjective...

ascillitoe: P.s. the above change changes the docstring back to what it is in `master`

alibi_detect/ad/adversarialae.py
ascillitoe: > Since `tensorflow` is optional, there might be some issue with having these types transparently available...

I've added a type alias in [53bd5f1](https://github.com/SeldonIO/alibi-detect/pull/739/commits/53bd5f1472e266d349f3caaf89b321cc38624a68), with postponed annotations so things still work when `tensorflow` isn't installed. Is this what you're referring to?

alibi_detect/saving/_tensorflow/tests/test_saving_tf.py
```diff
@@ -1,26 +1,78 @@
...
+import numpy as np
+from packaging import version
 
 backend = param_fixture("backend", ['tensorflow'])
 
 
 # Note: The full save/load functionality of optimizers (inc. validation) is tested in test_save_classifierdrift.
-def test_load_optimizer_object(backend):
+@pytest.mark.skipif(version.parse(tf.__version__) < version.parse('2.11.0'),
+                    reason="Skipping since tensorflow < 2.11.0")
```
ascillitoe, line 20: @jklaise this is admittedly a bit weird. The new testing logic assumed tf 2.11 was being used, so tests would have failed on old tensorflow versions. I've added two separate tests for tf>=2.11 and tf<2.11, so that we can test for backward compatibility with old tensorflow versions when desired, but this would have to be by manually installing tf<2.11 locally, since CI will install the latest version. 

Since we don't test old versions of deps anywhere else, I'm wondering if this is overkill, or if there's no harm in keeping both tests...

alibi_detect/utils/_types.py
```diff
@@ -4,27 +4,36 @@
...
 if has_pytorch:
     import torch
     supported_models_torch = (torch.nn.Module, )
     supported_optimizers_torch = (type, )  # Note type not object!
 supported_models_sklearn = (BaseEstimator, )
 supported_models_all = supported_models_tf + supported_models_torch + supported_models_sklearn
 supported_optimizers_all = supported_optimizers_tf + supported_optimizers_torch
+
+# type aliases, for use with mypy (must be FwdRef's if involving opt. deps.)
+Optimizer_tf: TypeAlias = Union['tf.keras.optimizers.Optimizer', 'tf.keras.optimizers.legacy.Optimizer',
```
jklaise, line 38: This is really a nitpick, but the convention for types is usually `PascalCase`, e.g. `OptimizerTF` (or perhaps `TFOptimizer`).

alibi_detect/saving/_tensorflow/tests/test_saving_tf.py
jklaise: I see, I think it's fine to leave this in so we can debug locally with older `tensorflow` versions if required.

alibi_detect/utils/_types.py
ascillitoe: Changed in [f819e63](https://github.com/SeldonIO/alibi-detect/pull/739/commits/f819e63981425c60252fc4be0b3cdea9e00fcc3a)!

alibi_detect/saving/_tensorflow/saving.py
```diff
@@ -158,7 +158,7 @@ def save_embedding_config(embed: TransformerEmbedding,
     return cfg_embed
 
 
-def save_optimizer_config(optimizer: tf.keras.optimizers.Optimizer):
+def save_optimizer_config(optimizer: Union[tf.keras.optimizers.Optimizer]):
```
jklaise, line 161: Unnecessary `Union`?

jklaise (approved):
LGTM!

alibi_detect/saving/_tensorflow/saving.py
ascillitoe: Changed to `Union[tf.keras.optimizers.Optimizer, tf.keras.optimizers.legacy.Optimizer]` in https://github.com/SeldonIO/alibi-detect/pull/739/commits/a3f0e90e593420c391a1638aa562f67ce869ce5d

PR #736, opened 2023-01-31
Feature gmm od

mauicv:
## What is this

Implementation of GMM outlier detector. See this [notebook](https://colab.research.google.com/gist/mauicv/3ce5f2a40f08d5d3aedcd8aa712e4acb/pca-detector-pr728ipynb.ipynb#scrollTo=f-yXBFS797g0&uniqifier=3) for example. Branches off [pca-od](https://github.com/SeldonIO/alibi-detect/pull/728).

## TODO:

- [x] sklearn backend
- [x] pytorch backend
- [x] docs
- [x] tests
- [x] example notebook
- [x] mypy
- [x] flake8
- [x] Optional deps
- [x] Merge kNN `fit`/`infer_threshold` changes

Issue #567, opened 2022-07-25
Review content of metadata in detector config files


ascillitoe:
Our detector `self.meta` attribute currently contains the following information:

```python
DEFAULT_META = {
    "name": None,
    "online": None,  # true or false
    "data_type": None,  # tabular, image or time-series
    "version": None,
    "detector_type": None  # drift, outlier or adversarial
}  # type: Dict
```

Whereas the metadata in detector config files contains:

```toml
[meta]
version = "0.10.0rc1"
config_spec = "0.1"
version_warning = false
```

We should review what content it makes sense to include in detector config files. Fields such as `detector_type` and `online` could be useful to view without instantiating a detector. In https://github.com/SeldonIO/alibi-detect/pull/564 it was decided not to include the entirety of `self.meta` in config files since this leads to excessive duplication (e.g. `name`, `backend` and `data_type`).

ascillitoe:
We should also review the intention behind the `data_type` kwarg, and whether it is even needed.

PR #746, opened 2023-02-21
Feature LOF outlier detector

mauicv:
## What is this

Implementation of lof outlier detector. See this [notebook](https://colab.research.google.com/gist/mauicv/7400a79643ec5807db41b3ef1ad9a94f/pca-detector-pr728ipynb.ipynb#scrollTo=zexT7qz-ucED) for example. Branches off [gmm-od](https://github.com/SeldonIO/alibi-detect/pull/736).

## TODO:

- [x] pytorch backend
- [x] docs
- [x] tests
- [x] example notebook
- [x] mypy
- [x] flake8
- [x] Optional deps
- [x] Merge kNN `fit`/`infer_threshold` changes


alibi_detect/od/_lof.py
```diff
@@ -0,0 +1,216 @@
...
+        The LOF detector is a non-parametric method for outlier detection. It computes the local density
+        deviation of a given data point with respect to its neighbors. It considers as outliers the
+        samples that have a substantially lower density than their neighbors.
+
+        The detector can be initialized with `k` a single value or an array of values. If `k` is a single value then
+        the score method uses the distance/kernel similarity to the k-th nearest neighbor. If `k` is an array of
+        values then the score method uses the distance/kernel similarity to each of the specified `k` neighbors.
+        In the latter case, an `aggregator` must be specified to aggregate the scores.
+
+        Note that, in the multiple k case, a normalizer can be provided. If a normalizer is passed then it is fit in
```
ojcobb, line 48: I think we should be a little clearer as to what the normalizer and aggregator refer to as it's not clear here or from the kwarg descriptions. I realise this applied to KNN too.

alibi_detect/od/_lof.py
```diff
@@ -0,0 +1,216 @@
...
+            Reference data used to fit the detector.
+        """
+        self.backend.fit(self.backend._to_tensor(x_ref))
+
+    @catch_error('NotFittedError')
+    @catch_error('ThresholdNotInferredError')
+    def score(self, x: np.ndarray) -> np.ndarray:
+        """Score `x` instances using the detector.
+
+        Computes the local outlier factor for each instance in `x`. If `k` is an array of values then the score for
```
ojcobb, line 128: Maybe worth just noting here that the outlier factor is the density of each instance in `x` relative to those of its neighbours in `x_ref`.

alibi_detect/od/pytorch/lof.py
```diff
@@ -0,0 +1,164 @@
...
+        If k is an array then we need to compute the average reachability for each k separately. To do
+        this we use a mask to weight the reachability of each k-close neighbor by 1/k and the rest to 0.
+        """
+        mask = torch.zeros_like(reachabilities[0], device=self.device)
+        for i, k in enumerate(self.ks):
+            mask[:k, i] = torch.ones(k, device=self.device)/k
+        return mask
+
+    def _compute_K(self, x, y):
+        """Compute the distance/similarity matrix matrix between `x` and `y`."""
```
ojcobb, line 81: Could we remove "/similarity" here? A similarity matrix would have entried that icnrease with similarity, whereas this is the opposite

ojcobb:
Ran through all the code and couldn't find anything worth commenting on! Just the nitpicks on docstrings above. Nice!

ascillitoe:
> couldn't find anything worth commenting on

Challenge accepted 😛

alibi_detect/od/_lof.py
```diff
@@ -0,0 +1,216 @@
...
+            ensembler = ensembler_cls(
+                normalizer=get_normalizer(normalizer),
+                aggregator=get_aggregator(aggregator)
+            )
+
+        self.backend = backend_cls(k, kernel=kernel, ensembler=ensembler, device=device)
+
+        # set metadata
+        self.meta['detector_type'] = 'outlier'
+        self.meta['data_type'] = 'numeric'
```
ascillitoe, line 110: Not isolated to this PR<, but noting that we seem to be a little inconsistent across the new and old outlier detectors wrt to when `data_type` is hard-coded, and when it is optionally set via a kwarg. For some, it is hardcoded to `time-series` (which makes sense), for some (e.g. the old `Mahalanobis`) it is set via kwarg, and for some it is hard coded to `numeric`. Maybe worth opening an issue to review this more generally? 

Already mentioned in https://github.com/SeldonIO/alibi-detect/issues/567#issuecomment-1193853273, but highlighting here since we are setting `data_type` in new detectors too...

alibi_detect/od/_lof.py
```diff
@@ -0,0 +1,216 @@
...
+    def score(self, x: np.ndarray) -> np.ndarray:
+        """Score `x` instances using the detector.
+
+        Computes the local outlier factor for each instance in `x`. If `k` is an array of values then the score for
+        each `k` is aggregated using the ensembler.
+
+        Parameters
+        ----------
+        x
+            Data to score. The shape of `x` should be `(n_instances, n_features)`.
```
ascillitoe, line 134: Question, asking because currently discussing this wrt to drift detectors. If `n_features=1`, does `(n_instances,)` work, or must it be `(n_instances, 1)`?

alibi_detect/od/_lof.py
```diff
@@ -0,0 +1,216 @@
...
+
+        Returns
+        -------
+        Outlier scores. The shape of the scores is `(n_instances,)`. The higher the score, the more anomalous the \
```
ascillitoe, line 138: Nitpick: In a few places `outlying` is used e.g. 
https://github.com/SeldonIO/alibi-detect/blob/21ca540f5b0ff8f2ee312e7a30ace0d27c238435/alibi_detect/od/_mahalanobis.py#L96

whereas in the `score` docstring (and for `_pca`, `_gmm`, `_knn`, `mahalanobis`) `anomalous` is used. Worth picking one or the other?

alibi_detect/od/tests/test__lof/test__lof_backend.py
```diff
@@ -0,0 +1,220 @@
...
+    x_ref = torch.randn((1024, 10))
+    lof_torch.fit(x_ref)
+    x = torch.randn((3, 10)) * torch.tensor([[1], [1], [100]])
+    lof_torch.infer_threshold(x_ref, 0.1)
+    outputs = lof_torch.predict(x)
+    assert torch.all(outputs.is_outlier == torch.tensor([0, 0, 1]))
+    assert torch.all(lof_torch(x) == torch.tensor([0, 0, 1]))
+
+
+@pytest.mark.skip(reason="Can't convert GaussianRBF to torch script due to torch script type constraints")
```
ascillitoe, line 118: What is the intention behind including this test if it is skipped in all cases?

ascillitoe (approved):
Few minor comments, otherwise LGTM!

alibi_detect/od/_lof.py
mauicv: It should be `(n_instances, 1)` which I think is the best approach here. I think it's best to have a simple consistent behaviour and let the user manage shapes.

alibi_detect/od/tests/test__lof/test__lof_backend.py
mauicv: I wrote it then decided not to include this functionality in the first set of outliers, however, it should be implemented at some point which is why I've left it in but skipped it. I can take it out if you prefer.

ojcobb (approved):

alibi_detect/od/_lof.py
ascillitoe: Agreed!

PR #752, opened 2023-03-03
Fix two bugs w/ kwargs and `preprocess_batch_fn` in `preprocess_fn` serialisation

ascillitoe:
This PR fixes two bugs in serialisation/deserialisation of the drift detector `preprocess_fn`:

1. kwargs in a `partial` to be serialised to `.dill` are now named according their kwarg name i.e. `preprocess_batch_fn` is serialised to `preprocess_fn/preprocess_batch_fn.dill`. Previously, this would have been saved to `preprocess_fn.dill`. This was loaded fine so not spotted in tests, but there would have been a problem if multiple kwargs in a `partial` were to be saved, since they'd all have the same name...
2. When `preprocess_fn` was a custom Python function with kwarg's (i.e. NOT `preprocess_drift`), the kwargs were not serialised. This has now been fixed and a test added to check for a such a bug in the future.

@jklaise @mauicv assuming we are ok to include the `mypy`, `flake8` and docs being built w/ 3.10 changes in the patch, we could merge this directly into `master` and release off that, rather than `patch/v0.11.1`?

## TODO: 

- [x] CHANGELOG


alibi_detect/saving/loading.py
```diff
@@ -253,7 +253,10 @@ def _load_preprocess_config(cfg: dict) -> Optional[Callable]:
         logger.warning('Unable to process preprocess_fn. No preprocessing function is defined.')
         return None
 
-    return partial(preprocess_fn, **kwargs)
+    if kwargs == {}:
+        return preprocess_fn
```
ascillitoe, line 257: Previously the preprocess function was wrapped in a `partial` even if there were no kwarg's. Now the bare function is returned in this case.

alibi_detect/saving/saving.py
```diff
@@ -292,7 +292,7 @@ def _save_preprocess_config(preprocess_fn: Callable,
 
         # Arbitrary function
         elif callable(v):
-            src, _ = _serialize_object(v, filepath, local_path)
+            src, _ = _serialize_object(v, filepath, local_path.joinpath(k))
```
ascillitoe, line 295: This is bug fix 1. i.e. add the kwarg name to the serialised dill file.

alibi_detect/saving/saving.py
```diff
@@ -302,7 +302,7 @@ def _save_preprocess_config(preprocess_fn: Callable,
     if 'preprocess_drift' in func:
         preprocess_cfg.update(kwargs)
     else:
-        kwargs.update({'kwargs': kwargs})
+        preprocess_cfg.update({'kwargs': kwargs})
```
ascillitoe, line 305: This is bug fix 2. Previously the `kwargs` were updated but this had no effect. It should have been the returned `preprocess_cfg` dict that was updated instead.

alibi_detect/saving/tests/models.py
```diff
@@ -89,7 +89,7 @@ def encoder_dropout_model(backend, current_cases):
 
 
 @fixture
-def preprocess_custom(encoder_model):
+def preprocess_uae(encoder_model):
```
ascillitoe, line 92: More descriptive naming. `preprocess_custom` actually contains a uae model, so renamed `preprocess_uae`. Otherwise confusion in `test_saving.py` where we are now testing "custom" preprocessing functions.

alibi_detect/saving/tests/test_saving.py
```diff
@@ -1132,14 +1130,40 @@ def test_save_preprocess(data, preprocess_fn, tmp_path, backend):
         assert isinstance(preprocess_fn_load.keywords['model'], nn.Module)
 
 
+@parametrize('preprocess_fn', [preprocess_simple, preprocess_simple_with_kwargs])
+def test_save_preprocess_custom(preprocess_fn, tmp_path):
```
ascillitoe, line 1134: New test function to test save/load of:

1. `preprocess_fn` where it is a Python func with no kwargs.
2. `preprocess_fn` where it is a `partial` with kwargs.

alibi_detect/saving/tests/test_saving.py
```diff
@@ -1152,6 +1176,8 @@ def test_save_preprocess_nlp(data, preprocess_fn, tmp_path, backend):
     assert cfg_preprocess['src'] == '@cd.' + registry_str + '.preprocess.preprocess_drift'
     assert cfg_preprocess['embedding']['src'] == 'preprocess_fn/embedding'
     assert cfg_preprocess['tokenizer']['src'] == 'preprocess_fn/tokenizer'
+    assert tmp_path.joinpath(cfg_preprocess['preprocess_batch_fn']).is_file()
+    assert cfg_preprocess['preprocess_batch_fn'] == 'preprocess_fn/preprocess_batch_fn.dill'
```
ascillitoe, line 1180: This tests that the `preprocess_batch_fn` is serialized to disk, and has the correct name. Implicitly tests bug fix 1...

alibi_detect/saving/tests/test_saving.py
```diff
@@ -553,7 +554,7 @@ def test_save_contextmmddrift(data, kernel, backend, tmp_path, seed):  # noqa: F
     assert cd_load._detector.n_permutations == N_PERMUTATIONS
     assert cd_load._detector.p_val == P_VAL
     assert isinstance(cd_load._detector.preprocess_fn, Callable)
-    assert cd_load._detector.preprocess_fn.func.__name__ == 'preprocess_simple'
+    assert cd_load._detector.preprocess_fn.__name__ == 'preprocess_simple'
```
ascillitoe, line 557: Change required since we no longer wrap the function in a `partial` when there are no kwargs.

jklaise (approved):
LGTM! Agree with doing the patch release from `master` instead of a patch branch.

mauicv (approved):
LGTM!

Issue #676, opened 2022-11-17
Support Python 3.11


jklaise:
Python 3.11 came out on 27/10/22 so we should look to support it soon. However, as always, we're mainly limited by upstream support from `tensorflow` and `torch`:
 - [x] https://github.com/tensorflow/tensorflow/issues/58032 (PR included in tf 2.12).
 - [ ] https://github.com/pytorch/pytorch/issues/86566 (3.11 support, not including `torch.compile`, included in torch 2.0).
- [x] https://github.com/numba/numba/issues/8304 (PR merged, will be included in 0.57.0 according [to this](https://github.com/numba/numba/issues/8304#issuecomment-1480876090)).

ascillitoe:
Adding `numba` to the list:
- [ ] https://github.com/numba/numba/issues/8304

This might actually be our main blocker since `torch` and `tensorflow` are optional now?

jklaise:
@ascillitoe good point, however, I think we should wait before both `tensorflow` and `torch` are Python 3.11 compatible to claim that both our libraries are too given how extensively we use those two (also CI would fail since we install both by default to test everything).

ascillitoe:
Ah yeh good point about CI!

PR #764, opened 2023-04-05
Bump tensorflow to 2.12

ascillitoe:
2.12 release notes [here](https://github.com/tensorflow/tensorflow/releases/tag/v2.12.0).

Possibly important changes:

- [x] Python 3.11 added and 3.7 dropped. - _No action needed. We are tracking upgrade to Python 3.11 in https://github.com/SeldonIO/alibi-detect/issues/676, and 3.7 will be dropped when EOL_
- [x] Moved all saving-related utilities to a new namespace, keras.saving, for example: keras.saving.load_model, keras.saving.save_model, keras.saving.custom_object_scope, keras.saving.get_custom_objects, keras.saving.register_keras_serializable,keras.saving.get_registered_name and keras.saving.get_registered_object. The previous API locations (in keras.utils and keras.models) will be available indefinitely, but we recommend you update your code to point to the new API locations. - _We could change `tf.keras.models.load_model` used in `saving._tensorflow.loading` to `tf.keras.saving.load_model`, but would need logic to cope with older tf versions, and old api locations available indefinitely, so best to leave._
- [x] The new Keras model saving format (.keras) is available. You can start using it via model.save(f"{fname}.keras", save_format="keras_v3"). In the future it will become the default for all files with the .keras extension.... - _We could explore moving to `keras_v3` instead of `SavedModel`, but would not be compatible with older tf versions so probably best to leave..._

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ extras_require = {
     # https://github.com/SeldonIO/alibi-detect/issues/375 and 387
     "tensorflow": [
         "tensorflow_probability>=0.8.0, <0.19.0",
-        "tensorflow>=2.2.0, !=2.6.0, !=2.6.1, <2.12.0",  # https://github.com/SeldonIO/alibi-detect/issues/375 and 387
+        "tensorflow>=2.2.0, !=2.6.0, !=2.6.1, <2.13.0",  # https://github.com/SeldonIO/alibi-detect/issues/375 and 387
     ],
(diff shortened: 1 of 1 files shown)


ascillitoe:
@jklaise @mauicv CI is now passing. I've put down my thoughts above on what I see as the main changes of interest. To me it looks like we are good to merge (same for `alibi`).

PR #774, opened 2023-04-19
avoid failure of plot_feature_outlier_image

signupatgmx:
I call the function this way:
```py
plot_feature_outlier_image(preds,  # output of an prediction
                                   ds, # my set of images
                                   X_recon=recon, # reconstructed set of images
                                   max_instances=5,       # max nb of instances to display
                                   n_channels=3,          # number of channels of the images
                                   figsize=(20,10),
                                   outliers_only=True)    # only show outlier predictions. this function fails when no outlier is detected and parameter set to True
```
In case there is no outlieres detected and the parameter outliers_only is set to True, then the function fails, because it calls plt.subplots(nrows=0, ....). 0 rows throws the error.

### DIFF
--- a/alibi_detect/utils/visualize.py
+++ b/alibi_detect/utils/visualize.py
@@ -77,7 +77,7 @@ def plot_feature_outlier_image(od_preds: Dict,
         instance_ids = list(range(len(od_preds['data']['is_outlier'])))
     n_instances = min(max_instances, len(instance_ids))
     instance_ids = instance_ids[:n_instances]
-    
+
     if outliers_only and n_instances == 0:
         return
 
--- a/alibi_detect/utils/visualize.py
+++ b/alibi_detect/utils/visualize.py
@@ -79,6 +79,7 @@ def plot_feature_outlier_image(od_preds: Dict,
     instance_ids = instance_ids[:n_instances]
 
     if outliers_only and n_instances == 0:
+        warnings.warn('No outliers found!', UserWarning, stacklevel=3)
         return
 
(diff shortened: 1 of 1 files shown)


mauicv (changes requested):
Hey @signupatgmx,
Thank you very much for opening this PR! I think the change makes sense, one minor ask would be to add a warning if the `outliers_only and n_instances == 0` condition is met. If the function returns without warning the user then it's not entirely clear what happened. Lmk if you are happy to do this otherwise I'm happy to pick it up!
Thanks regardless

signupatgmx:
Hallo @mauicv, as I'm not customized with your coding standards, I'd suggest you add a suitable warning output. I'm glad to have found your great library.
Kind regards.

alibi_detect/utils/visualize.py
```diff
@@ -77,6 +78,11 @@ def plot_feature_outlier_image(od_preds: Dict,
         instance_ids = list(range(len(od_preds['data']['is_outlier'])))
     n_instances = min(max_instances, len(instance_ids))
     instance_ids = instance_ids[:n_instances]
+
+    if outliers_only and n_instances == 0:
+        warnings.warn('No outliers found!', UserWarning, stacklevel=3)
```
ascillitoe, line 83: @mauicv  is the warning swallowed without `stacklevel=3`?

mauicv: Nope, I've removed it!

ascillitoe (approved):
LGTM!

PR #776, opened 2023-04-24
Set `fail_ci_if_error` to false on codecov github acton

mauicv:
### What is this

Sets `fail_ci_if_error=false` in the codecov github ci. Quick fix that prevents the ci from falling due to [this](https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954) issue. See also #768


ascillitoe:
The `ubuntu-latest, 3.9` build gives us an example of the codecov upload failing and CI still passing (which is what we want) 👍🏻

PR #777, opened 2023-04-24
Update licenses

ascillitoe:
@jklaise I can't figure out where `fsspec` has come from. Do we care?


jklaise (approved):
Not particularly, as long as the licence is permissive.

jklaise:
It seems there are more licence changes now?

Issue #759, opened 2023-03-16
Fix `Returns` sections of some docstrings


ascillitoe:
In many `Returns` sections of the API docs, paragraphs are being incorrectly split into bulleted lists, [for example](https://docs.seldon.io/projects/alibi-detect/en/latest/api/alibi_detect.cd.base.html#alibi_detect.cd.base.BaseClassifierDrift.get_splits):

![image](https://user-images.githubusercontent.com/32061685/225721905-428bf102-7a45-4c2d-8b6d-8705ffb19d14.png)

See https://github.com/SeldonIO/alibi-detect/pull/677#discussion_r1139220459 for info on causes and fixes.

ascillitoe:
Also, the `alibi-detect` `CONTRIBUTING.md` could be updated with the revised docstring guidelines added to `alibi` in https://github.com/SeldonIO/alibi/pull/571.

PR #789, opened 2023-05-15
Docstring return statements

mauicv:
## what is this

Some return statements in alibi detect are formatted incorrectly. This PR fixes this issue #759. These changes are all taken from #748 which will be closed. As an example of the kind of thing this fixes compare [this](https://docs.seldon.io/projects/alibi-detect/en/latest/api/alibi_detect.ad.html#alibi_detect.ad.AdversarialAE.predict) to [this](https://seldon--789.org.readthedocs.build/projects/alibi-detect/en/789/api/alibi_detect.ad.html#alibi_detect.ad.AdversarialAE.predict).

- [x] Add Docstrings section from alibi `CONTRIBUTING.md` [here](https://github.com/SeldonIO/alibi/blob/master/CONTRIBUTING.md#docstrings)
- [x] Check [this](https://github.com/SeldonIO/alibi-detect/pull/789#discussion_r1194844271) behaviour


.git-blame-ignore-revs
```diff
@@ -0,0 +1,3 @@
+894744537dc64f729a44f64fbcb58e6b3b5dc1aa
```
jklaise, line 1: Since we're squashing commits, there should only be one entry per PR. Can you check if it's the case that the hash of the first commit is used after the squash?

alibi_detect/utils/tensorflow/distance.py
```diff
@@ -198,7 +198,7 @@ def permed_lsdds(
 
     Returns
     -------
-    Vector of B LSDD estimates for each permutation, H_lam_inv which may have been inferred, and optionally
+    Vector of B LSDD estimates for each permutation, H_lam_inv which may have been inferred, and optionally \
```
jklaise, line 201: Are these explicit line breaks actually needed?

.git-blame-ignore-revs
mauicv: I was actually wondering about this and meant to check how it works, and it seems the above does [not](https://github.com/orgs/community/discussions/16880). It seems that we'll have to do two separate PRs one with the changes and one with the relevant commits in `.git-blame-ignore-revs`.

mauicv: Well I could squash the PR commits to one and then add another commit before merging into master instead of squashing?

jklaise (approved):
LGTM!

alibi_detect/utils/tensorflow/distance.py
ascillitoe: Didn't realise this, good spot @mauicv !

CONTRIBUTING.md
```diff
@@ -45,6 +45,107 @@ When documenting Python classes, we adhere to the convention of including docstr
...
+Returns
+-------
+If the user has specified grouping, then the input object is subsampled and an object of the same \
+type is returned. Otherwise, a `shap_utils.Data` object containing the result of a k-means algorithm \
+is wrapped in a `shap_utils.DenseData` object and returned. The samples are weighted according to the \
+frequency of the occurrence of the clusters in the original data.
+```
+ 
+- Returning an object which contains multiple attributes and each attribute is described individually. 
+In this case the attribute name is written between single back-ticks and the type, if provided, would be written in
```
ascillitoe, line 106: Nitpick:

> and the type, **if provided**, would be written

Could we be a little more specific here? Not detailing types in the general case (since already done by `sphinx-autodoc-typehints`), but giving types when multiple objects are contained in another object (i.e. a `dict`), makes a lot of sense IMO. However, in the latter case, are we saying this should always be done? Or its optional?

ascillitoe (approved):
One minor nitpick (or question even), but otherwise LGTM!

CONTRIBUTING.md
mauicv: I think we're hinting that it's better to give types however it's not something we've done historically. I've [opened](https://github.com/SeldonIO/alibi-detect/issues/791) an issue to change this across the codebase and then perhaps we can change the language here to be a little stronger.

Issue #801, opened 2023-05-30
Faced error - Value Error: Unknown Error.Try using MMD-based detection instead


Rithika1000:
[E_Commerce_Churn_Kaggle.zip](https://github.com/SeldonIO/alibi-detect/files/11596476/E_Commerce_Churn_Kaggle.zip)
This file has the data that i have used. I read it as a csv using pandas ( E_Commerce_Dataset-Baseline.csv as ref_df and E_Commerce_Dataset-Production.csv as test_df) and then implemented the below code.

First i preprocessed the data as follows
`
categorical_feature_name=['Preferred_Device', 'Payment_Mode',
       'Gender', 'Prefered_Category']

def Encoder(df):
  le = LabelEncoder()
  for i in categorical_feature_name:
      encoded_df = le.fit_transform(df[[i]])
      df.drop([i], axis=1, inplace=True)
      df[i]=encoded_df
 return df
ref_df = Encoder(ref_df)
test_df = Encoder(test_df)

x_ref = np.asarray(ref_df, np.float32)
x_test = np.asarray(test_df, np.float32)
`

I then calculated drift using the following code:
`
lsdd1 = LSDDDrift(x_ref, p_val=0.05, backend='pytorch')

lsdd1_pred = lsdd1.predict(x_test)
stat = 'LSDD value='
stat_val, p_val = lsdd1_pred['data']['distance'], lsdd1_pred['data']['p_val']
print(f'{stat} {stat_val:.3f} -- p-value {p_val:.3f}')
`
i then got the following error
![image](https://github.com/SeldonIO/alibi-detect/assets/123930151/59c19dbd-74ff-479b-8f78-47da01ff34f8)
![image](https://github.com/SeldonIO/alibi-detect/assets/123930151/a72d34cc-6a5a-470c-a40f-0518fda6aa56)

jklaise:
Hi, can you please format the code using Github code blocks? It's very hard to read otherwise. https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks

jklaise:
@Rithika1000 I've chatted to my team and have been advised that the LSDD method is only appropriate for detecting drift on continuous distributions. Looking [at the docs](https://docs.seldon.io/projects/alibi-detect/en/latest/cd/methods/lsdddrift.html#Overview) this makes sense as densities only exist for continuous data, but your example features categorical data which the method would not work with.

[...]

PR #802, opened 2023-05-30
Clarify that LSDD is for continuous data use cases

jklaise:
Add a note and remove the tick for categorical data for `LSDD`. This should help clarify questions like that in #801.

@ojcobb is this in line with the other detectors that have/don't have a tick for categorical data? I realize that as long as some clever pre-processing is done (e.g. embedding categorical data in some continuous space), then the algorithm would work, but not with categorical data "out-of-the-box" (meaning label- or ohe-encoded).


ojcobb:
Yeah all of the other detectors with ticks are at least flexible enough to allow categorical data through kernel or model choice. Realistically MMD isn't designed for categorical data either and the nice theoretical result of an MMD estimate converging to its true underlying value as sample sizes increase does assume a continuous kernel. However this result is not hugely relevant to drift detection with predetermined sample sizes anyway and the MMD detector will at least handle (one-hot) categorical variables in a reasonable way and not break like LSDD will.

ojcobb:
Although I previously recommended MMD as an alternative in the error message, perhaps we are better off recommending the mixed-type tabular detector designed specifically for this use-case?

jklaise:
OK happy to change the recommendation to use `TabularDrift`.

Issue #813, opened 2023-06-12
Return data from od fit methods


mauicv:
- [x] Currently when you can fit on `gmm` and `svm` detectors the user can specify verbosity, however, we don't return any logs of the fit process despite computing them internally, These should be returned.
- [x] Also, make sure that returned data is consistent.

jklaise:
Do you mean actually return some object with the logs or just fiddling with the `logging` levels?

Bigger than this issue, but it's a great point to consider how we could expose application level logging. From our point of view, we just need to sprinkle `logging` statements, but then it's up to the user/application to define where to log and which level to expose (as in, our `verbose`) flags would not affect these. Good to check with `scikit-learn` as they have `verbose` flags too.

mauicv:
I mean some object with the logs although perhaps logging would be better. Specifically in the case of `gmm` we allow a convergence threshold to be set. The training ends on the basis of this criterion being met or the training process running for `max_epochs`. I felt this should be returned to the user so they can check if this has occured or at least set this value on the detector somewhere.

jklaise:
OK I see, it makes sense to both return something that can be programatically checked and also logging progress, both of which are kind of orthogonal.

Issue #806, opened 2023-06-12
OD-Issues


mauicv:
This issue tracks the issues relating to the outlier detector implementation:

- [ ] #805 
- [ ] #787 
- [ ] #779 
- [ ] #763 
- [ ] #762 
- [ ] #761 
- [ ] #760
- [ ] #755

PR #814, opened 2023-06-12
SVM Outlier detector

mauicv:
### What is this?

This PR adds the SVM outlier detector. Example notebook [here](https://colab.research.google.com/gist/mauicv/219bfc893c48b46d5001dd614f95a91b/svm-detector-pr746.ipynb#scrollTo=Rms-kqCGL3Do).

### Runtime comparisons:
Note that because I've allowed the user to set the optimization by the `kwarg` rather than using their choice of device this means the user can run the `sdg` method on the `gpu`. In reality, this just means that the Nystrome approximation will be performed on the GPU and but the SVM will be fit on the CPU.

see [this](https://gist.github.com/mauicv/5238d13592c1ae4af1f546c1a54d0c54) notebook. `x_ref` here is a dataset sampled from a normal distribution in 512 dimensions. `dsize` is the size of the dataset sampled. There are four combinations for each of `device = 'cpu'/'gpu'` or optimization equal to `gd`, `sdg`. We want to see `gpu-gd` as the lowest curve. 

device-optimization-analyisis

### TODO:
- [x] Remove sklearn backend
- [x] Implement SGD and GD variants
- [x] Add error checks for device-optimization combinations
- [x] Test device implementation in notebook
- [x] Verify correct tensors in backend implementation
- [x] Fix tests
- [x] Rewrite docstrings
- [x] Write tests for device/optimizer kwarg checks
- [x] Requested PR changes
- [x] Run GPU and CPU notebook and double check results
- [x] FInal check

### Notes:

1. SVM is a kernel detector by default and uses `GaussianRBF` which isn't torch-scriptable yet. Hence this PR will not implement torch scriptable functionality for this detector.
2. <s>The `GaussianRBF` is incompatible with the SKlearn backend which isn't really an issue because SKlearn users can specify the kernel as a string and this behaviour is exposed through the wrapper API. Will probably require a note in the docs though.</s>
3. Originally this was implemented with the option of two separate backends. The PyTorch backend performs well for GPUs and the Sklearn backend performs well on CPUs. The benefit to this approach was that the Sklearn backend is not dependent on the PyTorch dependency however this came at the cost of the user having to use Sklearn kernels which have to be configured in a different way to our Kernels. This means that the initialization of each detector is quite different. Having a different API for each backend is likely to confuse users and will require lots of documentation to help them navigate. The value add for these detectors is primarily that they have GPU support. If users still want to use the CPU then the torch dependency shouldn’t be a massive blocker. Instead, this was changed to have GdSVMDetector and SdgSVMDetector both of which will be torch backends but not selectable by the user through the backend `kwarg`. Instead, the user specifies an optimization `kwarg`. This is similar to the PCA linear and kernel backends ([here](https://github.com/SeldonIO/alibi-detect/blob/3ddb2ca023e2095bb0f839fbe6d10f580c6e397f/alibi_detect/od/pytorch/pca.py#L101) and [here](https://github.com/SeldonIO/alibi-detect/blob/3ddb2ca023e2095bb0f839fbe6d10f580c6e397f/alibi_detect/od/pytorch/pca.py#L171)) so it's not a massive break from the approach taken elsewhere.


alibi_detect/od/sklearn/svm.py
```diff
@@ -0,0 +1,143 @@
...
+
+class SVMSklearn(SklearnOutlierDetector):
+    ensemble = False
+
+    def __init__(
+        self,
+        n_components: Optional[int] = None,
+        kernel: Union[Literal['linear', 'poly', 'rbf', 'sigmoid'], Callable] = 'rbf',
+        sigma: Optional[float] = None,
+        kernel_params: Optional[Dict] = None,
```
jklaise, line 18: Argument order between backends different. Docstring order too...

alibi_detect/od/sklearn/svm.py
```diff
@@ -0,0 +1,143 @@
...
+    ):
+        """SKlearn backend for the One class support vector machine (SVM) outlier detector.
+
+        Parameters
+        ----------
+        n_components
+            Number of features to construct. How many data points will be used to construct the mapping.
```
jklaise, line 25: This is a very different description from the `torch` backend and the wrapper classes, please review.

jklaise: Probably applies to other arguments too.

alibi_detect/od/sklearn/svm.py
```diff
@@ -0,0 +1,143 @@
...
+            Ignored by other kernels.
+        kernel_params
+            Additional parameters (keyword arguments) for kernel function passed as callable object.
+
+        Raises
+        ------
+        ValueError
+            If `n_components` is less than 1.
+        """
+        super().__init__()
```
jklaise, line 46: Superclass is abstract so no constructor exists and call not needed?

alibi_detect/od/sklearn/svm.py
```diff
@@ -0,0 +1,143 @@
...
+            to the kernel; see the documentation for
+            `sklearn.metrics.pairwise <https://scikit-learn.org/stable/modules/classes.html#pairwise-metrics/>`_.
+            Ignored by other kernels.
+        kernel_params
+            Additional parameters (keyword arguments) for kernel function passed as callable object.
+
+        Raises
+        ------
+        ValueError
+            If `n_components` is less than 1.
```
jklaise, line 44: Not sure I see where this can be raised?

alibi_detect/od/sklearn/svm.py
```diff
@@ -0,0 +1,143 @@
...
+            max_iter=max_iter,
+            verbose=verbose,
+            nu=nu
+        )
+        self.gmm = self.gmm.fit(x_ref)
+        self._set_fitted()
+        return {
+            'converged': self.gmm.n_iter_ < max_iter,
+            'n_iter': self.gmm.n_iter_,
+        }
```
jklaise, line 103: Is this the first detector to return some info from `fit`? Need a task to check for consistency before public release.

alibi_detect/od/sklearn/svm.py
```diff
@@ -0,0 +1,143 @@
...
+            - n_iter: number of iterations performed.
+        """
+        n_components = self.n_components if self.n_components is not None else len(x_ref)
+        self.nystroem = Nystroem(
+            kernel=self.kernel,
+            n_components=n_components,
+            gamma=1. / (2. * self.sigma ** 2) if self.sigma is not None else None,
+            kernel_params=self.kernel_params,
+        )
+        x_ref = self.nystroem.fit(x_ref).transform(x_ref)
```
jklaise, line 91: nit: could do `fit_transform(x_ref)`?

alibi_detect/od/sklearn/svm.py
```diff
@@ -0,0 +1,143 @@
...
+
+        Returns
+        -------
+        Formatted kwargs.
+        """
+        return dict(
+            nu=fit_kwargs.get('nu', 0.5),
+            tol=fit_kwargs.get('tol', 1e-3),
+            max_iter=fit_kwargs.get('max_iter', 1000),
+            verbose=fit_kwargs.get('verbose', 0),
```
jklaise, line 121: Slightly worried that defaults are defined in two places - here and in `fit`. Is there a better way of doing this?

alibi_detect/od/sklearn/svm.py
```diff
@@ -0,0 +1,143 @@
...
+        `np.ndarray` of scores with leading batch dimension.
+
+        Raises
+        ------
+        NotFittedError
+            Raised if method called and detector has not been fit.
+        """
+        self.check_fitted()
+        x = self.nystroem.transform(x)
+        return - self.gmm.score_samples(x)
```
jklaise, line 143: nit: remove space after the minus sign (assuming it's not a bug to have a minus sign!)

alibi_detect/od/sklearn/svm.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        n_components = self.n_components if self.n_components is not None else len(x_ref)
+        self.nystroem = Nystroem(
+            kernel=self.kernel,
+            n_components=n_components,
+            gamma=1. / (2. * self.sigma ** 2) if self.sigma is not None else None,
+            kernel_params=self.kernel_params,
+        )
+        x_ref = self.nystroem.fit(x_ref).transform(x_ref)
+        self.gmm = SGDOneClassSVM(
```
jklaise, line 92: Is the name `gmm` misleading?

alibi_detect/od/pytorch/svm.py
```diff
@@ -0,0 +1,307 @@
...
+        self.sigma = sigma
+        self._init_kernel()
+
+    def _init_kernel(self) -> None:
+        """Initialize the Kernel."""
+
+        if isinstance(self.kernel, str):
+            if self.kernel not in ['rbf']:
+                raise ValueError(
+                    f'Currently only the rbf Kernel is supported for the SVM torch backend, got {self.kernel}.'
```
jklaise, line 58: This is a bit misleading as the user can use a custom kernel.

alibi_detect/od/pytorch/svm.py
```diff
@@ -0,0 +1,307 @@
...
+            - n_iter: number of iterations performed.
+            - lower_bound: loss lower bound.
+        """
+
+        self.nystroem = _Nystroem(
+            self.kernel,  # type: ignore[arg-type]
+            self.n_components
+        )
+
+        X_nys = self.nystroem.fit(x_ref).transform(x_ref)
```
jklaise, line 114: Noting that in `sklearn` backend the transformed data was called `x_ref`, although I think `X_nys` is clearer. Would suggest making the change in `sklearn` backend for consistency (and suggest sticking with lower-case convention, so `x_nys`).

jklaise (changes requested):
Nice one! A few comments, mostly to do with tidiness.

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,234 @@
...
+        device: Optional[Union[Literal['cuda', 'gpu', 'cpu'], 'torch.device']] = None,
+        kernel: Union['torch.nn.Module', Literal['linear', 'poly', 'rbf', 'sigmoid']] = 'rbf',
+        sigma: Optional[float] = None,
+        kernel_params: Optional[Dict] = None,
+    ) -> None:
+        """Support vector machine (SVM) outlier detector.
+
+        The Support vector machine outlier detector fits a one-class SVM to the reference data.
+
+        Rather than the typical approach of optimizing the exact kernel OCSVM objective through a dual formulation,
```
ojcobb, line 38: I wonder if this might be a bit too much detail for the highest level docstring. However I do intend to iterate slightly on these docstrings when doing the documentation so happy to leave until then.

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,234 @@
...
+        here we instead map the data into the kernel's RKHS and then solve the linear optimization problem
+        directly through its primal formulation. The Nystroem approximation can optionally be used to speed up
```
ojcobb, line 40: The Nystroem approximation isn't optional

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,234 @@
...
+}
+
+
+class SVM(BaseDetector, ThresholdMixin, FitMixin):
+    def __init__(
+        self,
+        n_components: Optional[int] = None,
+        backend: Literal['pytorch', 'sklearn'] = 'sklearn',
+        device: Optional[Union[Literal['cuda', 'gpu', 'cpu'], 'torch.device']] = None,
+        kernel: Union['torch.nn.Module', Literal['linear', 'poly', 'rbf', 'sigmoid']] = 'rbf',
```
ojcobb, line 30: The kernel definition here has become a little unwieldy and inconsistent with how we do things elsewhere. Also, although it is technically possible, users should not be specifying linear or polynomial kernels for this detector as the method relies on the kernel inducing an infinite dimensional RKHS.

I wonder if we instead encourage users to pass kernels in exactly the same way as they do for other detectors (so here we'd just have the `kernel` kwarg and not the `sigma` and `kernel_params` kwarg). We then always do our own nystrtoem approximation to produce kernalised features and only rely on sklearn for the OCSVM part?

ojcobb: I think consistency across detector definitions is particularly important given we envisage the primary use-case of the detector is as a component within a large ensemble.

alibi_detect/od/pytorch/svm.py
```diff
@@ -0,0 +1,307 @@
...
+
+        if isinstance(self.kernel, str):
+            if self.kernel not in ['rbf']:
+                raise ValueError(
+                    f'Currently only the rbf Kernel is supported for the SVM torch backend, got {self.kernel}.'
+                )
+            if self.kernel == 'rbf':
+                if self.sigma is not None:
+                    sigma = torch.tensor(self.sigma, device=self.device)
+                self.kernel = GaussianRBF(sigma=sigma)
```
ojcobb, line 63: Here if `self.sigma` is `None` then `sigma` is not defined and `sigma=sigma` causes an error

alibi_detect/od/pytorch/svm.py
```diff
@@ -0,0 +1,307 @@
...
+                    converged = True
+                    break
+                else:
+                    iter += 1
+
+                if verbose and isinstance(pbar, tqdm):
+                    pbar.set_postfix(dict(loss=loss.cpu().detach().numpy().item()))
+
+        self._set_fitted()
+        return {
```
ojcobb, line 173: Not against returning a dict like this for the `fit` method but if so we should make sure we're consistent across detectors (e.g. currently the GMM detector doesn't do this)

alibi_detect/od/pytorch/svm.py
```diff
@@ -0,0 +1,307 @@
...
+        Parameters
+        ----------
+        kwargs
+            dictionary of Kwargs to format. See `fit` method for details.
+
+        Returns
+        -------
+        Formatted kwargs.
+        """
+        return dict(
```
ojcobb, line 191: It seems like we're define defaults in two places. Is this necessary?

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,234 @@
...
+        sigma: Optional[float] = None,
+        kernel_params: Optional[Dict] = None,
```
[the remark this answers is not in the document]
mauicv: I've added an issue to [here](https://github.com/SeldonIO/alibi-detect/issues/806)

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,234 @@
...
+            Number of components in the Nystroem approximation By default uses all of them.
+        device
+            Device type used. The default tries to use the GPU and falls back on CPU if needed. Can be specified by
+            passing either ``'cuda'``, ``'gpu'``, ``'cpu'`` or an instance of ``torch.device``. Only used for the
+            ``'pytorch'`` backend.
+        sigma
+            Kernel coefficient for 'rbf', 'poly' and 'sigmoid'. Only used if the kernel is specified as a string.
+        kernel_params
+            Additional parameters (keyword arguments) for kernel function passed as a dictionary. Only used for the
+            '``sklearn``' backend and the kernel is a custom function.
```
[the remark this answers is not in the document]
mauicv: Note: No longer relevant due to other changes...

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,234 @@
...
+        if backend == 'pytorch':
+            args['device'] = device
+        if backend == 'sklearn':
+            args['kernel_params'] = kernel_params
+        self.backend = backend_cls(**args)
+
+    def fit(
+        self,
+        x_ref: np.ndarray,
+        nu: float,
```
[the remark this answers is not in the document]
mauicv: SKlearn uses the same name so I'm inclined to keep it. I've moved it to the `__init__` however.

alibi_detect/od/sklearn/svm.py
mauicv: Ahh, forgot to add this :facepalm: Added it now!

alibi_detect/od/sklearn/svm.py
mauicv: It's not, the `gmm` detector also returns stuff although as per [this](https://github.com/SeldonIO/alibi-detect/issues/813) issue it's only currently returned from the backend... I've added a bullet point there to make sure we also ensure that returned data is consistent.

alibi_detect/od/sklearn/svm.py
mauicv: The sklearn Nystroem method has been removed so this is no longer relevant. We now only use our own implementation.

alibi_detect/od/sklearn/svm.py
mauicv: Good point, I've made an issue for [this](https://github.com/SeldonIO/alibi-detect/issues/819) as I use the same pattern in the `gmm` detector.

alibi_detect/od/sklearn/svm.py
mauicv: :facepalm: Good catch, removed!

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,233 @@
...
+from alibi_detect.version import __version__
+
+if TYPE_CHECKING:
+    import torch
+
+
+backends = {
+    'pytorch': {
+        'sgd': SdgSVMTorch,
+        'gd': DgSVMTorch
```
ojcobb, line 20: Here the class names have the `d` and `g` the wrong way round. Also could we perhaps use "batch gradient descent" (so `BgdSVMTorch`) for the Pytorch variant? I think it makes the distinction clearer as just gradient descent is general enough that it could refer to both variants.

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,233 @@
...
+    }
+}
+
+
+class SVM(BaseDetector, ThresholdMixin, FitMixin):
+    def __init__(
+        self,
+        nu: float,
+        kernel: 'torch.nn.Module',
```
ojcobb, line 29: Maybe we could pass the gaussian RBF as a default here?

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,233 @@
...
+        `n` is the number of reference instances and `m` defines the number of inducing points. This can therefore be
+        expensive for large reference sets and benefits from implementation on the GPU.
+
+        In general if using a small dataset then using the `''cpu''` with the optimization `''sgd''` is the best choice.
+        Whereas if using a large dataset then using the `''gpu''` with the optimization `''gd''` is the best choice.
+
+        Parameters
+        ----------
+        nu
+            The proportion of the training data that should be considered outliers. Note that this does not necessarily
```
ojcobb, line 60: Lets make clear this should be thought of as a regularisation parameter that affects how smooth the decsion boundary will be.

alibi_detect/od/pytorch/svm.py
```diff
@@ -0,0 +1,441 @@
...
+
+        Raises
+        ------
+        NotFittedError
+            Raised if method called and detector has not been fit.
+        """
+        self.check_fitted()
+        x_nys = self.nystroem.transform(x)
+        x_nys = x_nys.cpu().numpy()
+        return self._to_tensor(-self.svm.score_samples(x_nys))
```
ojcobb, line 191: Currently, due to differences between how we and `sklearn` set up our optimisations, scores returned via `sgd` vs `gd` variants differ by a linear transformation. Lets fix the intercept at 1 and scale coefficients accordingly such that we have consistency in this regard. (Note that out `intercept` vairable it equal to `sklearn`'s `offset` variable as `intercept = 1-offset`.)

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,233 @@
...
+
+backends = {
+    'pytorch': {
+        'sgd': SdgSVMTorch,
+        'gd': DgSVMTorch
+    }
+}
+
+
+class SVM(BaseDetector, ThresholdMixin, FitMixin):
```
ojcobb, line 25: Can we pass `None` as the default for `n_components`?

alibi_detect/od/pytorch/svm.py
```diff
@@ -0,0 +1,441 @@
...
+            Kernel function to use for outlier detection.
+        n_components
+            Number of components in the Nystroem approximation, by default uses all of them.
+        device
+            Device type used. The default tries to use the GPU and falls back on CPU if needed. Can be specified by
+            passing either ``'cuda'``, ``'gpu'``, ``'cpu'`` or an instance of ``torch.device``.
+        """
+        if (isinstance(device, str) and device in ('gpu', 'cuda')) or \
+                (isinstance(device, torch.device) and device.type == 'cuda'):
+            warnings.warn(('The `sgd` optimization option is best suited for CPU. If '
```
ojcobb, line 101: Can we modify to convey that `sgd` will run on cpu regardless and that in this case only the nystroem approximation will be done on gpu.

ojcobb:
Much cleaner detector definition and abstractions now - definitely a good shout to have user specify optimisation method themselves. Nice one!

alibi_detect/utils/pytorch/losses.py
```diff
@@ -0,0 +1,7 @@
+import torch
+
+
+def hinge_loss(preds: torch.Tensor) -> torch.Tensor:
+    "L(pred) = max(0, 1-pred) summed over multiple preds"
```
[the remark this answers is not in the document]
ascillitoe: 👍🏻  (checking Janis' requested changes whilst he is away)

alibi_detect/od/_svm.py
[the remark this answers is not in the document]
ascillitoe: 👍🏻

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,234 @@
...
+        """Support vector machine (SVM) outlier detector.
+
+        The Support vector machine outlier detector fits a one-class SVM to the reference data.
+
+        Rather than the typical approach of optimizing the exact kernel OCSVM objective through a dual formulation,
+        here we instead map the data into the kernel's RKHS and then solve the linear optimization problem
+        directly through its primal formulation. The Nystroem approximation can optionally be used to speed up
+        training and inference by approximating the kernel's RKHS.
+
+        We provide two backends for the one class svm, one based on PyTorch and one based on scikit-learn. The
```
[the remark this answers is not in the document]
ascillitoe: 👍🏻

alibi_detect/od/_svm.py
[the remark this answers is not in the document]
ascillitoe: 👍🏻

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,234 @@
...
+        Moreover, the Nystroem approximation has complexity `O(n^2m)` where `n` is the number of reference instances
+        and `m` defines the number of inducing points. This can therefore be expensive for large reference sets and
+        benefits from implementation on the GPU. In general if using a small dataset opt for the sklearn backend and
+        if using a large dataset opt for the pytorch backend.
+
+        Parameters
+        ----------
+        kernel
+            Used to define similarity between data points. If using the pytorch backend, this can be either a string
+            specifying a built-in kernel. Alternatively, a custom kernel can be passed as a subclass of the
```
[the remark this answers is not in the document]
ascillitoe: 👍🏻 

`kernel` description changed to due to other changes...

alibi_detect/od/_svm.py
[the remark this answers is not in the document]
ascillitoe: 👍🏻

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,233 @@
...
+        Parameters
+        ----------
+        nu
+            The proportion of the training data that should be considered outliers. Note that this does not necessarily
+            correspond to the false positive rate on test data, which is still defined when calling the
+            `infer_threshold` method.
+        kernel
+            Kernel function to use for outlier detection. Should be an instance of a subclass of `torch.nn.Module`.
+        n_components
+            Number of components in the Nystroem approximation By default uses all of them.
```
ascillitoe, line 66: `.` needed after `approximation`.

alibi_detect/od/sklearn/svm.py
ascillitoe: 👍🏻 , since no longer separate `torch` and `sklearn` backends. Will also be covered by #818

alibi_detect/od/sklearn/svm.py
ascillitoe: 👍🏻

alibi_detect/od/sklearn/svm.py
ascillitoe: 👍🏻

alibi_detect/od/sklearn/svm.py
ascillitoe: 👍🏻

alibi_detect/od/sklearn/svm.py
ascillitoe: 👍🏻

alibi_detect/od/sklearn/svm.py
ascillitoe: 👍🏻

alibi_detect/od/sklearn/svm.py
ascillitoe: 👍🏻

alibi_detect/od/sklearn/svm.py
ascillitoe: 👍🏻

alibi_detect/od/pytorch/svm.py
ascillitoe: 👍🏻

alibi_detect/od/pytorch/svm.py
ascillitoe: 👍🏻

alibi_detect/od/_svm.py
ascillitoe: Agree with `BGD`.

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,233 @@
...
+            The range of values to be considered for the gradient descent step size at each iteration. This is specified
+            as a tuple of the form `(min_eta, max_eta)` and only used for the ``'gd'`` optimization.
+        n_step_sizes
+            The number of step sizes in the defined range to be tested for loss reduction. This many points are spaced
+            equidistantly along the range in log space. This is only used for the ``'gd'`` optimization.
+        n_iter_no_change
+            The number of iterations over which the loss must decrease by `tol` in order for optimization to continue.
+            This is only used for the ``'gd'`` optimization..
+        verbose
+            Verbosity level during training. ``0`` is silent, ``1`` a progress bar or sklearn training output for the
```
ascillitoe, line 140: `verbose` description is referring to `sklearn` implementation only. Unclear if `verbose` is still relevant when the `torch` backend is used. Would be good to make the description more generic, or be more descriptive as to when the `verbose` kwarg is relevant.

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,233 @@
...
+            Convergence threshold used to fit the detector. Used for both ``'sgd'`` and ``'gd'`` optimizations.
+            Defaults to ``1e-3``.
+        max_iter
+            The maximum number of optimization steps. Used for both ``'sgd'`` and ``'gd'`` optimizations.
+        step_size_range
+            The range of values to be considered for the gradient descent step size at each iteration. This is specified
+            as a tuple of the form `(min_eta, max_eta)` and only used for the ``'gd'`` optimization.
+        n_step_sizes
+            The number of step sizes in the defined range to be tested for loss reduction. This many points are spaced
+            equidistantly along the range in log space. This is only used for the ``'gd'`` optimization.
```
ascillitoe, line 135: maybe `equidistantly` -> `uniformly`?

alibi_detect/od/pytorch/svm.py
```diff
@@ -0,0 +1,441 @@
...
+        self,
+        x_ref: torch.Tensor,
+        step_size_range: Tuple[float, float] = (1e-6, 1.0),
+        n_step_sizes: int = 16,
+        tol: float = 1e-6,
+        n_iter_no_change: int = 25,
+        max_iter: int = 1000,
+        verbose: int = 0,
+    ) -> Dict:
+        """Fit the SVM detector.
```
ascillitoe, line 243: If we are being more descriptive about what is being fit in `SgdSVMTorch.fit` method, is it worth being slightly more descriptive here? i.e. `Fit the pytorch SVM detector` or similar?

alibi_detect/od/pytorch/svm.py
```diff
@@ -0,0 +1,441 @@
...
+        n_iter_no_change
+            The number of iterations over which the loss must decrease by `tol` in order for optimization to continue.
+        max_iter
+            The maximum number of optimization steps.
+        verbose
+            Verbosity level during training. ``0`` is silent, ``1`` a progress bar.
+
+        Returns
+        -------
+        Dictionary with fit results. The dictionary contains the following keys
```
ascillitoe, line 266: `keys:` ?

ascillitoe (approved):
I've attempted to review the changes requested by @jklaise, and it looks like they've all been addressed. Also agree the new hidden sklearn backend approach is far cleaner. 

Except for a few nitpicks (and @ojcobb's comments), everything else looks good! Although only moderate confidence...

alibi_detect/od/_svm.py
mauicv: Ok, I've basically just added your comment on at the end:

> The proportion of the training data that should be considered outliers. Note that this does not necessarily correspond to the false positive rate on test data, which is still defined when calling the `infer_threshold` method. `nu` should be thought of as a regularization parameter that affects how smooth the svm decision boundary is.

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,233 @@
+from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Union
+
+import numpy as np
+
+from alibi_detect.base import (BaseDetector, FitMixin, ThresholdMixin,
+                               outlier_prediction_dict)
+from alibi_detect.exceptions import _catch_error as catch_error
+from alibi_detect.od.pytorch import DgSVMTorch, SdgSVMTorch
```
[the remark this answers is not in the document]
mauicv: Yeah, for some reason `SGD_SVMTorch` and `SgdSvmTorch` seems worse than `SgdSVMTorch` to me. I'm ambivalent though, @ojcobb, any preference?

alibi_detect/od/_svm.py
mauicv: I've changed it to `spaced evenly`!

alibi_detect/od/_svm.py
```diff
@@ -0,0 +1,236 @@
...
+        }
+        args['device'] = device
+        self.backend = backend_cls(**args)
+
+    def fit(
+        self,
+        x_ref: np.ndarray,
+        tol: float = 1e-6,
+        max_iter: int = 1000,
+        step_size_range: Tuple[float, float] = (1e-6, 1.0),
```
ojcobb, line 115: Can we change default `step_size_range` to `(1e-8, 1.0)`?

ojcobb (approved):
LGTM!

alibi_detect/od/_svm.py
mauicv: Sorry, missed this! Now fixed!

Issue #679, opened 2022-11-22
Inconsistency in `device` kwarg between detectors and `preprocess_drift` function


ascillitoe:
The expected type for the `device` kwarg in detectors and the `preprocess_drift` function is inconsistent. 

### Detectors
When `backend='pytorch'`, some detectors accept a `device` kwarg which should be `'cpu'`, `'cuda'` or `None`. 

Docs Description:

> device: cuda or gpu to use the GPU and cpu for the CPU. If the device is not specified, the detector will try to leverage the GPU if possible and otherwise fall back on CPU.

Type signature:
```python
     device: Optional[str] = None,
```

### `preprocess_drift` function

The [preprocess_drift](https://docs.seldon.io/projects/alibi-detect/en/stable/api/alibi_detect.cd.pytorch.html?highlight=preprocess_drift#alibi_detect.cd.pytorch.preprocess_drift) function expects `device` to be a `torch.device`:

Docs description:
> device ([Optional](https://docs.python.org/3/library/typing.html#typing.Optional)[device]) – Device type used. The default None tries to use the GPU and falls back on CPU if needed. Can be specified by passing either torch.device(‘cuda’) or torch.device(‘cpu’).

## Solution

We should be consistent here. Possible changes include:

- Allow `device` to be passed to detectors as a `torch.device`. This will also allow the device `index` to be given e.g. 

   ```
   >>> torch.device('cuda', 0)
   device(type='cuda', index=0)
   ```

- For convenience, also retain the option to pass as a string too (and add this to `preprocess_drift`).
- Improve typing for string options e.g. with `Literal` or `Enum`.

ascillitoe:
**Additional note:** presently although we don't officially support it, passing a torch.device to detectors will generally work. However, because we default to cuda:0 internally if a GPU is detected, models/tensors etc will be distributed across multiple devices if the user passes a device such as torch.device('cuda', 2).

PR #829, opened 2023-07-11
Refactor torch device types out of od and into _types

### LINKED ISSUE #763: Remove duplicated `utils._types` logic
The following logic [here](https://github.com/SeldonIO/alibi-detect/blob/c3707d2885d25b74dd5bb62cbfebadee12291e75/alibi_detect/utils/_types.py#L8):

```py
# Literal for typing
if sys.version_info >= (3, 8):
    from typing import Literal  # noqa
else:
    from typing_extensions import Literal  # noqa
```

in the `utils._types` file is unnecessary as `typing_extensions` already does this under the hood. Hence we should default to importing `Literal` from `typing_extensions` directly instead of from `utils_types`

### LINKED ISSUE #779: Refactor device types to `_types` for outlier detectors
The `device` parameter for the outlier detectors is getting quite long: `device: Optional[Union[Literal['cuda', 'gpu', 'cpu'], 'torch.device']] = None,`. May make sense to refactor this to `_types.py`. As well as wherever else this is the case in the library.

mauicv:
## What is this:

Defines `TorchDeviceTypes: TypeAlias = Optional[Union[Literal['cuda', 'gpu', 'cpu'], 'torch.device']]` in `_types.py` and refactors the typing for the device in the detectors. 

fixes #779, #679. Also fixes #763


alibi_detect/utils/_types.py
```diff
@@ -1,16 +1,13 @@
...
-from typing import Union, Type
+from typing import Union, Type, Optional
+
 
 # Literal for typing
-if sys.version_info >= (3, 8):
-    from typing import Literal  # noqa
-else:
-    from typing_extensions import Literal  # noqa
+from typing_extensions import Literal
```
ascillitoe, line 10: Looks like https://github.com/SeldonIO/alibi-detect/issues/763 might be resolved?

ascillitoe: Also related, but best to keep as separate PR:
https://github.com/SeldonIO/alibi-detect/issues/838

alibi_detect/cd/classifier.py
```diff
@@ -122,8 +123,9 @@ def __init__(
             Optional additional kwargs when fitting the classifier. Only relevant for 'tensorflow' and
             'pytorch' backends.
         device
-            Device type used. The default None tries to use the GPU and falls back on CPU if needed.
-            Can be specified by passing either 'cuda', 'gpu' or 'cpu'. Only relevant for 'pytorch' backend.
+            Device type used. The default tries to use the GPU and falls back on CPU if needed.
+            Can be specified by passing either ``'cuda'``, ``'gpu'``, ``'cpu'`` or an instance of
+            ``torch.device``. Only relevant for 'pytorch' backend.
```
ascillitoe, line 128: Just out of curiosity, if you update the `intersphinx_mapping` like https://github.com/pytorch/pytorch/issues/10400 and then reference `torch.device` like `:py:class:torch.device`  (should be in backticks), does it work?

alibi_detect/utils/_types.py
```diff
@@ -37,3 +34,5 @@
 # type aliases, for use with mypy (must be FwdRef's if involving opt. deps.)
 OptimizerTF: TypeAlias = Union['tf.keras.optimizers.Optimizer', 'tf.keras.optimizers.legacy.Optimizer',
                                Type['tf.keras.optimizers.Optimizer'], Type['tf.keras.optimizers.legacy.Optimizer']]
+
+TorchDeviceType: TypeAlias = Optional[Union[Literal['cuda', 'gpu', 'cpu'], 'torch.device']]
```
ascillitoe, line 38: re the forward reference `'torch.device'` in here, I can't think of a good fix at the moment, but just noting that this  introduces lots of additional sphinx warnings, and is not rendered "perfectly" in the docs (we've gone from 6 to 29 warnings, which makes me sad).

I suspect the forward ref would be resolved during docs compilation if we installed `alibi-detect[all]` on read-the-docs (https://github.com/SeldonIO/alibi-detect/pull/499) which is now allowed, but it seems wasteful...

alibi_detect/saving/_pytorch/saving.py
```diff
@@ -129,3 +130,20 @@ def save_embedding_config(embed: TransformerEmbedding,
...
+
+def save_device_config(device: TorchDeviceType):
+    """
+
+    Parameters
+    ----------
+    device
+        Device type used. The default tries to use the GPU and falls back on CPU if needed.
+            Can be specified by passing either ``'cuda'``, ``'gpu'``, ``'cpu'`` or an instance of
+            ``torch.device``.
```
ascillitoe, line 143: Unnecessary indent?

ascillitoe: Docstring also seems slightly inaccurate? Maybe just something like `Torch device to be serialised`?

alibi_detect/saving/saving.py
```diff
@@ -188,6 +189,11 @@ def _save_detector_config(detector: ConfigurableDetector,
     if optimizer is not None:
         cfg['optimizer'] = _save_optimizer_config(optimizer)
 
+    # Serialize device
+    device = cfg.get('device')
+    if device is not None:
```
ascillitoe, line 194: Ref

ascillitoe: Instead of the `_save_device_config` wrapper, isn't it easier just to do `cfg['device'] = save_device_config_pt(device)` here?

alibi_detect/saving/schemas.py
```diff
@@ -295,7 +295,7 @@ class PreprocessConfig(CustomBaseModel):
     Optional tokenizer for text drift. Either a string referencing a HuggingFace tokenizer model name, or a
     :class:`~alibi_detect.utils.schemas.TokenizerConfig`.
     """
-    device: Optional[Literal['cpu', 'cuda']] = None
+    device: Optional[Literal['cpu', 'cuda', 'gpu']] = None
```
ascillitoe, line 298: Content of this pending decision on https://github.com/SeldonIO/alibi-detect/pull/829/files#r1262574030

ascillitoe (commented):
Few minor comments, main one regarding serialisation. 

I'll do a final pass once tests are written. Regarding tests, I reckon we could get away with a single unit test saving with `save_device_config` and then running through `get_device`? Parameterised with all the supported `device` types...

alibi_detect/saving/saving.py
ascillitoe: p.s. Maybe `_save_device` would be more accurate than `_save_device_config`? `_save_optimizer_config` etc are named `_config` since they do actually return a "config `dict`", whereas `_save_device_config` is only returning a `str`.

alibi_detect/saving/_pytorch/saving.py
```diff
@@ -129,3 +130,20 @@ def save_embedding_config(embed: TransformerEmbedding,
...
+
+    Returns
+    -------
+    a string with value ``'cuda'`` or ``'cpu'``.
```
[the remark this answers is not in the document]
mauicv: Shall we just format the `str(torch.device('cuda:0'))` to remove the device index and raise a warning alerting the user to the change?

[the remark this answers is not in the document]
ascillitoe: I like this solution. It is simple, and prevents serialised detectors being unloadable e.g. if saved with `cuda:8` and loaded on a 4 gpu machine. 

If we extend the pydantic validation to support the device index in the future, we could still save as `cuda`, and the user could manually add a device index in the `config.toml` if they desired.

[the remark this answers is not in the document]
mauicv: Ah, so when saving the detector first gets the config, then validates and _then_ replaces the values with the string representations... :thinking: I've added the Pydantic validation as It seems like the best way of going about this.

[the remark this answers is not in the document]
ascillitoe: Damn I forgot about the pre-saving validation...

alibi_detect/cd/classifier.py
mauicv: Can I leave this to explore in a separate issue? This PR already has a much wider scope than initially intended! :sweat_smile:

alibi_detect/utils/_types.py
mauicv: :face_exhaling: arg... I'll open an issue. Maybe this PR might need to be reigned in!

alibi_detect/saving/schemas.py
```diff
@@ -79,6 +79,33 @@ def validate_optimizer(cls, optimizer: Any, values: dict) -> Any:
...
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate_device
+
+    @classmethod
+    def validate_device(cls, device: Any, values: dict) -> Any:
+        backend = values['backend']
+        if backend == Framework.TENSORFLOW or backend == Framework.SKLEARN:
+            if device is not None:
+                raise TypeError('`device` should not be specified for TensorFlow or Sklearn models. Leave as `None`.')
```
ascillitoe, line 96: Nitckpick: `TensorFlow or SKlearn models` - "models" is maybe a little inaccurate, since we are talking about detector backends here?

Issue #811, opened 2023-06-12
Set meta data on all OD


mauicv:
Currently, only kNN and LOF OD detectors set metadata, this should be done for all. See also #807

Issue #807, opened 2023-06-12
Review data_type in meta


mauicv:
> Not isolated to this PR, but noting that we seem to be a little inconsistent across the new and old outlier detectors wrt to when `data_type` is hard-coded, and when it is optionally set via a kwarg. For some, it is hardcoded to `time-series` (which makes sense), for some (e.g. the old `Mahalanobis`) it is set via kwarg, and for some it is hard coded to `numeric`. Maybe worth opening an issue to review this more generally? 

> Already mentioned in https://github.com/SeldonIO/alibi-detect/issues/567#issuecomment-1193853273, but highlighting here since we are setting `data_type` in new detectors too...

_Originally posted by @ascillitoe in https://github.com/SeldonIO/alibi-detect/pull/746#discussion_r1210093971_

mauicv:
In the case of the new outlier detectors, the expectation is that they're all tabular-numeric. If the user has image or text data they need to do some preprocessing first. This assumption isn't true for detectors like the old mahalanobis outlier detector which can take categorical or numeric data for instance.

PR #830, opened 2023-07-12
Set meta in new outlier detectors

mauicv:
## What is this

Sets meta in new outlier detectors

Related: #807
Fixes: #811


ascillitoe (approved):
LGTM!

Issue #809, opened 2023-06-12
Threshold datatype


mauicv:
> Atm, whereas other detectors indeed output the threshold as a singleton array, the sklearn GMM detector outputs a float. Perhaps a float is better though?

_Originally posted by @ojcobb in https://github.com/SeldonIO/alibi-detect/pull/736#discussion_r1188464030_

PR #832, opened 2023-07-12
Convert threshold to float type acorss the new outlier detectors

mauicv:
## What is this

This PR fixes: #809. Specifically it ensures that the threshold value outputted by the outlier detectors on a predict call is of type float across both Sklearn and PyTorch backends.


ojcobb (approved):
LGTM!

PR #833, opened 2023-07-12
Rename interface methods for mapping between frontend and backend

### LINKED ISSUE #812: Docstrings/Naming for `_to_numpy` method in SKLearn backend
See [this](https://github.com/SeldonIO/alibi-detect/pull/736#discussion_r1190973956) comment. 

- [] `_to_numpy` is really mapping from the backend `dtype` to the frontend type. This means the method is badly named so should be something like: `map_to_backend_dtype`. 
- [] I think we also need to make sure the docstrings are clear this is a interface method

mauicv:
## What is this:

FIxes #812. Renames `_to_tensor` and `_to_numpy` methods on the base Sklearn and PyTorch backend classes to `_to_frontend_dtype` and `_to_backend_dtype`. Also expands there docstrings.


jklaise (approved):
LGTM

PR #835, opened 2023-07-12
Return data from the backend fit method

mauicv:
## What is this:

this fixes: #813. Given we fit differently in different cases the data returned is different in case.

- `SVM(..., optimization='sgd')`: Returns dictionary containing `converged` and `n_iter` where `n_iter` is the number of iterations until convergence.
- `SVM(..., optimization='bgd')`: Returns dictionary containing `converged`, `n_iter` and `lower_bound` where `n_iter` is as above and `lower_bound` is the loss achieved during training.
- `GMM(..., backend='pytorch')`: `converged`, `lower_bound` and `n_epochs` - because the GMM PyTorch backend performs gradient descent over n epochs this value is the number of epochs performed.
- `GMM(..., backend='sklearn')`: `converged`, `lower_bound` and `n_iter`.

### DIFF
--- a/alibi_detect/od/_gmm.py
+++ b/alibi_detect/od/_gmm.py
@@ -126,7 +126,7 @@ class GMM(BaseDetector, ThresholdMixin, FitMixin):
         verbose
             Verbosity level used to fit the detector. Used for both ``'sklearn'`` and ``'pytorch'`` backends. Defaults to ``0``.
         """
-        self.backend.fit(
+        return self.backend.fit(
             self.backend._to_backend_dtype(x_ref),
             **self.backend.format_fit_kwargs(locals())
         )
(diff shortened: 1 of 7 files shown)


jklaise (approved):
LGTM but would add the info on what's returned in the public docstrings of `fit`. Also seems `fit_logs` isn't used?

Is the plan also to return info from the existing ODs? As a minimum we should open a ticket and investigate what those returns might be.

mauicv:
> LGTM but would add the info on what's returned in the public docstrings of `fit`. Also seems `fit_logs` isn't used?
> 
> Is the plan also to return info from the existing ODs? As a minimum we should open a ticket and investigate what those returns might be.

Agree w.r.t. the docstrings, good catch! I'll add that update. I'll open a ticket for the old ODs. They could definitely return something but its a bit of a bigger job.

PR #839, opened 2023-07-17
Run Windows and MacOS CI on Python 3.10

jklaise:
Seems that `torch` support is there and since we only run one version on each non-linux platform we might as well run on the latest supported.

Will keep an eye if the MacOS infinite runtime still persists on 3.11.


jklaise:
1.x of `torch` not available for Windows or MacOS platforms, until we upgrade to 2.0 will need to revert to testing on 3.10 for those two platforms: https://pypi.org/project/torch/1.13.1/#files

jklaise:
Test finished, will merge for 3.10 test support.

Issue #532, opened 2022-06-08
Alternatives to `opencv` (`cv2`)


jklaise:
`opencv` is a fairly heavy dependency and requires building (see recent CI failures), but we only use it in one place for making image perturbations (https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/utils/perturbation.py). We should either explore replacing it with another library that's lighter (potentially [scikit-image](https://scikit-image.org/docs/dev/api/skimage.filters.html)) or make it an optional dependency.

kkwekkeboom:
Another suggestion is to not make it a required dependency and start recommending `opencv-python-headless` as alternative for `opencv-python`. Much better in docker images.

Dilip-Jain:
Pull Request #853  
Just created a pull request to solve this issue. 
1. Replaced OpenCV's GaussianBlur and filter2D with their counterparts from SciPy.
2. Substituted OpenCV's affineTransform and warp functions with the corresponding functionalities from scikit-image.
3. OpenCV is now an optional library with cv2 import dynamically handled within perturbations.py (for future-proofing and people (system/environment) with OpenCV can leverage its functionality too).

PS: I implemented every alternative (known to me) using scipy, scikit-image, custom numpy convolution, etc., and compared the results using Mean Squared Error (MSE) and Structural Similarity Index (SSI). If anyone is interested in those comparisons, let me know I'll put the notebook somewhere online.

PR #853, opened 2023-10-01
Alternatives to `opencv` (`cv2`) #532 [Solved]

Dilip-Jain:
### Problem
@jklaise mentions -
> `opencv` is a fairly heavy dependency and requires building (see recent CI failures), but we only use it in one place for making image perturbations (https://github.com/SeldonIO/alibi-detect/blob/master/alibi_detect/utils/perturbation.py). We should either explore replacing it with another library that's lighter (potentially [scikit-image](https://scikit-image.org/docs/dev/api/skimage.filters.html)) or make it an optional dependency.

### Solution
Hello maintainers and contributors,

I am pleased to submit this pull request which addresses the concern of OpenCV being a heavy dependency in the image perturbations module (alibi_detect/utils/perturbation.py). The requirement of `cv2` (`opencv`), as realized, was limited and eliminable as it was used only for making image perturbations. In this pull request, I have replaced specific OpenCV functionalities with the lighter and already required libraries, `scikit-image` and `SciPy`, ensuring that no new external libraries are introduced.

#### Key Replacements:

1. Replaced OpenCV's GaussianBlur and filter2D with their counterparts from SciPy.
2. Substituted OpenCV's affineTransform and warp functions with the corresponding functionalities from scikit-image.
3. Modified setup.py: Updated the setup.py file to make the opencv-python requirement optional, allowing users to install it based on their needs.
4. Handled cv2 Import Dynamically: Modified the code in perturbation.py to handle the cv2 import using a try-except block. This ensures that if OpenCV is installed, it is used, but the code remains functional without OpenCV as well.

#### Benefits of the Changes:

- **Reduced Dependency Size:** By replacing OpenCV with scikit-image and SciPy, we significantly reduce the overall library size, making the project more lightweight.
- **Simplified CI Pipeline:** The elimination of OpenCV as a dependency simplifies the continuous integration (CI) process and mitigates potential CI failures related to building and managing OpenCV.
- **Improved Accessibility:** Utilizing scikit-image and SciPy, which are already required libraries, enhances accessibility for developers, eliminating the need to manage an additional heavyweight dependency. With optional OpenCV dependency, users now have the option to install OpenCV (`opencv-python`) based on their specific use cases, preventing unnecessary bloat for those who don't require it.
- **Future-Proofing:** The try-except import handling allows for seamless transitions in case OpenCV becomes a necessity in the future, minimizing code changes and maintaining backward compatibility.
- **Enhanced Flexibility:** Developers who already have OpenCV installed can continue to leverage its functionality within the image perturbation module without any disruption.
- The changes in this pull request ensure compatibility and alignment with the project's objectives, maintaining high-quality image perturbations while minimizing dependencies.

Thank you for considering this pull request, and I welcome any feedback or suggestions for further improvements.

### DIFF
--- a/alibi_detect/utils/perturbation.py
+++ b/alibi_detect/utils/perturbation.py
@@ -2,7 +2,6 @@ import random
 from io import BytesIO
 from typing import List, Tuple, Union
 
-import cv2
 import numpy as np
 import skimage as sk
 from alibi_detect.utils.data import Bunch
@@ -13,6 +12,14 @@ from PIL import Image
 from scipy.ndimage import zoom
 from scipy.ndimage.interpolation import map_coordinates
 from skimage.filters import gaussian
+
+try: 
+    import cv2
+except:
+    cv2 = None
+    from scipy.signal import convolve2d
+    from skimage.transform import AffineTransform, warp
+    from skimage import img_as_ubyte
 
 
 def apply_mask(X: np.ndarray,
@@ -632,8 +639,17 @@ def disk(radius: float, alias_blur: float = 0.1, dtype=np.float32) -> np.ndarray
     aliased_disk = np.array((X ** 2 + Y ** 2) <= radius ** 2, dtype=dtype)
     aliased_disk /= np.sum(aliased_disk)
 
-    # supersample disk to antialias
-    return cv2.GaussianBlur(aliased_disk, ksize=ksize, sigmaX=alias_blur)
+    if cv2 is not None:
+        # supersample disk to antialias
+        return cv2.GaussianBlur(aliased_disk, ksize=ksize, sigmaX=alias_blur)
+
+    else:
+        # Create Gaussian kernel
+        kernel = np.exp(-(X**2 + Y**2) / (2.0 * alias_blur**2))
+        kernel /= np.sum(kernel)
+
+        # Convolve with the aliased disk
+        return convolve2d(aliased_disk, kernel, mode='same', boundary='wrap')
 
 
 def defocus_blur(x: np.ndarray, radius: int, alias_blur: float, xrange: tuple = None) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


jklaise:
Hi @Dilip-Jain and thanks for your contribution, this looks very helpful!

I would suggest removing the `cv2` option completely as I don't see a need for it if we can achieve the same functionality with `scipy` and `skimage`, it also reduced the burden on supporting another dependency.

Have you had a chance of running some of the example notebooks that use these transforms and see if it all works as expected? E.g. grepping for `defocus_blur` in `doc/source/examples` gives about 6 notebooks where these are used.

Dilip-Jain:
Hi @jklaise, 
Most (all) the example notebooks with `defocus_blur` or `elastic_transform` were corruptions for the cifar10c dataset so testing one for all was enough. However, only after around half an hour, I realized these "corruptions" weren't even using perturbation but rather picking up .npy objects from your Google storage. 🤦‍♂️

Regardless, while making modifications, I compared the outcomes of the previous and the updated code using random images and calculated the Mean Squared Error (MSE) and Structural Similarity Index (SSI). Both scores were close to 1, indicating highly similar outputs. In fact, based on these scores, I decided which implementation to use—whether it was scipy, skimage, or even a custom numpy convolution—to replace the (then) existing code.

Lastly, just so we are on the same page, the current commit has cv2, but only as an optional dependency and you want me to remove it in its entirety. The next commit (updated pull request ⬇️) will do exactly that! 😊
Doubt: what about cv2 (opencv) mentioned in conf.py and license!!

jklaise:
@Dilip-Jain apologies, I somehow missed your latest comment. I've also realized that the functions are not actually used to produce runtime corruptions either in the codebase or the examples.

I'm currently looking at the PR and testing it myself (basically checking if outputs are the same using `np.testing.assert_allclose`). I hope you don't mind if I push the removal of `cv2` to this branch? Then we should be able to get this change in this week.

Dilip-Jain:
Hi @jklaise,
Any progress on this?

Dilip-Jain:
Hey @jklaise, is it still under consideration? Do you know if we need any more updates?

PR #857, opened 2023-10-05
Remove deprecated multichannel keyword from `skimage.filters.gaussian`

### LINKED ISSUE #856: `multichannel` parameter deprecated in `skimage`
This PR actually introduced a bug: https://github.com/SeldonIO/alibi-detect/pull/855

The `multichannel` parameter has been deprecated. Will need to slightly modify the `gaussian_blur`, `glass_blur` functions in `utils/perturbation.py` and potentially bump the lower version bound of `skimage`.

jklaise:
Fixes #856.

`channel_axis` was introduced in 0.19 so bumping lower bound for `scikit-image` to that.

### DIFF
--- a/alibi_detect/utils/perturbation.py
+++ b/alibi_detect/utils/perturbation.py
@@ -561,8 +561,7 @@ def zoom_blur(x: np.ndarray, max_zoom: float, step_zoom: float, xrange: tuple = 
         return x_z
 
 
-def glass_blur(x: np.ndarray, sigma: float, max_delta: int, iterations: int,
-               channel_axis: int = -1, xrange: tuple = None) -> np.ndarray:
+def glass_blur(x: np.ndarray, sigma: float, max_delta: int, iterations: int, xrange: tuple = None) -> np.ndarray:
     """
     Apply glass blur.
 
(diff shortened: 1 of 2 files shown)


alibi_detect/utils/perturbation.py
```diff
@@ -561,7 +561,8 @@ def zoom_blur(x: np.ndarray, max_zoom: float, step_zoom: float, xrange: tuple =
         return x_z
 
 
-def glass_blur(x: np.ndarray, sigma: float, max_delta: int, iterations: int, xrange: tuple = None) -> np.ndarray:
+def glass_blur(x: np.ndarray, sigma: float, max_delta: int, iterations: int,
+               channel_axis: int = -1, xrange: tuple = None) -> np.ndarray:
```
jklaise, line 565: The public interface has now changed to explicitly include `channel_axis`, although it's unclear whether the implementation would respect anything other than the default `-1` value (it assumes the layout starts with `[height, width]` axes). Still the function is not used anywhere publicly in the code base or examples so I think this is fine.

ascillitoe: If we're not sure `channel_axis` actually works for anything other than `-1`, would it not be safer to remove it as a kwarg and just hard-code it inside `glass_blur` and `gaussian_blur`? i.e. just say in docstrings that these functions assume `[h, w, c]`?

jklaise: Agreed, updated here: https://github.com/SeldonIO/alibi-detect/pull/857/commits/ceca05a2b28f7ddfbf0069664ec05c1f65863211

PR #858, opened 2023-10-05
Fix to ensure older tensorflow versions still work

ascillitoe:
Changing two type annotations for `tf.keras.optimizers.legacy.Optimizer` to forward references since `tf.keras.optimizers.legacy` doesn't exist in older (<2.9?) tensorflow versions. This resulted in the following error:

```
>>> from alibi_detect.saving import save_detector
AttributeError: module 'tensorflow.keras.optimizers' has no attribute 'legacy'
```

We could define new type aliases like `OptimizersTF` in `alibi_detect/utils/_types.py`, but it seems to be overly complicating things to me (we'd need three groups, one containing classes only, one types only, and one containing both...). We're also already using forwardref's for the tensorflow optimizers in other places...

P.s. have tested this with `tensorflow == 2.8` and the error no longer occurs (when it does with `master`).

### DIFF
--- a/alibi_detect/saving/_tensorflow/loading.py
+++ b/alibi_detect/saving/_tensorflow/loading.py
@@ -130,7 +130,7 @@ def load_kernel_config(cfg: dict) -> Callable:
     return kernel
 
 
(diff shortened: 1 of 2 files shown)


jklaise:
Nice one!

PR #882, opened 2024-03-04
fix: Add Pydantic v1 shim

jesse-c:
This allows other dependencies to be upgraded, whilst Alibi Detect itself is migrated.

I've made some small QoL improvements around the DevEx.


alibi_detect/utils/pydantic_v1/fields/__init__.py
```diff
@@ -0,0 +1,6 @@
+# mypy: disable-error-code="assignment, no-redef"
```
mauicv, line 1: Why do you have two files for pydantic v1 imports? `pydantic_v1` and `pydantic_v1.feilds`?

alibi_detect/utils/pydantic_v1/fields/tests/test_import_pydantic_v1_fields.py
```diff
@@ -0,0 +1,14 @@
+def test_import():
```
mauicv, line 1: I'm not sure we really need to test this? Isn't it pydantic functionality? All we really do is import differently

alibi_detect/utils/pydantic_v1/tests/test_import_pydantic_v1.py
```diff
@@ -0,0 +1,6 @@
+def test_import():
```
mauicv, line 1: Same point as the `test_import_pydantic_v1_feilds.py`

mauicv (approved):
Couple of points/question but otherwise LGTM!

alibi_detect/utils/pydantic_v1/fields/__init__.py
jesse-c: They're separate modules, and this was simply the first approach that came to mind. I imagine there could possibly be 1 file, which would have to expose the `fields` module within it, somehow?

alibi_detect/utils/pydantic_v1/fields/tests/test_import_pydantic_v1_fields.py
jesse-c: Code coverage started to fail as this introduced new code + some branching, in that new code, so I added a relatively small unit test to ensure that code coverage doesn't drop below the accepted level

alibi_detect/utils/pydantic_v1/tests/test_import_pydantic_v1.py
```diff
@@ -0,0 +1,6 @@
...
+    from alibi_detect.utils.pydantic_v1 import BaseModel
```
sakoush, line 2: @jesse-c this is only testing one version of pydantic being installed (probably v2) as far as I can tell. We probably should have a way to add a test for the other version of pydantic being installed to really make sure that the shim is working as expected?

jesse-c: Good point! I can do that. I took a quick look at how they do it with Ray [1] and it makes sense for a similar approach. I'll put something together to test out my idea.

[1] https://github.com/ray-project/ray/pull/40451

jesse-c: I've created https://github.com/SeldonIO/alibi-detect/pull/890.

PR #890, opened 2024-04-23
test: Add Pydantic version to matrix variable

jesse-c:
This is a follow-up to where the Pydantic v1 shim was added [1]. It's inspired by Ray's approach [2].

This locks the version in the CI and I think that's okay since we don't need to keep on updating them, as they're intended as smoke tests.

[1] https://github.com/SeldonIO/alibi-detect/pull/882#discussion_r1575971961
[2] https://github.com/ray-project/ray/pull/40451

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Test with pytest
         run: |
           pytest --randomly-seed=0 alibi_detect
-          # Note: The pytest-randomly seed is fixed at 0 for now. Once the legacy np.random.seed(0)'s 
+          # Note: The pytest-randomly seed is fixed at 0 for now. Once the legacy np.random.seed(0)'s
           # are removed from tests, this can be removed, allowing all tests to use random seeds.
 
       - name: Upload coverage to Codecov


.github/workflows/ci.yml
```diff
@@ -52,6 +53,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip setuptools wheel
           python -m pip install --upgrade --upgrade-strategy eager -r requirements/dev.txt
+          python -m pip install --upgrade pydantic==${{matrix.pydantic-version}}
```
jesse-c, line 56: The string interpolation isn't working as I've assumed, so I'll fix it.

mauicv (approved):

.github/workflows/ci.yml
```diff
@@ -35,9 +35,14 @@ jobs:
       matrix:
         os: [ ubuntu-latest ]
         python-version: [ '3.8', '3.9', '3.10', '3.11']
+        pydantic-version: [ '1.10.15', '2.7.0' ]
```
sakoush, line 38: nit: ` 2.7.1` has just been released (literally minutes ago!). I dont think though that we need to test with the latest versions.

sakoush (approved):
lgtm

.github/workflows/ci.yml
jesse-c: ✅

jesse-c:
There are docs and static checks fails, which are unrelated to the changes here [1].

[1] https://github.com/SeldonIO/alibi-detect/actions/runs/8808915172

